PR #39, opened 2017-01-09
[CLIPPER-43] Adds Clipper demo that uses CIFAR-10 dataset to compare birds and airplanes

dcrankshaw:
This PR shouldn't be merged until #36 #37 and #38 are merged and the demo can be test against them.


dcrankshaw:
Replaced  by #55

PR #42, opened 2017-01-17
[CLIPPER-62] Refactor RPC threading structure

Corey-Zumar:


src/libclipper/src/rpc_service.cpp
```diff
@@ -43,21 +44,28 @@ RPCService::RPCService()
 RPCService::~RPCService() { stop(); }
 
 void RPCService::start(const string ip, const int port) {
+  if(active_.load(std::memory_order_seq_cst)) {
+    throw std::runtime_error("Attempted to start RPC Service when it is already running!");
+  }
   const string address = "tcp://" + ip + ":" + std::to_string(port);
-  active_ = true;
+  active_.store(true, std::memory_order_seq_cst);
```
dcrankshaw: Explicitly calling `store()` is fine, but just so you know, the assignment operators are overloaded for `atomic<bool>` to call the appropriate atomic methods (http://en.cppreference.com/w/cpp/atomic/atomic/operator%3D).

dcrankshaw:
Note that this also fixes CLIPPER-39.

src/libclipper/src/rpc_service.cpp
dcrankshaw: Actually on second thought, let's use the assignment operator.

s/active.store(true,...);/active_ = true;

src/libclipper/src/rpc_service.cpp
```diff
@@ -43,21 +44,28 @@ RPCService::RPCService()
...
+    manage_service(address);
+  });
 }
 
-void RPCService::stop() { active_ = false; }
+void RPCService::stop() {
+  if(active_.load(std::memory_order_seq_cst)) {
+    active_.store(false, std::memory_order_seq_cst);
+    rpc_thread.join();
+  }
```
dcrankshaw, line 62: remove the explicit calls to `load` and `store`:

```cpp
if (active_) {
  active_ = false;
  rpc_thread.join();
}
```

src/libclipper/src/rpc_service.cpp
```diff
@@ -43,21 +44,28 @@ RPCService::RPCService()
...
+}
 
 int RPCService::send_message(const vector<vector<uint8_t>> msg,
                              const int zmq_connection_id) {
-  if (!active_) {
+  if (!active_.load(std::memory_order_seq_cst)) {
```
dcrankshaw: remove explicit call to `load`

src/libclipper/src/rpc_service.cpp
```diff
@@ -107,38 +112,33 @@ void RPCService::manage_service(const string address,
     std::this_thread::sleep_for(std::chrono::seconds(1));
   }
 
-  while (true) {
-    if (!active) {
-      shutdown_service(address, socket);
-      return;
-    }
+  while (active_.load(std::memory_order_seq_cst)) {
```
dcrankshaw: remove explicit call to `load`

dcrankshaw (changes requested):
Can you run this through clang-format?

Other than a couple minor stylistic changes, this is good to go. Passes all tests on my machine.

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done!

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done

dcrankshaw (approved):

dcrankshaw:
LGTM

PR #43, opened 2017-01-17
[CLIPPER-61] Fix hash function for serializable inputs, add cache metrics

Corey-Zumar:


src/libclipper/src/datatypes.cpp
```diff
@@ -121,11 +123,7 @@ size_t FloatVector::serialize(uint8_t *buf) const {
 InputType FloatVector::type() const { return InputType::Floats; }
 
 size_t FloatVector::hash() const {
-  size_t cur_hash = 0;
-  for (const auto d : data_) {
-    cur_hash ^= std::hash<float>()(d);
-  }
-  return cur_hash;
+  return primitive_input_hash(data_);
```
dcrankshaw, line 128: This is fine for this PR, but hashing floats is generally a "Bad Idea" (for the same reasons that floating point comparison can be tricky).

Can you add a TODO to the `FloatVector` and `DoubleVector` hash function declarations in `datatypes.hpp` noting that we should change this, and create a JIRA issue to track this as well (include the JIRA issue number in the TODO)?

src/libclipper/test/serialization_test.cpp
```diff
@@ -272,4 +272,130 @@ namespace {
...
+    std::vector<uint8_t> byte_vec_1;
+    for(int i = 0; i < 100; i++) {
+      uint8_t* bytes = reinterpret_cast<uint8_t*>(i);
+      for(int j = 0; j < static_cast<int>(sizeof(int) / sizeof(uint8_t)); j++) {
+        byte_vec_1.push_back(bytes[j]);
+      }
+    }
+    std::vector<uint8_t> byte_vec_2 = byte_vec_1;
+    std::vector<uint8_t> byte_vec_3 = byte_vec_1;
+    std::reverse(byte_vec_3.begin(), byte_vec_3.end());
```
dcrankshaw: Why are you treating the byte vectors differently from the other primitive types? I'm getting a segfault when I run this test.

Can you use get_primitive_hash_vectors?
```
std::vector<std::vector<uint8_t>> byte_hash_vecs = get_primitive_hash_vectors<uint8_t>();
```

src/libclipper/test/serialization_test.cpp
```diff
@@ -272,4 +272,130 @@ namespace {
...
+    // renders the first and third vectors identical, so they should have the same hash
+    ASSERT_EQ(DoubleVector(double_hash_vecs[0]).hash(), DoubleVector(double_hash_vecs[2]).hash());
+  }
+
+  TEST(SerializationTests, ByteVectorsHashCorrectly) {
+    // Obtains 3 vectors containing byte intepretations of the integers 0-99.
+    // The first two are identical, and the third is reversed
+    std::vector<uint8_t> byte_vec_1;
+    for(int i = 0; i < 100; i++) {
+      uint8_t* bytes = reinterpret_cast<uint8_t*>(i);
```
dcrankshaw: I think this line is responsible for the segfault. You're reinterpreting (0,1,2,...) as pointers. When you try to index into `bytes` a few lines later, you're at the bottom of the address space accessing unallocated memory.

dcrankshaw (changes requested):
The `ByteVector` unit test is broken. Other than that, just a couple of comments to add.

src/libclipper/test/serialization_test.cpp
Corey-Zumar: Thanks for pinpointing this! I've removed this byte-specific logic entirely (see comment below).

src/libclipper/test/serialization_test.cpp
Corey-Zumar: I was treating byte vectors differently because statically casting an integer to a byte causes a loss of information. For the purposes of a test, it doesn't seem like this should matter. Taking your suggestion, I've opted to call `get_primitive_hash_vectors` with `uint8_t` as a parameter.

src/libclipper/src/datatypes.cpp
Corey-Zumar: Done. Here's the JIRA issue: https://clipper.atlassian.net/browse/CLIPPER-63

src/libclipper/test/serialization_test.cpp
Corey-Zumar: On second thought, I've modified `get_primitive_hash_vectors` to correctly cast the **address** of each integer loop variable to a `uint8_t` pointer and add byte data by accessing this pointer at indices `0,...,sizeof(int) / sizeof(uint8_t)`. This seems to be working properly. Previously, as you stated, I was casting each integer to type `uint8_t` directly, as opposed to casting its address.

src/libclipper/test/serialization_test.cpp
Corey-Zumar: Yes, you're right! The addresses of these integers should be interpreted as pointers. I've fixed this and moved the logic into `get_primitive_hash_vectors()`.

src/libclipper/test/serialization_test.cpp
```diff
@@ -14,15 +15,16 @@ namespace {
   const int NUM_STRING_INPUTS = 300;
 
   template<typename T>
-  void get_primitive_data_vectors(clipper::InputType type, std::vector<std::vector<T>> &data_vectors) {
+  std::vector<std::vector<T>> get_primitive_data_vectors() {
```
dcrankshaw, line 18: This is clever, but I think it's overly complicated for test code. I'd rather have more verbose and simpler code in our unit tests. If a test starts failing, no one should be struggling to try to understand what the test code is doing.

If you make `j` a `uint8_t` then you're only ever casting from `uint8_t` to longer types which won't lose any information. Then you don't have to mess with pointers at all. Does that fix the problem?

src/libclipper/src/datatypes.cpp
```diff
@@ -141,6 +143,8 @@ size_t DoubleVector::serialize(uint8_t *buf) const {
 }
 
 size_t DoubleVector::hash() const {
+  // TODO [CLIPPER-63]: Find an alternative to hashing doubles directly,
+  // as this is generally a bad idea due to comparison difficulties
```
dcrankshaw: s/due to comparison difficulties/due to loss of precision from floating point representation

src/libclipper/test/serialization_test.cpp
```diff
@@ -350,37 +356,27 @@ namespace {
   }
 
   TEST(SerializationTests, ByteVectorsHashCorrectly) {
```
dcrankshaw: These aren't serialization tests.

Can you rename the test case for the tests you added to "InputHashTests" and rename the existing tests to "InputSerializationTests". Then let's rename this file to `input_test.cpp`.

dcrankshaw (changes requested):
Couple more comments about the tests.

dcrankshaw:
Also, can you rebase on develop so this is mergeable?

src/libclipper/test/serialization_test.cpp
Corey-Zumar: Yep, this fixes the problem! Made `j` a `uint8_t` index.

src/libclipper/test/serialization_test.cpp
Corey-Zumar: Done!

src/libclipper/src/datatypes.cpp
Corey-Zumar: Done!

Corey-Zumar:
Rebased on develop

dcrankshaw:
LGTM

dcrankshaw (approved):

PR #45, opened 2017-01-18
Adds support for running model containers with Docker

dcrankshaw:
Replaces #37

### DIFF
--- a/containers/python/rpc.py
+++ b/containers/python/rpc.py
@@ -4,6 +4,7 @@ import threading
 import numpy as np
 import struct
 from datetime import datetime
+import socket
 
 INPUT_TYPE_BYTES = 0
 INPUT_TYPE_INTS = 1
@@ -18,7 +19,7 @@ REQUEST_TYPE_FEEDBACK = 1
 def string_to_input_type(input_str):
     input_str = input_str.strip().lower()
     byte_strs = ["b", "bytes", "byte"]
-    int_strs = ["i", "ints", "int"]
+    int_strs = ["i", "ints", "int", "integer", "integers"]
     float_strs = ["f", "floats", "float"]
     double_strs = ["d", "doubles", "double"]
     string_strs = ["s", "strings", "string", "strs", "str"]
@@ -140,11 +141,11 @@ class Server(threading.Thread):
                         raw_content.split('\0')[
                             :-1], dtype=input_type_to_dtype(input_type))
                 else:
-                    inputs = np.split(
+                    inputs = np.array(np.split(
                         np.frombuffer(
                             raw_content, dtype=input_type_to_dtype(
                                 input_type)),
-                        splits)
+                        splits))
 
                 t3 = datetime.now()
 
@@ -202,7 +203,7 @@ class ModelContainerBase(object):
         pass
 
 
-def start(model, ip, port, model_name, model_version, input_type):
+def start(model, host, port, model_name, model_version, input_type):
     """
     Args:
         model (object): The loaded model object ready to make predictions.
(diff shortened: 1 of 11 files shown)


dcrankshaw:
@Corey-Zumar Can you take a look at this?

containers/python/rpc.py
```diff
@@ -202,7 +206,7 @@ def predict_strings(self, inputs):
         pass
 
 
-def start(model, ip, port, model_name, model_version, input_type):
+def start(model, host, port, model_name, model_version, input_type):
```
Corey-Zumar, line 210: As @giulio-zhou mentioned on #37, lets update the args docstring to reflect the change from `ip` to `host`.

containers/python/rpc.py
```diff
@@ -212,6 +216,11 @@ def start(model, ip, port, model_name, model_version, input_type):
         model_version (int): The version of the model
         input_type (str): One of ints, doubles, floats, bytes, strings.
     """
+
+    try:
+        ip = socket.gethostbyname(host)
+    except socket.error as e:
+        print("Error resolving %s: %s" % (host, e))
```
Corey-Zumar, line 224: We should exit here via `sys.exit(1)`. Otherwise, we attempt to connect to a potentially invalid address.

Corey-Zumar (changes requested):
I'm running into some issues when executing `./bin/build_docker_images.sh`. 

When the script executes `./configure --cleanup-quiet` within a newly created docker container, the following error is encountered: `./configure: line 237: g++: command not found`. Are there steps for satisfying the `g++` dependency that can be included automatically?

Additionally, when executing `cmake -D CMAKE_BUILD_TYPE=Release`, the following error is being thrown:
```
CMake Error at CMakeLists.txt:6 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

It is worth noting that the dependency for `cmake` itself is satisfied. I'm assuming that our script should set this environment variable as well?

dcrankshaw:
Weird that `g++` was missing. I thought I had added that to the docker build. The `CMAKE_CXX_COMPILER` issue is because g++ is missing. Digging into it now.

dcrankshaw:
Aha! The base docker file removed the g++ dependency. Looks like [this](https://github.com/frol/docker-alpine-gcc/commit/a679c362a135c7f663cd8674fa1359e1287bd492) is the offending commit. I'll switch the docker files to use the C++ image.

containers/python/rpc.py
dcrankshaw: Done

containers/python/rpc.py
dcrankshaw: Nice catch. Done.

dcrankshaw:
@Corey-Zumar addressed your comments. Docker images are building again.

Corey-Zumar:
Looks good! `./bin/build_docker_images.sh` now executes properly. If possible, it'd be good to silence the developer-specific warnings for Google Test. The following warning is displayed three times during the process of building the release version of Clipper:

```
CMake Warning (dev) at release/googletest-src/CMakeLists.txt:3 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    PROJECT_VERSION
    PROJECT_VERSION_MAJOR
    PROJECT_VERSION_MINOR
    PROJECT_VERSION_PATCH
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Also, are all of the line formatting changes all the result of clang-format? Good to merge if so.

Corey-Zumar (approved):

dcrankshaw:
Those warnings are hard to suppress. There's a [PR](https://github.com/google/googletest/issues/709) in the google test repo out there to fix it that's been sitting in limbo. I commented on it so hopefully it gets merged soon, but in the meantime I don't know of an easy fix.

PR #46, opened 2017-01-20
[CLIPPER-19, CLIPPER-56] Basic JSON Serialization/deserialization functionality and more informative error messages with RapidJSON library

giulio-zhou:


giulio-zhou:
This is only done for query_frontend to demonstrate how such a change would look.. If we decide to use this library, the changes can be easily translated to the management code.

giulio-zhou:
For reference, I get the following when I try to grab a key that's not in the JSON dictionary:
```
Assertion failed: (false), function operator[], file ./rapidjson/document.h, line 1103.
Abort trap: 6
```
Not the most informative error message. Parse errors are a bit more informative, like the following when I omit `]` from a JSON array:
```
JSON parse error: Missing a comma or ']' after an array element. (offset 62)
```
so perhaps those can be left as is for now. Semantic errors (array elements of the wrong type, missing keys, wrong value type, etc) will need to be handled on a case-by-case basis.

src/frontends/src/json_util.hpp
```diff
@@ -0,0 +1,203 @@
...
+  if (val.GetType() != expected_type) {
+    throw json_semantic_error("Type mismatch! JSON key " +
+                              std::string(key_name) + " expected type " +
+                              kTypeNames[expected_type] + "but found type " +
+                              kTypeNames[val.GetType()]);
+  }
+}
+
+/* Getters with error handling for double, float, long, int, string */
+double getDouble(rapidjson::Document& d, const char* key_name) {
```
giulio-zhou: The five non-array get functions all follow the same pattern. Wanted to get your input on the error message.

src/frontends/src/query_frontend.hpp
```diff
@@ -228,15 +170,19 @@ class RequestHandler {
...
+    if (!ok) {
+      std::stringstream ss;
+      ss << "JSON parse error: " << rapidjson::GetParseError_En(ok.Code())
+         << " (offset " << ok.Offset() << ")\n";
+      throw json_parse_error(ss.str());
+    }
 
-    long uid = pt.get<long>("uid");
-    std::shared_ptr<Input> input = decode_input(input_type, pt);
+    long uid = getLong(d, "uid");
```
giulio-zhou: First instance of parsing function reuse outside of `json_util.hpp`. The idea I mentioned previously is that you can simply call `getLong` or `getIntArray` without repeating the error-handling logic.

src/frontends/src/query_frontend_tests.cpp
```diff
@@ -110,19 +103,19 @@ TEST_F(QueryFrontendTest, TestDecodeMalformedJSON) {
   ASSERT_THROW(
       rh_.decode_and_handle_predict(gibberish_string1, "test", {},
                                     "test_policy", 30000, InputType::Doubles),
-      ptree_error);
+      json_parse_error);
```
giulio-zhou, line 106: Replaced the `ptree_error` and `ptree_bad_data` exceptions with the `json_parse_error` and `json_semantic_error` exceptions that I've defined. This way, we don't have to rely on the JSON library's own exceptions so these tests will remain exactly the same even if we swap out JSON parsing implementations.

src/libclipper/include/clipper/datatypes.hpp
```diff
@@ -20,6 +20,11 @@ enum class InputType {
   Strings = 4,
 };
 
+enum class OutputType {
```
giulio-zhou, line 23: Decided to consolidate the `OutputType` enums and the to/from-string functions in `datatypes.hpp/cpp`. Let me know what you think.

src/libs/CMakeLists.txt
```diff
@@ -41,3 +41,5 @@ export(TARGETS cxxopts FILE CxxOptsConfig.cmake)
 
 
 add_subdirectory(redox)
+
+add_subdirectory(rapidjson)
```
giulio-zhou: I added `rapidjson` as a submodule similar to how you added redox, where the remote repo that I link to is my personal fork (for now).

src/frontends/src/json_util.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+namespace clipper_json {
+
+class json_parse_error : public std::runtime_error {
+ public:
+  json_parse_error(const std::string& what) : std::runtime_error(what) {}
+  ~json_parse_error() throw(){};
+};
+
+class json_semantic_error : public std::runtime_error {
```
giulio-zhou: Would it be useful to have more granularity for semantic errors (e.g. type mismatch, only accept positive values, ...). These could extend `json_semantic_error`.

src/frontends/src/query_frontend.hpp
```diff
@@ -228,15 +170,19 @@ class RequestHandler {
       std::string json_content, std::string name,
       std::vector<VersionedModelId> models, std::string policy,
       long latency_slo_micros, InputType input_type) {
-    std::istringstream is(json_content);
-    ptree pt;
-    read_json(is, pt);
+    rapidjson::Document d;
```
giulio-zhou: Actually, this should be moved into `json_util.hpp`. Can do that in a future commit but the idea (for error handling and the output error message would be the same). `rapidjson::GetParseError_En(ok.Code())` outputs a useful error message as well as an offset `ok.Offset()` in the JSON string, but does not get more detailed than that.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "src/libs/redox"]
 	path = src/libs/redox
 	url = https://github.com/dcrankshaw/redox.git
+[submodule "src/libs/rapidjson"]
+	path = src/libs/rapidjson
+	url = https://github.com/giulio-zhou/rapidjson.git
```
dcrankshaw: Why add this as a submodule? It's a header-only library. Just copy the headers into `src/libs` the same way we did for the HTTP server or ZMQ. Make sure to include the license when you copy the headers over.

src/frontends/src/json_util.hpp
dcrankshaw: It may be useful in the future, but just distinguishing between semantic errors and parse errors is a good start. Can we provide the expected schema in the case of a semantic error?

src/frontends/src/json_util.hpp
```diff
@@ -0,0 +1,203 @@
...
+  check_type(d, key_name, rapidjson::kNumberType);
+  rapidjson::Value& v = d[key_name];
+  if (!v.IsDouble()) {
+    throw json_semantic_error("Input of type " + kTypeNames[v.GetType()] +
+                              "is not of type double");
+  }
+  return v.GetDouble();
+}
+
+float getFloat(rapidjson::Document& d, const char* key_name) {
```
dcrankshaw: these methods should be `snake_case` not `camelCase`

src/frontends/src/json_util.hpp
```diff
@@ -0,0 +1,203 @@
...
+      throw json_semantic_error("Array input of type " +
+                                kTypeNames[elem.GetType()] +
+                                "is not of type int");
+    }
+    vals.push_back(elem.GetInt());
+  }
+  return vals;
+}
+
+std::shared_ptr<Input> decode_input(InputType input_type,
```
dcrankshaw: Let's call this `parse_input` not `decode_input`

src/frontends/src/json_util.hpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    case InputType::Bytes: {
+      throw std::invalid_argument("Base64 encoded bytes are not supported yet");
+    }
+    default:
+      throw std::invalid_argument("input_type is not a valid type");
+  }
+}
+
+Output decode_output(OutputType output_type, rapidjson::Document& parsed_json) {
```
dcrankshaw: s/decode/parse

dcrankshaw (commented):
This looks like a reasonable approach. Certainly better than property trees! One high-level comment: most of this code is to help parsing JSON. Can you think about some utilities that will make it easier to serialize things to JSON? For example, check out the metrics code which writes metrics in JSON and see if there are ways we could clean that up as well. JSON is a really useful format for providing structured readable output as well (e.g. for metrics, logging, selection policy state, etc) and it would be nice if it were easy to serialize new objects to JSON.

.gitmodules
giulio-zhou: Main concern was that the rapidjson headers total around 15000 lines of code, which seemed a bit excessive. Otherwise there shouldn't be any problem.

src/frontends/src/json_util.hpp
giulio-zhou: Cool, will fix that.

src/frontends/src/json_util.hpp
giulio-zhou: Sounds good.

src/frontends/src/json_util.hpp
giulio-zhou: Do you mean something like
```
<error message>
Expected JSON format for feedback query request:
   * {
   *  "uid" := string,
   *  "input" := [double] | [int] | [string] | [byte] | [float],
   *  "model_name" := string,
   *  "model_version" := int,
   *  "label" := double
   * }
```
? Might be a little too much and also not completely appropriate for `json_semantic_error` in the current form (especially since the exception gets thrown pretty far down at the elementary parsing level). I think that at the very least, we can send the user an HTTP response containing a basic error message (from the exception) and the expected schema (by changing the arguments to the `respond_http` calls in the respective RequestHandler implementations). In the future, it might be worth trying to parse each argument rather than just stopping at the first error so that we can enumerate out each issue with the input JSON, but we may have to account for increased overhead in trying to compile these more useful error messages.

giulio-zhou:
Thanks for the feedback, I'll make the appropriate changes and think about how JSON serialization utilities. The library actually does contain a fairly rich set of JSON writing utilities which, until now, I haven't examined since I've been focusing on just deserialization. I think a good place to start would be to define some functions like `add_double_array(std::vector<double> values_to_add, std::string key_name, rapidjson::Document& d)` to add serialized key-value pairs and `add_object(rapidjson::Document& d, std::string key_name, rapidjson::Document& d_to_add)` for nested JSON objects. Let me know what you think.

src/libclipper/test/json_util_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+  parse_json(wrong_name_type_json, d);
+  ASSERT_THROW(get_candidate_models(d, "wrong_name_type"), json_semantic_error);
+
+  parse_json(wrong_version_type_json, d);
+  ASSERT_THROW(
+        get_candidate_models(d, "wrong_version_type"),
+        json_semantic_error);
+}
+
+TEST(JsonUtilTests, TestParseNestedObject) {
```
giulio-zhou, line 208: @Corey-Zumar Let me know if you have any feedback on this example JSON deserialization with nested JSON objects.

src/libclipper/test/json_util_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+  rapidjson::Document d;
+  d.SetObject();
+  add_string_array(string_array, "string_array", d);
+  add_double_array(double_array, "double_array", d);
+  add_int_array(int_array, "int_array", d);
+  std::string output_json = to_json_string(d);
+  EXPECT_EQ(output_json, expected_json); 
+}
+
+TEST(JsonUtilTests, TestCorrectJsonNestedObjects) {
```
giulio-zhou, line 47: @Corey-Zumar In contrast to the example below, this is an example of how you'd go about creating nested JSON objects.

dcrankshaw:
@giulio-zhou Is this ready for review?

giulio-zhou:
Commit is ready for review. I also pushed a minor change to query_frontend and management_frontend servers to return the JSON schema with an appended error message. Let me know what you think.

.gitmodules
dcrankshaw: Change this to dcrankshaw/rapidson

src/frontends/src/query_frontend.hpp
```diff
@@ -5,101 +5,53 @@
...
-    }
-    case OutputType::Int: {
-      double y_hat = parsed_json.get<int>("label");
-      return Output(y_hat, versioned_model);
-    }
-    default:
-      throw std::invalid_argument("output_type is not a valid type");
-  }
-}
+const std::string PREDICTION_JSON_SCHEMA =
```
dcrankshaw: Use string literals here (http://en.cppreference.com/w/cpp/language/string_literal): `R"(...)";`

src/frontends/src/query_frontend.hpp
```diff
@@ -5,101 +5,53 @@
...
+   "JSON format for prediction query request:\n"
+   "{\n"
+   "\"uid\" := string,\n"
+   "\"input\" := [double] | [int] | [string] | [byte] | [float],\n"
+   "}\n";
+
+const std::string UPDATE_JSON_SCHEMA =
```
dcrankshaw: String literal

src/frontends/src/query_frontend.hpp
```diff
@@ -184,8 +137,12 @@ class RequestHandler {
           std::string content = ss.str();
           respond_http(content, "200 OK", response);
         });
-      } catch (const ptree_error& e) {
-        respond_http(e.what(), "400 Bad Request", response);
+      } catch (const json_parse_error& e) {
+        std::string error_msg = PREDICTION_JSON_SCHEMA + e.what();
```
dcrankshaw: Let's add some explanation here. The error message should read like:
```
Error parsing JSON: e.what(). Expected JSON schema: PREDICTION_JSON_SCHEMA
```

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -0,0 +1,362 @@
...
+#include <stdexcept>
+
+using clipper::Input;
+using clipper::InputType;
+using clipper::Output;
+using clipper::OutputType;
+using clipper::VersionedModelId;
+using rapidjson::Type;
+
+static std::vector<std::string> kTypeNames = {
```
dcrankshaw: What are these type names? Can you add a comment?

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -0,0 +1,362 @@
...
+  kNullType = 0,      // null
+  kFalseType = 1,     // false
+  kTrueType = 2,      // true
+  kObjectType = 3,    // object
+  kArrayType = 4,     // array
+  kStringType = 5,    // string
+  kNumberType = 6     // number
+}; */
+
+namespace clipper_json {
```
dcrankshaw: Let's use nested namespaces. Make this `clipper::json`

src/libclipper/test/json_util_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+  add_string(string_val, "string_val", d);
+  add_double(double_val, "double_val", d);
+  add_int(int_val, "int_val", d);
+  std::string output_json = to_json_string(d);
+  EXPECT_EQ(output_json, expected_json); 
+}
+
+TEST(JsonUtilTests, TestCorrectJsonArrays) {
+  std::string expected_json =
+      "{\"string_array\":[\"test1\",\"word2\",\"phrase3\"],"
```
dcrankshaw: Use raw string literal here

src/management/src/management_frontend.hpp
```diff
@@ -38,14 +42,37 @@ const std::string GET_METRICS = ADMIN_PATH + "/metrics$";
...
-template <typename T>
-std::vector<T> as_vector(ptree const& pt, ptree::key_type const& key) {
-  std::vector<T> r;
-  for (auto& item : pt.get_child(key)) {
-    r.push_back(item.second.get_value<T>());
-  }
-  return r;
-}
+const std::string APPLICATION_JSON_SCHEMA =
+    "JSON format for application add request:\n"
```
dcrankshaw: raw string literal

src/management/src/management_frontend.hpp
```diff
@@ -112,8 +122,12 @@ class RequestHandler {
           try {
             std::string result = add_application(request->content.string());
             respond_http(result, "200 OK", response);
-          } catch (const ptree_error& e) {
-            respond_http(e.what(), "400 Bad Request", response);
+          } catch (const json_parse_error& e) {
+            std::string err_msg = APPLICATION_JSON_SCHEMA + e.what();
```
dcrankshaw: Same comments about the error messages as `query_frontend`

dcrankshaw (changes requested):
This looks good. Definitely an improvement over property trees! I made a few minor comments.

src/libclipper/include/clipper/json_util.hpp
giulio-zhou: Will do.

src/libclipper/include/clipper/json_util.hpp
giulio-zhou: The intended comment is listed below the vector declaration. Is the content of the comment clear enough? I can move the comment up and change it if necessary.

src/frontends/src/query_frontend.hpp
giulio-zhou: Added a `json_error_msg` formatting function that does that.

src/frontends/src/query_frontend.hpp
giulio-zhou: Fixed.

src/frontends/src/query_frontend.hpp
giulio-zhou: Fixed.

src/management/src/management_frontend.hpp
giulio-zhou: Fixed.

src/management/src/management_frontend.hpp
giulio-zhou: Same as above for `query_frontend`, added a formatting function.

src/libclipper/test/json_util_test.cpp
giulio-zhou: Actually, would you prefer that I changed each of these into string literals or just change the multi-line declarations?

src/libclipper/test/json_util_test.cpp
Corey-Zumar: This definitely looks cleaner than property trees, and it's intuitive. Nice!

src/libclipper/test/json_util_test.cpp
Corey-Zumar: Nice! Definitely clearer / easier to use than property trees

src/libclipper/test/json_util_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+TEST(JsonUtilTests, TestAddToNonObject) {
+  rapidjson::Document d;
+  d.SetArray();
+  ASSERT_THROW(add_int(d, "int_val", 3), json_semantic_error);
+}
+
+TEST(JsonUtilTests, TestParseCandidateModels) {
+  // Semantic checking to assert that model_version > 0?
```
Corey-Zumar: Did we plan on adding this? Remove comment if not

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+long get_long(rapidjson::Value& d, const char* key_name) {
+  rapidjson::Value& v =
+      check_kv_type_and_return(d, key_name, rapidjson::kNumberType);
+  if (!v.IsInt64()) {
+    throw json_semantic_error("Input of type " + kTypeNames[v.GetType()] +
+                              " is not of type long");
+  }
+  return (long)v.GetInt64();
```
Corey-Zumar: `return static_cast<long>(v.GetInt64())`

src/management/src/management_frontend.hpp
```diff
@@ -63,30 +90,21 @@ std::string lookup_selection_state(
...
-}
-
 void respond_http(std::string content, std::string message,
                   std::shared_ptr<HttpServer::Response> response) {
   *response << "HTTP/1.1 " << message
             << "\r\nContent-Length: " << content.length() << "\r\n\r\n"
             << content << "\n";
 }
 
+std::string json_error_msg(const std::string& exception_msg,
```
Corey-Zumar: Can we doc this to indicate that this message is used to return a user-facing error? This isn't clear by method name alone.

Corey-Zumar (approved):
Just a few comments - LGTM

src/libclipper/test/json_util_test.cpp
dcrankshaw: Anywhere that you need to use an excessive number of escape characters (for quotes, new lines, etc) you can just use a raw string literal.

giulio-zhou: Alright, will fix that shortly.

giulio-zhou: Fixed.

src/management/src/management_frontend.hpp
giulio-zhou: Cool, added a comment.

src/libclipper/include/clipper/json_util.hpp
giulio-zhou: Fixed.

dcrankshaw:
@giulio-zhou Fix conflicts with develop. Ping me once you do that and I'll merge.

dcrankshaw (approved):

src/libclipper/test/json_util_test.cpp
giulio-zhou: Decided to remove the comment. For now, we'll only handle formatting errors and basic semantic errors (e.g. type mismatches).

Issue #49, opened 2017-01-22
Why it exits in configure line 221?


sswv:
Is the "exit 0" in line 221 for testing only? Should it be removed? Thanks!
```
  # cd into the extracted directory and install
  cd cmake-*
  exit 0
```

dcrankshaw:
Yes that is a leftover from debugging. Thanks for pointing this out. #51 should fix this.

PR #51, opened 2017-01-23
Removes exit from configure when downloading cmake into local deps directory

dcrankshaw:
Somehow an extraneous `exit 0` snuck into our configure script (git blame says this is my fault) when downloading CMake to install locally, as pointed out in #49. This fixes the issue.

Test on Ubuntu 14.04 on a clean EC2 instance.

### DIFF
--- a/configure
+++ b/configure
@@ -218,7 +218,6 @@ if [ -z $CMAKE ]; then
   tar -xzvf cmake.tar.gz
   # cd into the extracted directory and install
   cd cmake-*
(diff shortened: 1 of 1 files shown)


dcrankshaw:
@giulio-zhou Can you approve this? It's a one line fix.

giulio-zhou (approved):

dcrankshaw:
Thanks @giulio-zhou

PR #53, opened 2017-01-25
Replace instances of std::shared_timed_mutex with std::mutex

Corey-Zumar:
Replaces `std::shared_timed_mutex` with `std::mutex` in metrics header and impl


giulio-zhou (approved):

dcrankshaw:
LGTM

PR #55, opened 2017-01-31
[CLIPPER-73] CIFAR SSH-only demo

dcrankshaw:


Corey-Zumar:
Password-protected SSH keys result in the user being prompted for password input when the Clipper Manager is instantiated. This results in a Jupyter Notebook crash resulting from an incompatible call to python's `get_pass()` method.

dcrankshaw:
@Corey-Zumar I think this is working again

examples/cifar_demo/rise_demo_frontend_dev.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   },
+   "outputs": [],
+   "source": [
+    "app_name = \"cifar_demo\"\n",
+    "candidate_models = [\n",
+    "    {\"model_name\": \"sklearn_cifar\", \"model_version\": 1},\n",
+    "    {\"model_name\": \"tf_cifar\", \"model_version\": 1},\n",
+    "]\n",
+    "\n",
+    "clipper.register_application(\n",
```
Corey-Zumar, line 150: Can we enumerate these arguments as key-value pairs for user clarity?
`selection_policy="EXP4"`, etc...

examples/cifar_demo/rise_demo_data_scientist.ipynb
```diff
@@ -0,0 +1,279 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "model_added = clipper.deploy_model(\n",
```
Corey-Zumar, line 105: enumerate these arguments as key-value pairs for user clarity

examples/cifar_demo/rise_demo_data_scientist.ipynb
```diff
@@ -0,0 +1,279 @@
...
```
Corey-Zumar, line 170: enumerate these arguments as key-value pairs for user clarity

examples/cifar_demo/rise_demo_data_scientist.ipynb
```diff
@@ -0,0 +1,279 @@
...
+    "    \"tf_cifar\",\n",
+    "    1,\n",
+    "    os.path.abspath(\"tf_cifar_model\"),\n",
+    "    # TODO: change test to latest\n",
```
Corey-Zumar: We should remove inline TODOs from the demo

examples/cifar_demo/rise_demo_data_scientist.ipynb
```diff
@@ -0,0 +1,279 @@
...
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "model_added = clipper.deploy_model(\n",
+    "    \"sklearn_cifar\",\n",
+    "    1,\n",
+    "    lr_model,\n",
+    "    # TODO: change test to latest\n",
```
Corey-Zumar: Remove inline TODO

containers/python/TensorFlowCifarDockerfile
```diff
@@ -0,0 +1,13 @@
+# TODO TODO TODO: Before merge, change :test to :latest
```
Corey-Zumar: Why are there 3 TODOs here?

containers/python/SklearnCifarDockerfile
```diff
@@ -0,0 +1,11 @@
+# TODO TODO TODO: Before merge, change :test to :latest
```
Corey-Zumar: Three TODOs?

examples/cifar_demo/README.md
```diff
@@ -0,0 +1,28 @@
+# Demo
+
+This demo walks the user through setting up Clipper, deploying models,
+and querying Clipper for predictions.
+
+## Setup
+
+Before you can run the demo, you must have access to a machine running
+Docker that you can SSH into. If running on EC2, you can use AMI `ami-3ba0f05b`,
+an Ubuntu image that has Docker and Docker-Compose installed.
```
Corey-Zumar, line 10: If we can get https://github.com/ucbrise/clipper/pull/60 merged, we should include information about running the demo locally. This seems like a very critical addition, as there is arguably much less overhead to installing Docker locally than launching AWS instances.

Corey-Zumar (changes requested):
Definitely an improvement in that the demo works as intended with the EXP4 selection policy. However, the following issues remain:
- The `get_pass()` issue is still occurring when attempting to use a password-protected SSH key with Jupyter Notebook 4.0.6
- If we expect users to experiment with different selection policies, the demo does not have compatible CIFAR binary prediction for EXP3, UCB, or EpsilonGreedy. Disregard this point if it's a nonissue for the demo.
- Nit: there's an extra, empty cell at the bottom of the `fronted_dev` notebook

I've left several other comments to address as well - most of them are minor

dcrankshaw:
I can't reproduce the `get_pass()` issue. It works fine with Jupyter 4.2.1. Given time constraints I'm inclined to ignore it for now.

examples/cifar_demo/rise_demo_frontend_dev.ipynb
dcrankshaw: That's what it was originally. But the only way to do that while also still having them be required is a hack with kwargs that I thought looked uglier. I think you're probably right though, I'm changing it back.

examples/cifar_demo/rise_demo_data_scientist.ipynb
dcrankshaw: These TODOs need to be addressed at the very end. `latest` is the docker build from the develop branch so it will break the demo until the PR gets merged. Holding off until we get everything else addressed.

examples/cifar_demo/README.md
dcrankshaw: Sounds good, but let's do that in the PR for #60.

dcrankshaw:
Fixed the issue preventing the other selection policies from working. Problem was [here](https://github.com/dcrankshaw/clipper/blob/cifar_ec2_demo/src/libclipper/src/selection_policies.cpp#L149). Changed it to return 0.0 in case of no predictions.

examples/cifar_demo/rise_demo_frontend_dev.ipynb
dcrankshaw: Hmm actually on second thought it makes the clipper_manager library much more confusing. [This StackOverflow answer](http://stackoverflow.com/a/5940228/814642) describes what I was doing before, and I think having all the functions just take kwargs is confusing. In the tutorial I'll make sure to explain what the arguments are.

containers/python/SklearnCifarDockerfile
dcrankshaw: Removed

containers/python/TensorFlowCifarDockerfile
dcrankshaw: Removed

examples/cifar_demo/rise_demo_data_scientist.ipynb
dcrankshaw: Removed

examples/cifar_demo/rise_demo_data_scientist.ipynb
dcrankshaw: Removed

management/clipper_manager.py
```diff
@@ -43,7 +43,7 @@
                 '--redis_ip=redis',
                 '--redis_port=%d' % REDIS_PORT],
             'depends_on': ['redis'],
-            'image': 'clipper/management_frontend:latest',
+            'image': 'clipper/management_frontend:test', # TODO: change back to latest before merge
```
dcrankshaw: removed todo

management/clipper_manager.py
```diff
@@ -52,7 +52,7 @@
             'depends_on': [
                 'redis',
                 'mgmt_frontend'],
-            'image': 'clipper/query_frontend:latest',
+            'image': 'clipper/query_frontend:test', # TODO: change back to latest before merge
```
dcrankshaw: removed todo

dcrankshaw:
@Corey-Zumar addressed your comments. Note that the demo won't work until it gets merged because I changed a bunch of docker files to from `test` to `latest`. This changes them from pulling the test version of the docker images (the ones corresponding to the code in this PR) to pulling the docker images from the code on `develop`.

examples/cifar_demo/README.md
Corey-Zumar: That makes more sense

examples/cifar_demo/rise_demo_frontend_dev.ipynb
Corey-Zumar: Sounds good, as long as the user is aware of the role of each argument

Corey-Zumar:
LGTM. I'll file an issue for `get_pass()` and see if I can find a fix and a way of reliably replicating it on another system.

Corey-Zumar (approved):

PR #56, opened 2017-01-31
[CLIPPER-78] Future compositions

dcrankshaw:
Replaces Boost future composition functions (`when_all`, `when_any`) with a more efficient version. The big issue with the Boost implementation was that it created a new thread for each call to one of the composition functions.

The API on these replacements is a little rough. It works fine for the specific situation in [`query_processor`](https://github.com/ucbrise/clipper/blob/2d521eb0922c3b431b4e1003c5022095cd88713d/src/libclipper/src/query_processor.cpp#L144-L147) where they are needed, but it's not very general purpose. Is it worth investing the time to craft a better API when it's unclear if there will ever be demand for it?

### DIFF
--- a/src/libclipper/include/clipper/future.hpp
+++ b/src/libclipper/include/clipper/future.hpp
@@ -1,0 +1,86 @@
+#ifndef CLIPPER_LIB_FUTURE_HPP
+#define CLIPPER_LIB_FUTURE_HPP
+
+#include <atomic>
+#include <cassert>
+#include <memory>
+#include <thread>
+#include <vector>
+
+#include "boost/thread.hpp"
+
+namespace clipper {
+
+namespace future {
+
+/// The first element in the pair is a future that will complete when
+/// all the futures provided have completed. The value of the future
+/// will always be true (there were some issues with a future<void>).
+/// The second element is a vector
+/// of futures that have the same values and will complete at the same time
+/// as the futures passed in as argument.
+template <class T>
+std::pair<boost::future<void>, std::vector<boost::future<T>>> when_all(
+    std::vector<boost::shared_future<T>> futures,
+    std::shared_ptr<std::atomic<int>> num_completed) {
+  if (futures.size() == 0) {
+    return std::make_pair(boost::make_ready_future(),
+                          std::vector<boost::future<T>>{});
+  }
+  int num_futures = futures.size();
+  auto completion_promise = std::make_shared<boost::promise<void>>();
+  std::vector<boost::future<T>> wrapped_futures;
+  for (auto f = futures.begin(); f != futures.end(); ++f) {
+    // PROBLEM: Passing a reference to an atomic (num_completed) but the atomic
+    // can go out of scope and be destructed
+    wrapped_futures.push_back(f->then(
+        [num_futures, completion_promise, num_completed](auto result) mutable {
+          if (*num_completed + 1 == num_futures) {
+            completion_promise->set_value();
+            //
+          } else {
+            *num_completed += 1;
+          }
+          return result.get();
+        }));
+  }
+
+  return std::make_pair<boost::future<void>, std::vector<boost::future<T>>>(
+      completion_promise->get_future(), std::move(wrapped_futures));
+}
+
+template <typename R>
+boost::future<R> wrap_when_any(
+    boost::future<R> f, std::shared_ptr<std::atomic<int>> completed,
+    std::shared_ptr<boost::promise<void>> completion_promise) {
+  return f.then([completion_promise, completed](auto result) mutable {
+
+    // Only set completed to true if the original value was false. This ensures
+    // that we only set the value once, and therefore only complete the promise
+    // once.
+    int num_finished = completed->fetch_add(1);
+    if (num_finished == 0) {
+      completion_promise->set_value();
+    }
+    return result.get();
+  });
+}
+
+template <typename R0, typename R1>
+std::tuple<boost::future<void>, boost::future<R0>, boost::future<R1>> when_any(
+    boost::future<R0> f0, boost::future<R1> f1,
+    std::shared_ptr<std::atomic<int>> completed) {
+  auto completion_promise = std::make_shared<boost::promise<void>>();
+  auto wrapped_f0 = wrap_when_any(std::move(f0), completed, completion_promise);
+  auto wrapped_f1 = wrap_when_any(std::move(f1), completed, completion_promise);
+
+  return std::make_tuple<boost::future<void>, boost::future<R0>,
+                         boost::future<R1>>(completion_promise->get_future(),
+                                            std::move(wrapped_f0),
+                                            std::move(wrapped_f1));
+}
+
+}  // namespace future
+}  // namespace clipper
+
+#endif  // define CLIPPER_LIB_FUTURE_HPP
--- a/src/libclipper/include/clipper/future.hpp
+++ b/src/libclipper/include/clipper/future.hpp
@@ -13,12 +13,18 @@ namespace clipper {
 
 namespace future {
 
-/// The first element in the pair is a future that will complete when
-/// all the futures provided have completed. The value of the future
-/// will always be true (there were some issues with a future<void>).
-/// The second element is a vector
-/// of futures that have the same values and will complete at the same time
-/// as the futures passed in as argument.
+/*
+ * The first element in the pair is a future that will complete when
+ * all the futures provided have completed. The second element is a vector
+ * of futures that have the same values and will complete at the same time
+ * as the futures passed in as argument.
+ *
+ * \param futures A list of futures to compose
+ * \param num_completed A counter indicating the number of futures that
+ * have completed so far. The reason this is an argument to the function
+ * instead of internal state is that the caller of `when_all` must make sure
+ * that the counter stays in scope for the lifetime of all the futures.
+*/
 template <class T>
 std::pair<boost::future<void>, std::vector<boost::future<T>>> when_all(
     std::vector<boost::shared_future<T>> futures,
@@ -31,8 +37,6 @@ std::pair<boost::future<void>, std::vector<boost::future<T>>> when_all(
   auto completion_promise = std::make_shared<boost::promise<void>>();
   std::vector<boost::future<T>> wrapped_futures;
   for (auto f = futures.begin(); f != futures.end(); ++f) {
-    // PROBLEM: Passing a reference to an atomic (num_completed) but the atomic
-    // can go out of scope and be destructed
     wrapped_futures.push_back(f->then(
         [num_futures, completion_promise, num_completed](auto result) mutable {
           if (*num_completed + 1 == num_futures) {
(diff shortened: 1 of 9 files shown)


src/libclipper/include/clipper/future.hpp
```diff
@@ -0,0 +1,105 @@
...
+        }));
+  }
+
+  return std::make_pair<boost::future<void>, std::vector<boost::future<T>>>(
+      completion_promise->get_future(), std::move(wrapped_futures));
+}
+
+template <typename R>
+boost::future<R> wrap_when_any(
+    boost::future<R> f, std::shared_ptr<std::atomic<int>> completed,
```
Corey-Zumar: Nit: can we change `f` to `future` and `completed` to `num_completed`?

src/libclipper/include/clipper/future.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * instead of internal state is that the caller of `when_all` must make sure
+ * that the counter stays in scope for the lifetime of all the futures.
+ *
+ * \return A tuple of three futures. The first is the composed future that will
+ * complete when either of the two futures provided as argument completes. The
+ * second two elements in the tuple are futures that will complete at the same
+ * time and with the same value as the futures passed in as arguments.
+ */
+template <typename R0, typename R1>
+std::tuple<boost::future<void>, boost::future<R0>, boost::future<R1>> when_any(
```
Corey-Zumar: Is there a reason that this function only takes two futures? It seems that this could be extended to a variable number of futures.

src/libclipper/src/query_processor.cpp
```diff
@@ -212,46 +219,50 @@ boost::future<FeedbackAck> QueryProcessor::update(FeedbackQuery feedback) {
...
 
   // copy the vector
-  vector<shared_future<Output>> predict_task_completion_futures =
+  vector<boost::shared_future<Output>> predict_task_completion_futures =
       task_executor_.schedule_predictions({predict_tasks});
 
   vector<boost::future<FeedbackAck>> feedback_task_completion_futures =
       task_executor_.schedule_feedback(std::move(feedback_tasks));
 
+  // TODO: replace with clipper::future implementation of when_all
```
Corey-Zumar: Why is this a TODO?

src/libclipper/include/clipper/future.hpp
Corey-Zumar: Or, if this value is either `0` or `1`, could we use a `bool` instead?

src/libclipper/src/datatypes.cpp
```diff
@@ -143,8 +139,10 @@ size_t FloatVector::serialize(uint8_t *buf) const {
 InputType FloatVector::type() const { return InputType::Floats; }
 
 size_t FloatVector::hash() const {
-  // TODO [CLIPPER-63]: Find an alternative to hashing floats directly, as this is
-  // generally a bad idea due to loss of precision from floating point representations
+  // TODO [CLIPPER-63]: Find an alternative to hashing floats directly, as this
+  // is
```
Corey-Zumar: Fix doc formatting

src/libclipper/src/datatypes.cpp
```diff
@@ -163,8 +161,10 @@ size_t DoubleVector::serialize(uint8_t *buf) const {
 }
 
 size_t DoubleVector::hash() const {
-  // TODO [CLIPPER-63]: Find an alternative to hashing doubles directly, as this is
-  // generally a bad idea due to loss of precision from floating point representations
+  // TODO [CLIPPER-63]: Find an alternative to hashing doubles directly, as this
+  // is
```
Corey-Zumar: Fix doc formatting

src/libclipper/include/clipper/future.hpp
```diff
@@ -0,0 +1,105 @@
...
+    std::shared_ptr<std::atomic<int>> num_completed) {
+  if (futures.size() == 0) {
+    return std::make_pair(boost::make_ready_future(),
+                          std::vector<boost::future<T>>{});
+  }
+  int num_futures = futures.size();
+  auto completion_promise = std::make_shared<boost::promise<void>>();
+  std::vector<boost::future<T>> wrapped_futures;
+  for (auto f = futures.begin(); f != futures.end(); ++f) {
+    wrapped_futures.push_back(f->then(
```
Corey-Zumar, line 40: Based on my understanding, it seems that invoking the continuation function `then` results in the creation of an additional thread when the future `f` is complete. This thread executes the supplied method and is quickly destroyed.

If we were to execute `when_all` on a large number of futures that all complete around the same period of time, wouldn't we be prone to a large spike in thread creations that may still pose a problem in the face of a ulimit? The alternative, inefficient boost implementation seems to guarantee that only a single additional thread is created per batch of futures, whereas we'd create X threads for a call to the new `when_all()` implementation with X futures supplied, albeit at different times. 

Please let me know if there's something I'm missing / provide your assessment of this tradeoff if it's accurate.

src/libclipper/src/query_processor.cpp
Corey-Zumar: Additionally, there's an invocation of `boost::when_all()` on line 313 without a TODO.

src/libclipper/src/query_processor.cpp
```diff
@@ -125,79 +126,85 @@ future<Response> QueryProcessor::predict(Query query) {
...
-  ](auto result_future) mutable {
-    log_info(LOGGING_TAG_QUERY_PROCESSOR, "ENTERED CONTINUATION LAMBDA");
-
-    auto result = result_future.get();
-    log_info(LOGGING_TAG_QUERY_PROCESSOR, std::boolalpha);
-    log_info_formatted(
-        LOGGING_TAG_QUERY_PROCESSOR, "All tasks finished: ", std::get<0>(result).is_ready());
-    log_info_formatted(LOGGING_TAG_QUERY_PROCESSOR, "Timer Fired: ", std::get<1>(result).is_ready());
+  response_ready_future.then([
+    query, query_id, p = std::move(promise), s = std::move(serialized_state),
```
Corey-Zumar: nit: use more descriptive variable names than `p` and `s`

Corey-Zumar (changes requested):
The implementation looks good overall. My only concern is whether or not the performance is actually substantially better. Please see the comment regarding `future->then()`

src/libclipper/include/clipper/future.hpp
dcrankshaw: So you raise a good point. It's a little confusing and the documentation isn't great. As far as I can tell, `future::then` does not create a new thread. I believe this is dependent on the default [`launch_policy`](http://www.boost.org/doc/libs/1_63_0/doc/html/thread/synchronization.html#thread.synchronization.futures.reference.unique_future.then). It appears from the documentation that the default launch policy is `launch::none` which means that the continuation is called on an unspecified thread of execution. In practice, when running this on my laptop at least, no new threads are created by the call to `future::then`. On the other hand, `when_any` and `when_all` definitely create new threads.

Here's the experiment I used to verify this, you can try running it yourself. Check out [`dcrankshaw/scheduler-prototype`](https://github.com/dcrankshaw/clipper/tree/scheduler-prototype). In `query_processor.cpp`, there are [two versions of `QueryProcessor::predict`](https://github.com/dcrankshaw/clipper/blob/34a93051d670b7c78563608bd2e9c7e47cf460ad/src/libclipper/src/query_processor.cpp#L105-L285). The uncommented one uses the Clipper implementations of the `when_all` and `when_any`. The commented out version uses the Boost implementations. Neither version specifies a launch policy for the calls to `future::then`. Included in the branch is a benchmark. You can run the benchmark on each version of the predict function and you'll see that the Boost implementation has worse performance and crashes when the thread ulimit is reached.

__Here are instructions for running the benchmark:__

Start the containers the benchmark expects (two noop containers):
```bash
cd containers/python
./start_containers
```

Build and run the benchmark:
```bash
./configure --release && cd release
make bench
./src/frontends/bench
# It will prompt you for the number of requests to make. Try 100000.
```

When you're done with the experiment, Ctrl-C the containers script and it will stop the container processes for you.

src/libclipper/include/clipper/future.hpp
dcrankshaw: Changed it to an `atomic_flag`

src/libclipper/include/clipper/future.hpp
dcrankshaw: It only takes 2 futures because I only needed it to take 2 futures. I'm sure there's a way with template magic, but I'm not sure how to extend it to take a variable number of futures, each with a different type.

src/libclipper/src/query_processor.cpp
dcrankshaw: Ahh it was a TODO because I hadn't done it yet. Removed all uses of boost future composition functions.

src/libclipper/src/datatypes.cpp
dcrankshaw: Done

src/libclipper/src/datatypes.cpp
dcrankshaw: Done

dcrankshaw:
@Corey-Zumar I addressed the rest of your comments. Try out that benchmark and see if you see the same results as me.

src/libclipper/include/clipper/future.hpp
Corey-Zumar: Thanks for the explanation. Your benchmark is definitely a compelling demonstration of the improvement made by using the new `when_all()` and `when_any()` implementations. I found that the time to completion for 100000 requests with the new implementation is roughly half the time for the same number of requests with the boost implementation.

Corey-Zumar:
@dcrankshaw The benchmark produces dramatic performance differences that favor the new implementation, and the use of the old boost implementation results in an indication that, on average, zero tasks were completed for 100000 requests (not sure if this is related - signifies a crash?). LGTM.

Corey-Zumar (approved):

src/libclipper/include/clipper/future.hpp
```diff
@@ -0,0 +1,150 @@
...
+        [num_futures, completion_promise, num_completed](auto result) mutable {
```
mwhittaker, line 41: Is this continuation run on multiple threads concurrently? If so, it seems like there is a race condition when accessing `num_completed`. Maybe you want to use something like [fetch_add](http://en.cppreference.com/w/cpp/atomic/atomic/fetch_add)?

src/libclipper/include/clipper/future.hpp
```diff
@@ -0,0 +1,150 @@
...
+      completion_promise->get_future(), std::move(wrapped_futures));
+}
+
+template <typename R>
+boost::future<R> wrap_when_both(
+    boost::future<R> future, std::shared_ptr<std::atomic<int>> num_completed,
+    std::shared_ptr<boost::promise<void>> completion_promise) {
+  return future.then([completion_promise, num_completed](auto result) mutable {
+
+    int num_already_completed = num_completed->fetch_add(1);
```
mwhittaker, line 62: Oh, you use `fetch_add` here!

src/libclipper/include/clipper/future.hpp
dcrankshaw: Yeah you're right, there is a race condition. Good catch!

(Documenting here to be explicit)

Consider the case where there are 2 futures (so `num_futures = 2`) and neither has completed yet.

They may both be completed simultaneously in separate threads. They will both fail the test in the if statement because `num_completed + 1 < num_futures` (0 + 2 < 2). Then _after the test_ they will both atomically update `num_completed`. The increment to `num_completed` will be safe because it's an atomic add, but now both futures have completed so the future composition should be marked as finished, and yet the `completion_promise` will never be set.

The fix is simple:
replace the if-else statement with:

```
if (num_completed.fetch_add(1) + 1== num_futures) {
    completion_promise->set_value();
}
```

Consider the previous example with two futures. Because `num_completed` is atomic, the two `fetch_add` calls will be serial. One of them is guaranteed to be the second call, in which case `num_completed.fetch_add(1)` returns 1 (the previous value of the atomic variable), so the conditional will always evaluate to true exactly once.

dcrankshaw: The repeated check of an atomic for a single condition is a total concurrency code smell. I have no idea why I wrote the check that way lol.

Btw, the C++ arithmetic operators are [overloaded](http://en.cppreference.com/w/cpp/atomic/atomic/operator_arith2) for atomics, so the update to the atomic variable itself is safe.

PR #57, opened 2017-02-03
[CLIPPER-84] Fix the docker builds

dcrankshaw:
The addition of the spdlog and rapidjson libraries broke the build for Docker (alpine linux with musl) and on Ubuntu with gcc.

This PR fixes those problems.

### DIFF
--- a/ManagementFrontendDockerfile
+++ b/ManagementFrontendDockerfile
@@ -3,6 +3,8 @@ FROM frolvlad/alpine-gxx
 COPY ./ /clipper
 
 RUN apk add --no-cache git bash make boost-dev cmake libev-dev hiredis-dev zeromq-dev \
+    && cd /clipper/src/libs/spdlog \
+    && git apply ../patches/make_spdlog_compile_linux.patch \
     && cd /clipper \
     && ./configure --cleanup-quiet \
     && ./configure --release \
--- a/src/libs/patches/make_spdlog_compile_linux.patch
+++ b/src/libs/patches/make_spdlog_compile_linux.patch
@@ -1,0 +1,25 @@
+From c0fcb37692da01049b4acfd92a779738c14d32ff Mon Sep 17 00:00:00 2001
+From: Dan Crankshaw <dscrankshaw@gmail.com>
+Date: Thu, 2 Feb 2017 15:05:56 -0800
+Subject: [PATCH] make compile on alpine
+
+---
+ include/spdlog/details/os.h | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/include/spdlog/details/os.h b/include/spdlog/details/os.h
+index b63ce66..dae02f2 100644
+--- a/include/spdlog/details/os.h
++++ b/include/spdlog/details/os.h
+@@ -378,7 +378,7 @@ inline std::string errno_str(int err_num)
+         return "Unkown error";
+ 
+ #elif defined(__FreeBSD__) || defined(__APPLE__) || defined(ANDROID) || defined(__SUNPRO_CC) || \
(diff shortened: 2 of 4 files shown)


dcrankshaw:
@giulio-zhou Can you review this?

giulio-zhou:
Yeah, I'll take a look.

ManagementFrontendDockerfile
```diff
@@ -3,6 +3,8 @@ FROM frolvlad/alpine-gxx
 COPY ./ /clipper
 
 RUN apk add --no-cache git bash make boost-dev cmake libev-dev hiredis-dev zeromq-dev \
+    && cd /clipper/src/libs/spdlog \
+    && git apply ../patches/make_spdlog_compile_linux.patch \
```
dcrankshaw, line 7: I didn't want to fork spdlog just for this tiny change for the Docker builds, so I just created a patch that we can apply.

src/libs/patches/make_spdlog_compile_linux.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/include/spdlog/details/os.h b/include/spdlog/details/os.h
+index b63ce66..dae02f2 100644
+--- a/include/spdlog/details/os.h
++++ b/include/spdlog/details/os.h
+@@ -378,7 +378,7 @@ inline std::string errno_str(int err_num)
+         return "Unkown error";
+ 
+ #elif defined(__FreeBSD__) || defined(__APPLE__) || defined(ANDROID) || defined(__SUNPRO_CC) || \
+-      ((_POSIX_C_SOURCE >= 200112L) && ! defined(_GNU_SOURCE)) // posix version
++      ((_POSIX_C_SOURCE >= 200112L) && ! defined(_GNU_SOURCE)) || !defined(__GLIBC__)   // posix version
```
Corey-Zumar, line 19: What does the addition of `|| !defined(__GLIBC__)` do?

src/libs/patches/make_spdlog_compile_linux.patch
```diff
@@ -0,0 +1,25 @@
...
+Subject: [PATCH] make compile on alpine
+
+---
+ include/spdlog/details/os.h | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/include/spdlog/details/os.h b/include/spdlog/details/os.h
+index b63ce66..dae02f2 100644
+--- a/include/spdlog/details/os.h
++++ b/include/spdlog/details/os.h
```
Corey-Zumar, line 13: What's the difference between the`a` and `b` directories?

Corey-Zumar (approved):
LGTM. Just a couple of learning questions about what was done to fix this.

src/libs/patches/make_spdlog_compile_linux.patch
dcrankshaw: The Docker images for Clipper are built on Alpine Linux which is a minimal Linux distribution. This is a standard approach for making Docker images smaller because otherwise they get very bloated very fast. But the C standard library implementation distributed with Alpine is [musl](http://wiki.musl-libc.org/wiki/Main_Page) not glibc (the GNU C standard library). I'm not sure exactly what the differences are in this particular piece of code, but I found this fix in an [open PR](https://github.com/gabime/spdlog/pull/283/commits/c4fa9cb603494699d4ffbf20f69e88c27b26192a) to spdlog after some googling.

src/libs/patches/make_spdlog_compile_linux.patch
dcrankshaw: This is just the format for a git patch file (generated with `git format-patch`).

PR #61, opened 2017-02-05
removed output type

dcrankshaw:


Corey-Zumar:
Can you expand on the bug caused by `output_type`? I'm not sure if this is related, but I've been having trouble serving CIFAR predictions with the latest management and query docker frontends.

dcrankshaw:
It wasn't a bug, the output_type just distinguished between double and int outputs. But that's not a useful distinction to make in Clipper currently, and it added a lot of needless complexity to the codebase. It's definitely possible that the removal broke something though (this is why we need integration tests!). What was the problem?

dcrankshaw:
Folded this in with #55

PR #68, opened 2017-02-17
[CLIPPER-65] Performance / Resilience Test Based on CIFAR Demo

Corey-Zumar:


Corey-Zumar:
@dcrankshaw

dcrankshaw:
jenkins test this please

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+			test_x, test_y, train_x, train_y = self.get_cifar_data()
+			print("Training model...")
+			model = lm.LogisticRegression()
+			model.fit(train_x, train_y)
+			joblib.dump(model, model_location)
+			print("Model trained!")
+			print("Logistic Regression test score: %f" % model.score(test_x, test_y))
+
+if __name__ == '__main__':
+	setup = BenchSetup("localhost", "../examples/cifar_demo")
```
dcrankshaw: Can you have the location of the CIFAR dataset be a command line argument?

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+	def get_cifar_data(self):
+		train_x, train_y = cifar_utils.filter_data(
+			*cifar_utils.load_cifar(self.cifar_dir_path, cifar_filename="cifar_train.data", norm=False))
+		test_x, test_y = cifar_utils.filter_data(
+			*cifar_utils.load_cifar(self.cifar_dir_path, cifar_filename="cifar_test.data", norm=False))
+
+		return test_x, test_y, train_x, train_y
+
+	def train_sklearn_model(self):
+		model_location = BASE_DATA_PATH + SKLEARN_MODEL_FILE
```
dcrankshaw: use `os.path.join()` not string concatenation

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
+import sys
+import os
+import json
+import requests
```
dcrankshaw: json and requests imports aren't used

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+			model = lm.LogisticRegression()
+			model.fit(train_x, train_y)
+			joblib.dump(model, model_location)
+			print("Model trained!")
+			print("Logistic Regression test score: %f" % model.score(test_x, test_y))
+
+if __name__ == '__main__':
+	setup = BenchSetup("localhost", "../examples/cifar_demo")
+	setup.run()
+
```
dcrankshaw: Get rid of the tabs in this file

dcrankshaw:
jenkins, test this please

dcrankshaw:
@Corey-Zumar Looks like there's a problem with Jenkins, don't worry about the failing tests for now. I'm reviewing this right now.

shaneknapp:
test this please

shaneknapp:
test this please

shaneknapp:
test this please

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+import json
+import requests
+sys.path.append("..")
+
+from management import clipper_manager
+from examples.tutorial import cifar_utils
+from sklearn import linear_model as lm
+from sklearn.externals import joblib
+from fabric.api import *
+
```
dcrankshaw: In order to get the imports to work correctly, I added the following files
```
management/__init__.py
examples/__init__.py
examples/tutorial/__init__.py
```

Then I changed the imports to the following:
```py
sys.path.append(os.path.abspath("../management"))
sys.path.append(os.path.abspath("../examples"))

import clipper_manager
from tutorial import cifar_utils
from sklearn import linear_model as lm
from sklearn.externals import joblib
from fabric.api import *
```

dcrankshaw: We will want to rethink directory organization in the future, but for now this should work. I don't want an __init__.py file in the root of the source code.

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+
+APP_NAME = "bench"
+BASE_DATA_PATH = "data/"
```
dcrankshaw: This is confusingly named, because this is where the model, not the data, will be stored.

Change it to:

```py
BASE_MODEL_PATH = model/"
```

bench/bench_init.py
```diff
@@ -0,0 +1,63 @@
...
+		if os.path.isfile(model_location):
+			model = joblib.load(model_location)
+			print("Found and loaded model!")
+		else:
+			print("Loading CIFAR data...")
+			test_x, test_y, train_x, train_y = self.get_cifar_data()
+			print("Training model...")
+			model = lm.LogisticRegression()
+			model.fit(train_x, train_y)
+			joblib.dump(model, model_location)
```
dcrankshaw: This will fail if `BASE_DATA_PATH` doesn't exist yet. Add a check to see if the directory exists, and if it doesn't, create it.

src/frontends/src/bench.cpp
```diff
@@ -3,21 +3,159 @@
...
+  std::cout << "Enter the delay between batches, in milliseconds: ";
+  std::cin >> batch_delay;
+
+  config.emplace(CONFIG_KEY_PATH, path);
+  config.emplace(CONFIG_KEY_NUM_THREADS, num_threads);
+  config.emplace(CONFIG_KEY_NUM_BATCHES, num_batches);
+  config.emplace(CONFIG_KEY_BATCH_SIZE, batch_size);
+  config.emplace(CONFIG_KEY_BATCH_DELAY, batch_delay);
+  return config;
+};
```
dcrankshaw, line 161: Reading from the command-line is really nice for one-off benchmarks but impossible to script.

Can you a command line arg `-f FILENAME.json` that reads a JSON config file to get the benchmark parameters. We already have a command line options library in the codebase, check out [`query_frontend_main.cpp`](https://github.com/ucbrise/clipper/blob/35cb8ee0a9fe8f8ea8601d06b8cc8adcfe2fc227/src/frontends/src/query_frontend_main.cpp#L2-L18) for an example on how to use it.

The behavior should be to first parse command line args. If the filename arg is present, read the JSON file to get the benchmark params. If it's not present, use the existing approach of asking the user interactively.

src/frontends/src/bench.cpp
```diff
@@ -3,21 +3,159 @@
...
+constexpr double SKLEARN_BIRD_LABEL = 0;
+
+std::unordered_map<int, std::vector<std::vector<double>>> load_cifar(
+    std::unordered_map<std::string, std::string>& config) {
+  std::ifstream cifar_file(config.find(CONFIG_KEY_PATH)->second, std::ios::binary);
+  std::istreambuf_iterator<char> cifar_data(cifar_file);
+  std::unordered_map<int, std::vector<std::vector<double>>> vecs_map;
+  for (int i = 0; i < 10000; i++) {
+    int label = static_cast<int>(*cifar_data);
+    std::cout << label << std::endl;
```
dcrankshaw: Don't print the label here.

src/frontends/src/bench.cpp
```diff
@@ -3,21 +3,159 @@
...
+      vecs_map.emplace(label, new_label_vecs);
+    }
+  }
+  return vecs_map;
+}
+
+void send_predictions(std::unordered_map<std::string, std::string>& config, QueryProcessor &qp,
+                      std::unordered_map<int, std::vector<std::vector<double>>> &cifar_data,
+                      std::shared_ptr<metrics::RatioCounter> accuracy_ratio) {
+  std::vector<std::vector<double>> planes_vecs = cifar_data.find(0)->second;
```
dcrankshaw: I get a segfault at this line. I'm guessing something is wrong with loading the cifar data?

Let me know if you can't reproduce it, and I can dig into it a little more.

dcrankshaw (changes requested):
This is a good start. There's a bug in the benchmark code though, so I haven't been able to actually get through the full benchmark.

Also, can you format and lint the Python code to make sure it conforms with PEP8?

bench/bench_init.py
Corey-Zumar: Replaced imports with the one's you've defined.

bench/bench_init.py
Corey-Zumar: Removed

bench/bench_init.py
Corey-Zumar: Agreed. Changed this!

bench/bench_init.py
Corey-Zumar: Done

bench/bench_init.py
Corey-Zumar: Done.

bench/bench_init.py
Corey-Zumar: Done. Also modified `setup_bench.sh` to accept this path as a command line argument.

src/frontends/src/bench.cpp
Corey-Zumar: I think the instructions I provided for running the benchmark are insufficient / unclear. For the path to the CIFAR binary data set, I'm passing a path to a file within the **binary** version of CIFAR hosted at https://www.cs.toronto.edu/~kriz/cifar.html . I'm using `data_batch_1.bin`. I should have stated this explicitly.

If you were passing a path to analogous file (`data_batch_<n>.bin`) within the binary data set and still received the seg fault, please let me know.

src/frontends/src/bench.cpp
Corey-Zumar: Done. We can now specify a path to a JSON configuration file with the command line argument `-f ` or `--filename`. The configuration file should be formatted as follows:

```
{
   "cifar_data_path":"<path to a data_batch_<n>.bin file in the cifar binary data set>",
   "num_threads":"<number of threads>",
   "num_batches":"<number of batches>",
   "batch_size":"<batch size>",
   "batch_delay_millis":"<batch delay millis>"
}
```

dcrankshaw:
@Corey-Zumar This is reviewed and ready for you to address the comments

src/frontends/src/bench.cpp
Corey-Zumar: Removed this line

dcrankshaw:
jenkins test this please

Corey-Zumar:
Formatted `bench_init.py` to conform to PEP8

dcrankshaw:
jenkins test this please

bench/setup_bench.sh
```diff
@@ -0,0 +1,13 @@
+trap "exit" INT TERM
+trap "kill 0" EXIT
+
+echo -n "Enter path to the CIFAR python data set: "
+read cifar_path
```
dcrankshaw: Don't interactively ask for input, just take the location as a command line param:

```
cifar_path=$1
```

src/frontends/src/bench.cpp
```diff
@@ -3,21 +3,199 @@
...
+      send_predictions(test_config, qp, cifar_data, accuracy_ratio);
+    });
+    threads.push_back(std::move(thread));
+  }
+  for (auto &thread : threads) {
+    thread.join();
+  }
+  std::string metrics = metrics::MetricsRegistry::get_metrics().report_metrics();
+  log_info("BENCH", metrics);
+  std::this_thread::sleep_for(std::chrono::seconds(5));
```
dcrankshaw: Why are you sleeping at the end here?

dcrankshaw (changes requested):
Looks good. A couple minor comments.

Can you add some documentation about how to run the benchmark? At a minimum, you need to specify what the dataset for the model container should be and what the dataset for the bench binary should be.

Also, I found a bug in the RPCService destructor. 


Replace [`RPCService::shutdown_service`](https://github.com/ucbrise/clipper/blob/35cb8ee0a9fe8f8ea8601d06b8cc8adcfe2fc227/src/libclipper/src/rpc_service.cpp#L130-L133) with
```cpp
void RPCService::shutdown_service(socket_t &socket) {
  size_t buf_size = 32;
  char buf[buf_size];
  socket.getsockopt(ZMQ_LAST_ENDPOINT, (void *)&buf, &buf_size);
  std::string last_endpoint = std::string(buf);
  socket.unbind(last_endpoint);
  socket.close();
 }
```

The problem is we were calling `disconnect` not `unbind`. The mess with ZMQ_LAST_ENDPOINT is because we are using wildcard addresses (http://api.zeromq.org/4-2:zmq-unbind#toc3).

dcrankshaw:
jenkins test this please

bench/setup_bench.sh
Corey-Zumar: Done

src/frontends/src/bench.cpp
Corey-Zumar: I shouldn't be. I think this was an illogical attempt to log metrics in the face of the RPC error you identified above. Removed.

Corey-Zumar (commented):
Added a README with a list of required files and a detailed explanation of how to run the test. I also integrated your fix for the RPC error that you identified - thanks for finding that - and addressed your other comments.

bench/README.md
```diff
@@ -0,0 +1,56 @@
+# Running the performance benchmark
+
+## Required Files
+### CIFAR10 Python Dataset for SKLearn Model
+This benchmark serves an SKLearn model that depends on the CIFAR10 Python dataset for training. This dataset can be obtained from [https://www.cs.toronto.edu/~kriz/cifar.html](https://www.cs.toronto.edu/~kriz/cifar.html).
```
dcrankshaw: This actually isn't true. The `cifar_utils` functions expect a CSV. You can use https://github.com/ucbrise/clipper/blob/develop/examples/tutorial/download_cifar.py to download the dataset and transform it into a CSV. If you look at the [tutorial](https://github.com/ucbrise/clipper/blob/develop/examples/tutorial/tutorial_part_one.ipynb), we first download the dataset and create the CSV, then load it to train the model.

dcrankshaw (commented):
Looks good other than some confusion in the docs about the CIFAR python dataset.

bench/README.md
Corey-Zumar: Thanks! Fixed this. Please take another look.

dcrankshaw (approved):

dcrankshaw:
jenkins test this please

PR #69, opened 2017-02-19
Add Java example client

yunshengb:
Add a basic Java client in examples/java. 
Rename examples/basic_query to examples/python.
Update README to reflect the changes above.


dcrankshaw:
A Clipper client is just a REST client. I'm not sure that we need examples in multiple languages.

PR #79, opened 2017-03-06
[CLIPPER-92] Change RPC to event-based message handling

Corey-Zumar:
Notes:

* The implementation differs slightly from the specification included in the [JIRA issue](https://clipper.atlassian.net/browse/CLIPPER-92). Notably, the `on_new_connection` callback is not passed from the task executor to the RPC service. Rather, this method is invoked via the task executor's Redis state subscriber when the addition of a new container is detected.

### DIFF
--- a/src/libclipper/include/clipper/containers.hpp
+++ b/src/libclipper/include/clipper/containers.hpp
@@ -87,8 +87,8 @@ class ActiveContainers {
   // the queues. The queues are independently threadsafe.
   boost::shared_mutex m_;
 
-  // A map of models to their replica task queues. The replicas
-  // for each model are represented as a map keyed on replica id
+  // A mapping of models to their replica task queues. The replicas
+  // for each model are represented as a map keyed on replica id.
   std::unordered_map<VersionedModelId,
                      std::map<int, std::shared_ptr<ModelContainer>>,
                      decltype(&versioned_model_hash)>
--- a/src/libclipper/include/clipper/containers.hpp
+++ b/src/libclipper/include/clipper/containers.hpp
@@ -55,7 +55,7 @@ class ActiveContainers {
   ActiveContainers(ActiveContainers &&) = default;
   ActiveContainers &operator=(ActiveContainers &&) = default;
 
-  void add_container(VersionedModelId model, int id, InputType input_type);
+  void add_container(VersionedModelId model, int connection_id, int replica_id, InputType input_type);
 
   /// TODO: this method should be deprecated / removed when per-model
   /// queueing is implemented, as it currently functions as an efficiency
--- a/src/libclipper/include/clipper/containers.hpp
+++ b/src/libclipper/include/clipper/containers.hpp
@@ -87,6 +87,8 @@ class ActiveContainers {
   // the queues. The queues are independently threadsafe.
   boost::shared_mutex m_;
 
+  // A map of models to their replica task queues. The replicas
+  // for each model are represented as a map keyed on replica id
   std::unordered_map<VersionedModelId,
                      std::map<int, std::shared_ptr<ModelContainer>>,
                      decltype(&versioned_model_hash)>
--- a/src/libclipper/include/clipper/containers.hpp
+++ b/src/libclipper/include/clipper/containers.hpp
@@ -57,6 +57,10 @@ class ActiveContainers {
 
   void add_container(VersionedModelId model, int id, InputType input_type);
 
+  /// TODO: this method should be deprecated / removed when per-model
+  /// queueing is implemented, as it currently functions as an efficiency
+  /// method in the context of assigning tasks to containers directly
+  ///
   /// This method returns a vector of all the active containers (replicas)
   /// of the specified model. This is threadsafe because each individual
   /// ModelContainer object is threadsafe, and this method returns
--- a/src/libclipper/include/clipper/containers.hpp
+++ b/src/libclipper/include/clipper/containers.hpp
@@ -67,6 +67,13 @@ class ActiveContainers {
   std::vector<std::shared_ptr<ModelContainer>> get_model_replicas_snapshot(
       const VersionedModelId &model);
 
+  /// This method returns the active container specified by the
+  /// provided model id and replica id. This is threadsafe because each
+  /// individual ModelContainer object is threadsafe, and this method returns
+  /// a shared_ptr to a ModelContainer object.
+  std::shared_ptr<ModelContainer> get_model_replica(
+      const VersionedModelId &model, const int replica_id);
+
   /// Get list of all models that have at least one active replica.
   std::vector<VersionedModelId> get_known_models();
 
@@ -76,9 +83,8 @@ class ActiveContainers {
   // the queues. The queues are independently threadsafe.
   boost::shared_mutex m_;
 
-  // Each queue corresponds to a single model container.
   std::unordered_map<VersionedModelId,
-                     std::vector<std::shared_ptr<ModelContainer>>,
+                     std::map<int, std::shared_ptr<ModelContainer>>,
                      decltype(&versioned_model_hash)>
       containers_;
 };
--- a/src/libclipper/include/clipper/rpc_service.hpp
+++ b/src/libclipper/include/clipper/rpc_service.hpp
@@ -33,7 +33,7 @@ using RPCRequest =
     std::tuple<const int, const int, const std::vector<std::vector<uint8_t>>,
                const long>;
 
-template <class T>
+template<class T>
 class VectorHasher {
  public:
   size_t operator()(const vector<T> &v) const {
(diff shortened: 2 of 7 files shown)


dcrankshaw:
jenkins test this please

src/libclipper/include/clipper/containers.hpp
```diff
@@ -76,9 +87,8 @@ class ActiveContainers {
   // the queues. The queues are independently threadsafe.
   boost::shared_mutex m_;
 
-  // Each queue corresponds to a single model container.
   std::unordered_map<VersionedModelId,
-                     std::vector<std::shared_ptr<ModelContainer>>,
+                     std::map<int, std::shared_ptr<ModelContainer>>,
```
dcrankshaw: What's the key of the map?

src/libclipper/include/clipper/rpc_service.hpp
```diff
@@ -31,6 +32,14 @@ using RPCResponse = std::pair<const int, vector<uint8_t>>;
...
 std::tuple<const int, const int, const std::vector<std::vector<uint8_t>>, const long>;
 
+template <class T>
+class VectorHasher {
+ public:
+  size_t operator()(const vector<T> &v) const {
+    return boost::hash_range(v.begin(), v.end());
+  }
+};
+
```
dcrankshaw: Don't we already have a method somewhere to hash vectors? For the vector input types to make a hash key?

src/libclipper/include/clipper/rpc_service.hpp
```diff
@@ -84,6 +99,9 @@ class RPCService {
   std::unordered_map<VersionedModelId, int, decltype(&versioned_model_hash)>
       replica_ids_;
   std::shared_ptr<metrics::Histogram> msg_queueing_hist;
```
dcrankshaw: Nit: the `rpc_thread` and `msg_queuing_hist` member variables are missing an underscore postfix. Can you add that?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -98,6 +100,13 @@ class TaskExecutor {
             active_containers_->add_container(
                 vm, std::stoi(container_info["zmq_connection_id"]),
                 parse_input_type(container_info["input_type"]));
+            int replica_id = std::stoi(container_info["model_replica_id"]);
+            // TODO: These callbacks should be submitted to a thread pool
+            // for execution on separate threads
+            on_new_container_added(vm, replica_id);
+            std::thread([&]() {
```
dcrankshaw: Don't create a new thread, just execute the callback directly.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -158,6 +165,7 @@ class TaskExecutor {
   redox::Redox redis_connection_;
   redox::Subscriber redis_subscriber_;
   bool active_ = false;
+  const int max_batch_size_ = 5;
```
dcrankshaw: Can you stick a TODO in here noting that we will replace the static fixed batch size with dynamic batching?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -177,78 +185,64 @@ class TaskExecutor {
...
-    int max_responses = 10;
-    while (active_) {
-      auto responses = rpc_->try_get_responses(max_responses);
-      if (responses.empty()) {
-        continue;
-      }
-      std::unique_lock<std::mutex> l(inflight_messages_mutex_);
-      for (auto r : responses) {
-        auto keys = inflight_messages_[r.first];
+  void on_new_response(rpc::RPCResponse response) {
```
dcrankshaw: nit: can you change the method name to `on_response_recv`?

src/libclipper/src/containers.cpp
```diff
@@ -29,7 +29,7 @@ void ModelContainer::send_prediction(PredictTask task) {
 ActiveContainers::ActiveContainers()
     : containers_(
           std::unordered_map<VersionedModelId,
-                             std::vector<std::shared_ptr<ModelContainer>>,
+                             std::map<int, std::shared_ptr<ModelContainer>>,
```
dcrankshaw, line 35: what's the key for the map?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -177,78 +185,64 @@ class TaskExecutor {
   /// receiving
   /// thread that will each continuously try to send/receive.
```
dcrankshaw: Delete this TODO

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -177,78 +185,64 @@ class TaskExecutor {
...
-            // inflight_messages_
-            // map between the time a message is sent and when it gets inserted
-            // into the map
-            std::unique_lock<std::mutex> l(inflight_messages_mutex_);
-            // std::vector<const std::vector<uint8_t>> serialized_inputs;
+  void on_new_container_added(VersionedModelId model_id, int replica_id) {
+    // TODO: Create a new model queue if this is the first connected container
+    // hosting the specified model (i.e. replica_id == 0)
+    UNUSED(model_id);
+    UNUSED(replica_id);
```
dcrankshaw: Instead of using the UNUSED macro, comment out the argument names. E.g.

```cpp
void on_new_container_added(VersionedModelId /*model_id*/, int /*replica_id*/) {
```

This is just a style thing, I think it's a little cleaner.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -177,78 +185,64 @@ class TaskExecutor {
...
-        for (auto c : containers) {
-          auto batch = c->dequeue_predictions(max_batch_size);
-          if (batch.size() > 0) {
-            // move the lock up here, so that nothing can pull from the
-            // inflight_messages_
-            // map between the time a message is sent and when it gets inserted
-            // into the map
-            std::unique_lock<std::mutex> l(inflight_messages_mutex_);
-            // std::vector<const std::vector<uint8_t>> serialized_inputs;
+  void on_new_container_added(VersionedModelId model_id, int replica_id) {
```
dcrankshaw: Since we're not passing this callback to the RPCService, we don't actually need this method. We already have a callback executed when a new container is added, the lambda that gets called when the Redis container table listener is fired.

Delete this method.

src/libclipper/src/rpc_service.cpp
```diff
@@ -223,6 +235,22 @@ void RPCService::receive_message(
...
+    std::pair<VersionedModelId, int> container_info = container_info_entry->second;
+
+    // TODO: These callbacks should be submitted to a threadpool for
+    // execution on separate threads
+    std::thread([&]() {
+      new_response_callback_(response);
+    }).detach();
+    std::thread([&]() {
+      container_ready_callback_(container_info.first, container_info.second);
+    }).detach();
```
dcrankshaw: Instead of spawning a separate thread to execute these, just execute them directly in the thread of execution. It will be slow, but we will be adding a threadpool this week in #80.

src/libclipper/src/rpc_service.cpp
```diff
@@ -98,6 +104,7 @@ void RPCService::manage_service(const string address) {
   // Note that zeromq socket id is a byte vector
   log_info_formatted(LOGGING_TAG_RPC, "RPC thread started at address: ", address);
   boost::bimap<int, vector<uint8_t>> connections;
+  std::unordered_map<std::vector<uint8_t>, std::pair<VersionedModelId, int>, VectorHasher<uint8_t>> containers;
```
dcrankshaw: Can you give `containers` a more descriptive variable name and add a comment about it's purpose? I got confused because I didn't look at the type signature closely enough and thought it was storing the actual `ModelContainer` objects. It's really a map from ZeroMQ connection ID to Clipper container ID (where a clipper container ID is the three part key `model_name, model_version, container_number`).

dcrankshaw (changes requested):
Mostly looks good. Some fairly minor comments.

src/libclipper/include/clipper/rpc_service.hpp
Corey-Zumar: Done

src/libclipper/include/clipper/containers.hpp
Corey-Zumar: The key is the replica id. I've added a comment to this effect.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Removed!

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Removed the method. Will do this in the future.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/src/containers.cpp
Corey-Zumar: The key is the replica id. I added a comment to the header in which the map is declared.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done. Renamed to "connections_containers_map", added comments regarding the data it stores.

src/libclipper/include/clipper/rpc_service.hpp
Corey-Zumar: I moved the method `<T> primitive_input_hash(std::vector<T)` in `datatypes.cpp` to a generic `hash_vector` method in `util.hpp`. The new method makes use of `boost::hash_range()` and is now used for hashing input vectors for serialization as well as for the rpc use case of hashing connection ids.

src/libclipper/src/rpc_service.cpp
Corey-Zumar: Done.

Corey-Zumar (commented):
Addressed your comments. If this merged prior to #80, we should make sure that we don't build and push a docker image with an extremely slow rpc layer.

dcrankshaw (approved):

dcrankshaw:
Jenkins test this please

dcrankshaw:
LGTM. Once jenkins finishes testing I'll merge.

dcrankshaw:
jenkins, test this please

dcrankshaw:
jenkins, test this please

PR #80, opened 2017-03-06
[CLIPPER-93] Adds threadpool

dcrankshaw:
Adds a threadpool implementation that arbitrary functions can be submitted to. This is intended for more cooperative use of physical resources between the different layers of the system.

### DIFF
--- a/src/frontends/src/query_frontend_main.cpp
+++ b/src/frontends/src/query_frontend_main.cpp
@@ -24,7 +24,7 @@ int main(int argc, char* argv[]) {
   clipper::Config& conf = clipper::get_config();
   conf.set_redis_address(options["redis_ip"].as<std::string>());
   conf.set_redis_port(options["redis_port"].as<int>());
-  conf.set_default_threadpool_size(options["threadpool_size"].as<int>());
+  conf.set_task_execution_threadpool_size(options["threadpool_size"].as<int>());
   conf.ready();
 
   query_frontend::RequestHandler<clipper::QueryProcessor> rh(
--- a/src/frontends/src/query_frontend_main.cpp
+++ b/src/frontends/src/query_frontend_main.cpp
@@ -12,14 +12,19 @@ int main(int argc, char* argv[]) {
                            "Clipper query processing frontend");
   // clang-format off
   options.add_options()
-    ("redis_ip", "Redis address", cxxopts::value<std::string>()->default_value("localhost"))
-    ("redis_port", "Redis port", cxxopts::value<int>()->default_value("6379"));
+    ("redis_ip", "Redis address",
+        cxxopts::value<std::string>()->default_value("localhost"))
+    ("redis_port", "Redis port",
+        cxxopts::value<int>()->default_value("6379"))
+    ("threadpool_size", "Number of threads for the task execution threadpool",
+        cxxopts::value<int>()->default_value("4"));
   // clang-format on
   options.parse(argc, argv);
 
   clipper::Config& conf = clipper::get_config();
   conf.set_redis_address(options["redis_ip"].as<std::string>());
   conf.set_redis_port(options["redis_port"].as<int>());
+  conf.set_default_threadpool_size(options["threadpool_size"].as<int>());
   conf.ready();
 
   query_frontend::RequestHandler<clipper::QueryProcessor> rh(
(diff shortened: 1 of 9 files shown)


src/frontends/src/query_frontend_main.cpp
```diff
@@ -12,14 +12,19 @@ int main(int argc, char* argv[]) {
...
   // clang-format off
   options.add_options()
-    ("redis_ip", "Redis address", cxxopts::value<std::string>()->default_value("localhost"))
-    ("redis_port", "Redis port", cxxopts::value<int>()->default_value("6379"));
+    ("redis_ip", "Redis address",
+        cxxopts::value<std::string>()->default_value("localhost"))
+    ("redis_port", "Redis port",
+        cxxopts::value<int>()->default_value("6379"))
+    ("threadpool_size", "Number of threads for the task execution threadpool",
+        cxxopts::value<int>()->default_value("4"));
```
Corey-Zumar, line 20: We should experiment with the default number of threads. I'm not sure that 4 is sufficient to handle high load.

src/frontends/src/rpc_service_bench.cpp
```diff
@@ -25,7 +26,7 @@ void await_responses(std::shared_ptr<clipper::rpc::RPCService> rpc_service,
       if (start_time != times_map->end()) {
         long elapsed = end - start_time->second;
         total_time_elapsed += elapsed;
-        printf("%d ms\n", (int)elapsed);
+        std::cout << (int)elapsed << " ms" << std::endl;
```
Corey-Zumar: Use a call to `log_info_formatted`

src/frontends/src/rpc_service_bench.cpp
```diff
@@ -172,9 +173,9 @@ void run_benchmarks() {
 
   shutdown = true;
   rpc_service->stop();
-  printf("%f ms\n",
-         ((double)total_time_elapsed) /
-             (num_iterations_per_benchmark * num_benchmarks));
+  double time_per_iter = ((double)total_time_elapsed) /
+                         (num_iterations_per_benchmark * num_benchmarks);
+  std::cout << time_per_iter << " ms" << std::endl;
```
Corey-Zumar: `log_info_formatted` instead of `cout`

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -0,0 +1,269 @@
+#ifndef CLIPPER_LIB_THREADPOOL_HPP
```
Corey-Zumar, line 1: Is there a source from which this threadpool was modified, or did you roll your own? If applicable, we should reference it here, and it would be helpful to see what was changed.

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+ private:
+  std::atomic_bool done_;
+  ThreadSafeQueue<std::unique_ptr<IThreadTask>> work_queue_;
+  std::vector<std::thread> threads_;
+};
+
+namespace DefaultThreadPool {
```
Corey-Zumar: Consider using a name here that's more specific to the task executor. `TaskExecutionThreadPool`, for example. The description for the configuration argument specifying the number of threads in this pool is specific to task execution: *"Number of threads for the task execution threadpool"*.

Corey-Zumar (approved):
After testing this with #79, this threadpool will work properly in conjunction with interactive rpc. All tests pass. 

Included a couple of comments regarding naming and documentation. Do we intend to have thread pools associated with different system components (TaskExecutor, etc) or a single thread pool shared between them?

src/frontends/src/query_frontend_main.cpp
dcrankshaw: Agreed. I filed [CLIPPER-101](https://clipper.atlassian.net/browse/CLIPPER-101) to explore this. I don't think we can make this decision until the system is completed though.

src/libclipper/include/clipper/threadpool.hpp
dcrankshaw: Added a reference link in the comments. Only things changed were some variable names and I added unit tests.

src/libclipper/include/clipper/threadpool.hpp
dcrankshaw: Sure.

Corey-Zumar:
Looks good! Will merge after status checks pass

PR #84, opened 2017-03-13
[CLIPPER-129] Update RPC benchmark to use event-based RPC service

dcrankshaw:
This also cleans up the benchmark to make it a little cleaner and easier to use.


src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+// taken from http://stackoverflow.com/a/12468109/814642
+std::string gen_random_string(size_t length) {
+  auto randchar = []() -> char {
+    const char charset[] =
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
+    const size_t max_index = (sizeof(charset) - 1);
+    return charset[rand() % max_index];
```
Corey-Zumar: We need to seed the generator before `rand()` is called. `srand(time(NULL));`, for example

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+    std::string str = gen_random_string(150);
+    std::shared_ptr<SerializableString> input =
+        std::make_shared<SerializableString>(str);
+    request.add_input(input);
+  }
+  return request;
+}
+
+rpc::PredictionRequest create_request(InputType input_type) {
+  int num_inputs = 500;
```
Corey-Zumar: We should make `num_inputs` a parameter.

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  clipper::Config &conf = clipper::get_config();
+  conf.set_redis_address(options["redis_ip"].as<std::string>());
+  conf.set_redis_port(options["redis_port"].as<int>());
+  conf.ready();
+  Benchmarker benchmarker(options["num_messages"].as<int>(),
+                          InputType::Doubles);
+  benchmarker.start();
+
+  // auto jh = std::thread([&benchmarker]() { benchmarker.start(); });
```
Corey-Zumar: Remove commented code

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+  std::unique_lock<std::mutex> l(benchmarker.bench_completed_cv_mutex);
+  benchmarker.bench_completed_cv_.wait(
+      l, [&benchmarker]() { return benchmarker.bench_completed_ == true; });
+  // jh.join();
```
Corey-Zumar: Remove commented code

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (messages_completed_ < num_messages_) {
+      send_message();
+    } else {
+      bench_completed_ = true;
+      bench_completed_cv_.notify_all();
+    }
+  }
+
+  std::condition_variable_any bench_completed_cv_;
+  std::mutex bench_completed_cv_mutex;
```
Corey-Zumar: This should have a trailing underscore: `bench_completed_cv_mutex_`

bench/README.md
```diff
@@ -37,24 +37,24 @@ If a configuration file is not specified, the benchmark will prompt you for the
 **These steps are given relative to the current directory.**
 
 1. Execute the following:
-  ```
-  $ ./setup_bench.sh <path_to_cifar_python_dataset>
+  ```sh
```
Corey-Zumar, line 40: That's really cool. Didn't know about `sh`

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+        cxxopts::value<int>()->default_value("100"));
+  // clang-format on
+  options.parse(argc, argv);
+
+  clipper::Config &conf = clipper::get_config();
+  conf.set_redis_address(options["redis_ip"].as<std::string>());
+  conf.set_redis_port(options["redis_port"].as<int>());
+  conf.ready();
+  Benchmarker benchmarker(options["num_messages"].as<int>(),
+                          InputType::Doubles);
```
Corey-Zumar: We should add an argument allowing us to specify the input type.

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+        recv_time_millis - cur_msg_start_time_millis_;
+    std::vector<float> deserialized_outputs =
+        deserialize_outputs(response.second);
+    float sum = 0.0;
+    for (float f : deserialized_outputs) {
+      sum += f;
+    }
+    log_info_formatted(LOGGING_TAG_RPC_BENCH, "output sum: {}", sum);
+    msg_latency_hist_->insert(message_duration_millis);
+    throughput_meter_->mark(1);
```
Corey-Zumar: Because we only send a new message when a response is received, it seems that our throughput will always be relatively low.

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+    cur_msg_start_time_millis_ =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now().time_since_epoch())
+            .count();
+    cur_message_id_ =
+        rpc_->send_message(request_.serialize(), benchmark_container_id_);
+  }
+
+  void on_response_recv(rpc::RPCResponse response) {
+    // process the response
```
Corey-Zumar: Remove comment

src/benchmarks/src/rpc_service_bench.cpp
```diff
@@ -0,0 +1,286 @@
...
+  std::vector<std::shared_ptr<DoubleVector>> input_vec;
+  std::vector<std::shared_ptr<Input>> inputs = get_primitive_inputs(
+      num_inputs, 784, InputType::Doubles, type_vec, input_vec);
+  double sum = 0.0;
+  for (auto i : inputs) {
+    for (double d : std::static_pointer_cast<DoubleVector>(i)->get_data()) {
+      sum += d;
+    }
+  }
+  log_info_formatted(LOGGING_TAG_RPC_BENCH, "Request inputs sum: {}", sum);
```
Corey-Zumar: Why are we computing and logging this sum for doubles?

Corey-Zumar (changes requested):
Looks good overall, mostly nits and doc requests.

bench/README.md
dcrankshaw: You can also use ` ```console ` or ` ```bash `

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Good catch

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: I was using it as a checksum to make sure that the Java container implementation was working correctly. I'll remove it.

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Done. Make it a command line parameter.

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Done.

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: So that's actually how the containers work. Each container can only handle one message at a time. So this should reflect the same throughput as the full system. Keep in mind that we're sending mini-batches though, so each message contains many requests.

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Good catch.

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Done

src/benchmarks/src/rpc_service_bench.cpp
dcrankshaw: Done

dcrankshaw (commented):
Addressed review comments

dcrankshaw:
@Corey-Zumar addressed your comments and tested it with the Python Noop container.

Sorry for polluting the PR a bit with whitespace changes, I guess I still haven't fully worked out the format checker in the CI script.

Corey-Zumar (approved):
LGTM

dcrankshaw:
@Corey-Zumar Fixed formatting so the checks all pass. Merge this when you get a chance.

PR #86, opened 2017-03-15
[CLIPPER-102] Integrate threadpool with TaskExecutor, RPC layer

Corey-Zumar:


src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -101,9 +102,10 @@ class TaskExecutor {
             active_containers_->add_container(
                 vm, std::stoi(container_info["zmq_connection_id"]),
                 replica_id, parse_input_type(container_info["input_type"]));
-            // TODO: This callback should be submitted to a thread pool
-            // for execution on a separate thread
-            on_container_ready(vm, replica_id);
+
+            TaskExecutionThreadPool::submit_job([=]() {
```
dcrankshaw: What is this closure capturing? Can we explicitly capture only the variables we need?

dcrankshaw: I think this can be: `TaskExecutionThreadPool::submit_job([this, vm, replica_id]() {`

dcrankshaw (changes requested):
Looks good. Just make the closure capture explicit.

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Agreed, it's better to explicitly capture only the required variables. Fixed!

dcrankshaw (approved):

dcrankshaw:
LGTM

PR #88, opened 2017-03-18
[CLIPPER-94][CLIPPER-133] Model-based queueing and starvation fix

Corey-Zumar:
These two issues are addressed together because applying our starvation fixed to the previous container-based queuing solution led to invalid memory accesses by deprecated prediction scheduling code in the task executor.


src/frontends/src/bench.cpp
```diff
@@ -70,7 +70,6 @@ void send_predictions(std::unordered_map<std::string, std::string> &config, Quer
     std::vector<boost::future<Response>> futures;
     std::vector<int> binary_labels;
     for (int i = 0; i < batch_size; i++) {
-      std::srand(time(NULL));
```
Corey-Zumar: Seeding at every iteration of the loop caused `rand()` to consistently produce the same value. We now seed once prior to the execution of `send_predictions()`

src/libclipper/include/clipper/containers.hpp
```diff
@@ -12,6 +12,8 @@
 
 namespace clipper {
 
+using Deadline = std::chrono::time_point<std::chrono::high_resolution_clock>;
```
dcrankshaw, line 15: Nice. We should just use a deadline throughout. Do you feel like changing the `Query` class to have a deadline instead of a latency? You can set it by looking at the latency SLO of the associated application in `query_frontend.hpp` and figure out what the deadline should be as soon as the query is received.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -58,13 +58,90 @@ class PredictionCache {
   std::shared_ptr<metrics::RatioCounter> hit_ratio_;
 };
 
-template <typename Scheduler>
+struct DeadlineCompare {
+  bool operator()(const std::pair<Deadline, PredictTask>& lhs,
+                  const std::pair<Deadline, PredictTask>& rhs) {
+    return lhs.first > rhs.first;
+  }
+};
```
dcrankshaw: Can you add a unit test for this? I tend to mess up my comparisons and I want to be safe here.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -58,13 +58,90 @@ class PredictionCache {
...
+
+  ModelQueue(ModelQueue&&) = default;
+  ModelQueue& operator=(ModelQueue&&) = default;
+
+  ~ModelQueue() = default;
+
+  void add_task(PredictTask task) {
+    std::unique_lock<std::mutex> l(queue_mutex_);
+    Deadline deadline = std::chrono::high_resolution_clock::now() +
+        std::chrono::microseconds(task.latency_slo_micros_);
```
dcrankshaw: Yeah as I mentioned above, let's set the deadline for the query once when it first arrives, then use it throughout execution.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -58,13 +58,90 @@ class PredictionCache {
...
+    queue_.emplace(deadline, std::move(task));
+    queue_not_empty_.notify_all();
```
dcrankshaw: What's this notification for?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -58,13 +58,90 @@ class PredictionCache {
...
+
+  // TODO: If we implement the scheduler so that it calls
+  // get_earliest_deadline(), then uses that to compute the batch size,
+  // then calls dequeue with that batch size, it's possible for a new
+  // task with an earlier deadline to be submitted. I'm not sure what the
+  // correct behavior here should be.
+  Deadline get_earliest_deadline() {
+    std::unique_lock<std::mutex> l(queue_mutex_);
+    if (queue_.size() == 0) {
+      queue_not_empty_.wait(l, [this] { return queue_.size() > 0; });
```
dcrankshaw: Based on our discussion about task starvation, this should be a non-blocking function right? Maybe it should return a `boost::option`?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -108,6 +185,11 @@ class TaskExecutor {
             });
             // TODO: Create a new model queue if this is the first connected container
```
dcrankshaw: Delete the TODO if this PR addresses it

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -162,29 +247,41 @@ class TaskExecutor {
...
   std::shared_ptr<metrics::Counter> predictions_counter;
   std::shared_ptr<metrics::Meter> throughput_meter;
   std::shared_ptr<metrics::Histogram> latency_hist;
+  std::unordered_map<const VersionedModelId, ModelQueue, decltype(&versioned_model_hash)> model_queues_;
+  static constexpr int INITIAL_MODEL_QUEUES_MAP_SIZE = 100;
+
+  bool create_model_queue_if_necessary(const VersionedModelId &model_id) {
+    auto queue_added = model_queues_.emplace(std::piecewise_construct,
+                                              std::forward_as_tuple(model_id),
+                                              std::forward_as_tuple());
```
dcrankshaw: What are the semantics of `emplace()` here? This won't create a new entry if the key already exists?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -162,29 +247,41 @@ class TaskExecutor {
...
       throw std::runtime_error("TaskExecutor failed to find previously registered active container!");
     }
+    auto model_queue_entry = model_queues_.find(container->model_);
+    if(model_queue_entry == model_queues_.end()) {
+      throw std::runtime_error("Failed to find model queue associated with a previously registered container!");
+    }
     
-    while(true) {
-      auto batch = container->dequeue_predictions(max_batch_size_);
+    for(int i = 0; i < 1000; i++) {
```
dcrankshaw: Honestly I don't think we even need to spin here. Delete the for-loop, just have an if-statement.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -162,29 +247,41 @@ class TaskExecutor {
...
+      Deadline earliest_deadline = model_queue_entry->second.get_earliest_deadline();
```
dcrankshaw: Note that in the current implementation, `get_earliest_deadline()` is blocking. I made a comment up by the function definition to make it non-blocking but repeating it here.

dcrankshaw (changes requested):
Looks good. Can we unit-test ModelQueue and any of these other pieces? It would be good to improve our test coverage before the release.

dcrankshaw:
jenkins test this please

dcrankshaw:
The reason the CI format check wasn't picking things up is because I had turned it off 😃 . I just turned it back on which is why the jenkins test failed.

Run `./bin/format_code.sh` to fix it.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: From [cppreference](http://en.cppreference.com/w/cpp/container/unordered_map/emplace): "(emplace) 		
Inserts a new element into the container constructed in-place with the given args if there is no element with the key in the container."

Added a comment to this effect.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Good idea. Changed so that it's non-blocking and returns `boost::optional<Deadline>`.

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Obsolete condition variable notifier for blocking calls to `get_earliest_deadline()`. Removed.

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Wrote a nonblocking version that returns a `boost::optional<Deadline>` instead

src/libclipper/include/clipper/task_executor.hpp
Corey-Zumar: Done!

src/libclipper/include/clipper/containers.hpp
Corey-Zumar: I think we should do this in a separate PR, as it involves changes in several pieces of infrastructure that aren't immediately related to this task.

Corey-Zumar (commented):
Addressed comments and added an additional test on `ModelQueue` for correct task ordering.

src/libclipper/include/clipper/containers.hpp
dcrankshaw: Good point

dcrankshaw (approved):
LGTM

PR #89, opened 2017-03-18
[CLIPPER-105] Refactor selection policy to support single-model applications with default output

dcrankshaw:
This is a big PR, but it's not quite as big as it looks. I refactored the JSON utils to split it between a .hpp and a .cpp file and I fixed some formatting with the code formatter. The big changes are in the query processor and the selection policies.

I refactored the selection policy API to be an abstract base class that implementations inherit from. This makes adding a new policy much simpler than the templates-based API we had previously and cleans up the QueryProcessor implementation substantially. I also ripped out all the existing policies and added the DefaultOutputSelectionPolicy which returns the model prediction if it arrives in time, and otherwise returns a static default output that is configured on a per-application basis.

Currently this PR does not update `clipper_manager.py` or the tutorial to the new selection policy implementation, which means they are broken.

Do you think we should update them here and fix this all in one PR so that we go from working codebase to working codebase? Or should we separate them into to PRs because they are somewhat distinct changes?

### DIFF
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -161,7 +161,7 @@ class RequestHandler {
     // selection policies have a default output?
     //
     // Initialize selection state for this application
-    if (policy == "DefaultOutputSelectionPolicy") {
+    if (policy == clipper::DefaultOutputSelectionPolicy::get_name()) {
       clipper::DefaultOutputSelectionPolicy p;
       clipper::Output parsed_default_output(std::stod(default_output), {});
       auto init_state = p.init_state(parsed_default_output);
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -121,7 +121,7 @@ class RequestHandler {
             std::string default_output = app_info["default_output"];
             int latency_slo_micros = std::stoi(app_info["latency_slo_micros"]);
             add_application(name, candidate_model_names, input_type, policy,
-                            default_output_, latency_slo_micros);
+                            default_output, latency_slo_micros);
           }
         });
 
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -118,9 +118,10 @@ class RequestHandler {
             InputType input_type =
                 clipper::parse_input_type(app_info["input_type"]);
             std::string policy = app_info["policy"];
+            std::string default_output = app_info["default_output"];
             int latency_slo_micros = std::stoi(app_info["latency_slo_micros"]);
             add_application(name, candidate_model_names, input_type, policy,
-                            latency_slo_micros);
+                            default_output_, latency_slo_micros);
           }
         });
 
@@ -154,7 +155,21 @@ class RequestHandler {
 
   void add_application(std::string name, std::vector<std::string> models,
                        InputType input_type, std::string policy,
-                       long latency_slo_micros) {
+                       std::string default_output, long latency_slo_micros) {
+    // TODO: QueryProcessor should handle this. We need to decide how the
+    // default output fits into the generic selection policy API. Do all
+    // selection policies have a default output?
+    //
+    // Initialize selection state for this application
+    if (policy == "DefaultOutputSelectionPolicy") {
+      clipper::DefaultOutputSelectionPolicy p;
+      clipper::Output parsed_default_output(std::stod(default_output), {});
+      auto init_state = p.init_state(parsed_default_output);
+      clipper::StateKey state_key{name, clipper::DEFAULT_USER_ID, 0};
+      query_processor_.get_state_table()->put(state_key,
+                                              p.serialize(init_state));
+    }
+
     auto predict_fn = [this, name, input_type, policy, latency_slo_micros,
                        models](std::shared_ptr<HttpServer::Response> response,
                                std::shared_ptr<HttpServer::Request> request) {
--- a/src/libclipper/include/clipper/persistent_state.hpp
+++ b/src/libclipper/include/clipper/persistent_state.hpp
@@ -25,7 +25,7 @@ size_t state_key_hash(const StateKey& key);
 // Threadsafe, non-copyable state storage
 class StateDB {
  public:
-  StateDB();
+  explicit StateDB();
   ~StateDB();
 
   // Disallow copies because of the mutex
--- a/src/libclipper/include/clipper/persistent_state.hpp
+++ b/src/libclipper/include/clipper/persistent_state.hpp
@@ -25,7 +25,7 @@ size_t state_key_hash(const StateKey& key);
 // Threadsafe, non-copyable state storage
 class StateDB {
  public:
-  explicit StateDB();
+  StateDB();
   ~StateDB();
 
   // Disallow copies because of the mutex
--- a/src/libclipper/include/clipper/selection_policies.hpp
+++ b/src/libclipper/include/clipper/selection_policies.hpp
@@ -12,6 +12,20 @@ namespace clipper {
 
 const std::string LOGGING_TAG_SELECTION_POLICY = "SELECTIONPOLICY";
 
+/**
+ * An instance of a SelectionPolicy class must be stateless and Clipper
+ * provides no guarantees about which instance of a SelectionPolicy
+ * will be used at any given time. The only reason SelectionPolicy
+ * objects are created at all is that in C++ using object hierarchies
+ * is the simplest way to achieve the type of polymorphic specialization
+ * we need for selection policies.
+ *
+ * Because these SelectionPolicy instances are stateless, all state
+ * needed for processing is encapsulated in a SelectionState object
+ * which is managed by Clipper and stored in a persistent
+ * database. The separate of policy and state allows Clipper to re-use
+ * the same SelectionPolicy instance with different SelectionStates.
+ */
 class SelectionState {
  public:
   SelectionState() = default;
@@ -102,6 +116,9 @@ class DefaultOutputSelectionPolicy : public SelectionPolicy {
   DefaultOutputSelectionPolicy& operator=(DefaultOutputSelectionPolicy&&) =
       default;
   ~DefaultOutputSelectionPolicy() = default;
+
+  static std::string get_name() const;
+
   std::shared_ptr<SelectionState> init_state(Output default_output) const;
 
   std::vector<PredictTask> select_predict_tasks(
--- a/src/libclipper/include/clipper/selection_policies.hpp
+++ b/src/libclipper/include/clipper/selection_policies.hpp
@@ -19,10 +19,6 @@ class SelectionState {
   SelectionState& operator=(const SelectionState&) = default;
   SelectionState(SelectionState&&) = default;
   SelectionState& operator=(SelectionState&&) = default;
-  /**
-   * Constructor to create a SelectionState object from serialized
-   * representation.
-   */
   virtual ~SelectionState();
   virtual std::string get_debug_string() const = 0;
 };
@@ -41,22 +37,6 @@ class SelectionPolicy {
   SelectionPolicy(SelectionPolicy&&) = default;
   SelectionPolicy& operator=(SelectionPolicy&&) = default;
   virtual ~SelectionPolicy();
-  // virtual SelectionState initialize(
-  //     const std::vector<VersionedModelId>& candidate_models) const = 0;
-  // virtual std::shared_ptr<SelectionState> update_candidate_models(
-  //     std::shared_ptr<SelectionState> state,
-  //     const std::vector<VersionedModelId>& candidate_models) const = 0;
-
-  // Used to identify a unique selection policy instance. For example,
-  // if using a bandit-algorithm that does not tolerate variable-armed
-  // bandits, one could hash the candidate models to identify
-  // which policy instance corresponds to this exact set of arms.
-  // Similarly, it provides flexibility in how to deal with different
-  // versions of the same arm (different versions of same model).
-  virtual long hash_models(
-      const std::vector<VersionedModelId>& /*models*/) const {
-    return 0;
-  }
 
   // Query Pre-processing: select models and generate tasks
   virtual std::vector<PredictTask> select_predict_tasks(
--- a/src/libclipper/include/clipper/selection_policies.hpp
+++ b/src/libclipper/include/clipper/selection_policies.hpp
@@ -28,6 +28,12 @@ class SelectionState {
 };
 
 class SelectionPolicy {
+  // Note that we need to use pointers to the SelectionState because
+  // C++ doesn't let you have an object with an abstract type on the stack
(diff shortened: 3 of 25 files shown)


Corey-Zumar:
Regarding required changes to `clipper_manager.py` and the tutorial, a separate PR seems to make the most sense. I think it's best to hold off on merging this PR until a PR has been submitted and reviewed that makes the necessary changes to the manager and tutorial components.

Reviewing this now...

src/libclipper/include/clipper/persistent_state.hpp
```diff
@@ -25,7 +25,7 @@ size_t state_key_hash(const StateKey& key);
 // Threadsafe, non-copyable state storage
 class StateDB {
  public:
-  StateDB();
+  explicit StateDB();
```
Corey-Zumar: A zero-argument constructor doesn't benefit from the `explicit` keyword

src/libclipper/include/clipper/selection_policies.hpp
```diff
@@ -8,324 +8,123 @@
...
-using ModelInfo = std::unordered_map<std::string, double>;
-
-/** Model Map
- * A map of the models to their corresponding model information
- */
-using Map = std::unordered_map<VersionedModelId, ModelInfo,
-                               std::function<size_t(const VersionedModelId&)>>;
-
-class BanditPolicyState {
+class SelectionState {
```
Corey-Zumar, line 29: We should have some documentation here explaining the role of selection state in enacting selection policies.

src/libclipper/include/clipper/selection_policies.hpp
```diff
@@ -8,324 +8,123 @@
...
-  std::string debug_string() const;
-  Map model_map_;
-  double weight_sum_ = 0.0;  // Only for Exp3, Exp4
+  SelectionState() = default;
+  SelectionState(const SelectionState&) = default;
+  SelectionState& operator=(const SelectionState&) = default;
+  SelectionState(SelectionState&&) = default;
+  SelectionState& operator=(SelectionState&&) = default;
+  virtual ~SelectionState() = default;
+  virtual std::string get_debug_string() const = 0;
```
Corey-Zumar, line 37: When would this be called? Is there an intended format for the debug string?

src/libclipper/include/clipper/selection_policies.hpp
```diff
@@ -8,324 +8,123 @@
...
-  static VersionedModelId select(BanditPolicyState& state);
+  DefaultOutputSelectionPolicy() = default;
+  DefaultOutputSelectionPolicy(const DefaultOutputSelectionPolicy&) = default;
+  DefaultOutputSelectionPolicy& operator=(const DefaultOutputSelectionPolicy&) =
+      default;
+  DefaultOutputSelectionPolicy(DefaultOutputSelectionPolicy&&) = default;
+  DefaultOutputSelectionPolicy& operator=(DefaultOutputSelectionPolicy&&) =
+      default;
+  ~DefaultOutputSelectionPolicy() = default;
+  std::shared_ptr<SelectionState> init_state(Output default_output) const;
```
Corey-Zumar: What does this method do? It seems like state initialization will be necessary for many selection policies; we should consider defining a virtual `init_state` method in `SelectionPolicy` and overriding it here.

src/libclipper/src/selection_policies.cpp
```diff
@@ -1,569 +1,117 @@
...
-    score_sum += state.model_map_[model_id]["weight"] * p.y_hat_;
-    weight_sum += state.model_map_[model_id]["weight"];
-    models.push_back(model_id);
-  }
-
-  double y_hat = score_sum / weight_sum;
-  // Turn y_hat into either 0 or 1
-  if (y_hat < 0.5) {
-    y_hat = 0.0;
+  int num_candidate_models = query.candidate_models_.size();
```
Corey-Zumar: nit: this should be a `size_t`

src/libclipper/src/selection_policies.cpp
```diff
@@ -1,569 +1,117 @@
...
-}
-
-std::string Exp3Policy::serialize_state(BanditPolicyState state) {
-  return state.serialize();
+std::string DefaultOutputSelectionState::get_debug_string() const {
+  rapidjson::Document d;
+  d.SetObject();
+  json::add_double(d, "y_hat", default_output_.y_hat_);
+  std::vector<std::string> empty_vec;
+  json::add_string_array(d, "models_used", empty_vec);
```
Corey-Zumar, line 35: The fact that we're adding an empty "models_used" field to the debug string for an output state that will never depend on any models indicates some intended consistency/format for these debug strings. We should define it.

src/libclipper/src/selection_policies.cpp
```diff
@@ -1,569 +1,117 @@
...
-  double y_hat = score_sum / weight_sum;
-  // Turn y_hat into either 0 or 1
-  if (y_hat < 0.5) {
-    y_hat = 0.0;
+  int num_candidate_models = query.candidate_models_.size();
+  if (num_candidate_models == 0) {
+    log_error_formatted(LOGGING_TAG_SELECTION_POLICY,
+                        "No candidate models for query with label {}",
+                        query.label_);
+  } else if (num_candidate_models == 1) {
```
Corey-Zumar: Avoid duplicate code:

```
std::vector<PredictTask> DefaultOutputSelectionPolicy::select_predict_tasks(
    std::shared_ptr<SelectionState> /*state*/, Query query,
    long query_id) const {
  std::vector<PredictTask> tasks;
  size_t num_candidate_models = query.candidate_models_.size();
  if (num_candidate_models == 0) {
    log_error_formatted(LOGGING_TAG_SELECTION_POLICY,
                        "No candidate models for query with label {}",
                        query.label_);
  } else {
    if(num_candidate_models > 1) {
      log_error_formatted(LOGGING_TAG_SELECTION_POLICY,
                          "{} candidate models provided for query with label "
                              "{}. Picking the first one.",
                          num_candidate_models, query.label_);
    }
    tasks.emplace_back(query.input_, query.candidate_models_.front(), 1.0,
                       query_id, query.latency_budget_micros_);
  }
  return tasks;
}
```


```

File: src/libclipper/src/selection_policies.cpp [hunk 8]
```diff
@@ -1,569 +1,117 @@
...
-  VersionedModelId selected_model;
-  if (state.model_map_.empty()) {  // Edge case
-    std::cout << "No models to select from." << std::endl;
+Output DefaultOutputSelectionPolicy::combine_predictions(
+    const std::shared_ptr<SelectionState>& state, Query /*query*/,
+    std::vector<Output> predictions) const {
+  if (predictions.size() == 1) {
+    return predictions.front();
+  } else if (predictions.empty()) {
+    return std::static_pointer_cast<DefaultOutputSelectionState>(state)
```
Corey-Zumar: This could cause undefined behavior if `state` isn't an instance of `std::shared_ptr<DefaultOutputSelectionState>`. As far as I understand it, using a dynamic cast will perform a runtime check and throw a consistent error in the case where the types are incompatible.

src/libclipper/test/redis_test.cpp
```diff
@@ -147,17 +147,19 @@ TEST_F(RedisTest, AddApplication) {
       std::make_pair("simple_svm", 2), std::make_pair("music_cnn", 4)};
   InputType input_type = InputType::Doubles;
   std::string policy = "exp3_policy";
```
Corey-Zumar: This is no longer an accurate name.

src/libclipper/test/selection_policies_test.cpp
```diff
@@ -9,298 +9,84 @@
...
-  }
-  // Test if times selected model_0 > model_1 > model_2
-  ASSERT_GT(select_0, select_1);
-  ASSERT_GT(select_1, select_2);
-
-  /* Serialization Test */
-  auto bytes = Exp3Policy::serialize_state(state);
-  auto new_state = Exp3Policy::deserialize_state(bytes);
-  ASSERT_EQ(state.weight_sum_, new_state.weight_sum_);
+TEST_F(DefaultOutputSelectionPolicyTest, InitState) {}
```
Corey-Zumar: This is an empty test

src/libclipper/test/selection_policies_test.cpp
```diff
@@ -9,298 +9,84 @@
...
+
+TEST_F(DefaultOutputSelectionPolicyTest, TestSelectPredictTasks) {
```
Corey-Zumar: We should add some documentation or use a more descriptive test name.

src/frontends/src/query_frontend.hpp
```diff
@@ -131,7 +132,21 @@ class RequestHandler {
...
   void add_application(std::string name, std::vector<VersionedModelId> models,
                        InputType input_type, std::string policy,
-                       long latency_slo_micros) {
+                       std::string default_output, long latency_slo_micros) {
+    // TODO: QueryProcessor should handle this. We need to decide how the
+    // default output fits into the generic selection policy API. Do all
+    // selection policies have a default output?
+    //
+    // Initialize selection state for this application
+    if (policy == "DefaultOutputSelectionPolicy") {
```
Corey-Zumar: We should define a static `get_name()` function within each selection policy. 

`policy == DefaultOutputSelectionPolicy::get_name()` is cleaner and less error prone.

src/libclipper/src/query_processor.cpp
```diff
@@ -101,48 +39,40 @@ std::shared_ptr<StateDB> QueryProcessor::get_state_table() const {
...
     // TODO better error handling
-    return boost::make_ready_future(
-        Response{query, query_id, 20000, Output{1.0, {std::make_pair("m1", 1)}},
-                 std::vector<VersionedModelId>()});
+    return error_response;
   }
+  std::shared_ptr<SelectionPolicy> current_policy = current_policy_iter->second;
+
+  auto state_opt = state_db_->get(StateKey{query.label_, query.user_id_, 0});
+  if (!state_opt) {
```
Corey-Zumar, line 56: After changing the specified selection policy in the end-to-end benchmark from "EXP3" to "DefaultOutputSelectionPolicy", the benchmark is still logging an error here:
```
[04:22:23.560][error] [QUERYPR...] No selection state found for query with label: test
```

Corey-Zumar (changes requested):
Having trouble running the end-to-end benchmark with these policy changes. Can you take a look at this? 

In general, the changes make sense - comments are mostly nits and documentation.

src/libclipper/include/clipper/persistent_state.hpp
dcrankshaw: Good catch. I added a single-argument constructor then deleted it and forgot to remove explicit.

src/libclipper/include/clipper/selection_policies.hpp
dcrankshaw: Done

src/libclipper/include/clipper/selection_policies.hpp
dcrankshaw: It is called from the admin frontend. There's no particular format any policy needs to conform to. Whatever the implementer thinks would be useful information to expose for debugging/runtime inspection.

src/libclipper/include/clipper/selection_policies.hpp
dcrankshaw: The problem with defining a virtual init_state method is that it's likely the arguments to it will vary based on the selection policy. So I'm not sure what the function signature would be. For the 0.1 release, we only have one selection policy (the DefaultOutputSelectionPolicy). init_state is called each time a new application is created, which is the correct behavior for _this_ selection policy. When we expand our functionality in 0.2 to support more sophisticated types of selection policies, we will need to revisit this. But I already knew we would need to refine our selection policy API anyway. [CLIPPER-99](https://clipper.atlassian.net/browse/CLIPPER-99) will track this effort.

src/libclipper/src/selection_policies.cpp
dcrankshaw: Not really. The debug string here is what the default output would be. An output is the pair <y_hat, models_used> and if the default output is being returned, no models were used. Debug string format is purposely left up to the user.

Also, next up on the TODO list is [CLIPPER-107](https://clipper.atlassian.net/browse/CLIPPER-107), which will change Output to be a json string. So this output will be refactored anyway.

src/libclipper/src/selection_policies.cpp
dcrankshaw: Done

src/libclipper/src/selection_policies.cpp
dcrankshaw: Done

src/libclipper/test/redis_test.cpp
dcrankshaw: Fixed

src/libclipper/src/selection_policies.cpp
dcrankshaw: Changed `static_pointer_cast`s to `dynamic_pointer_cast`.

src/libclipper/test/selection_policies_test.cpp
dcrankshaw: Oops. Removed it. There's not much to test there.

src/libclipper/test/selection_policies_test.cpp
dcrankshaw: Split the SelectPredictTask and CombinePredictions tests into individual test cases with more descriptive test names.

src/frontends/src/query_frontend.hpp
dcrankshaw: Good idea.

dcrankshaw:
I fixed the `clipper_manager` library, but this PR still breaks the tutorial. Unfortunately, we won't really be able to fix it until [CLIPPER-111](https://clipper.atlassian.net/browse/CLIPPER-111) is implemented so that we can update the version of a model associated with an application.

Corey-Zumar (approved):
Looks good to me. We shouldn't merge this until [CLIPPER-111](https://clipper.atlassian.net/browse/CLIPPER-111) is addressed and the tutorial is fixed.

dcrankshaw:
Once #115 is merged, this can be rebased on develop and will be ready for merge.

Corey-Zumar:
Note: Tutorial calls to `deploy_model()` need to be refactored to specify a default output instead of a selection policy.

dcrankshaw:
@Corey-Zumar I rebased on develop to include the model versioning PR #115 and fixed the tutorial to specify a default output instead of a selection policy. I believe this is ready to go. Definitely do a last pass over the code, and run the tutorial to make sure it works (I ran it as well without any problems). When you run the tutorial, note that you'll need to build the Clipper docker containers locally to test.

Corey-Zumar:
Confirmed that the `end_to_end_benchmark` is still functional, the tutorial is working properly, and all tests pass. LGTM. Merging...

PR #95, opened 2017-03-23
[CLIPPER-135] Add flag for optional sudo execution in clipper manager

### LINKED ISSUE #94: clipper asks for SSH password
When i connect to clipper, even though i am providing the SSH key, clipper still asks for the password. This only happens once. 

````
cm.Clipper(host, user, key)
````

Maybe something to do with Fabric api? I am not logging in as root, but the user has sudo permission

Corey-Zumar:
Fixes #94


dcrankshaw (approved):
LGTM

PR #101, opened 2017-04-02
Capture output in clipper_manager local commands

dcrankshaw:
Fabric's `local` command works a bit differently than `run` and `sudo` and by default doesn't capture output. This breaks some of the local commands which rely on parsing that output. The place I found it crop up was when trying to deploy a model container where the Docker image was built locally but didn't exist on DockerHub.

This should fix the bug.


Corey-Zumar (approved):
LGTM

PR #103, opened 2017-04-03
Make vagrant provision run without root permission

lucasmoura:
While developing with Vagrant, I just realized that the provision step is being executed with root permissions. 

https://www.vagrantup.com/docs/provisioning/shell.html#privileged

Therefore, all files created during that step would belong to the root user. With that configuration, if an user using vagrant wants to run the tests, it will have to run as the root user, which is not necessary.

vagrant_bug

The above Figure shows the problem. This MR fixes that issue, by not allowing the provision step to be run with root permission.


dcrankshaw:
jenkins test this please

dcrankshaw (approved):
Nice catch. LGTM.

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

Issue #77, opened 2017-02-22
Possible `labels_to_str` undefined behavior.


mwhittaker:
The `labels_to_str` function [dereferences `labels.end() - 1`][1] where `labels` is a vector. If `labels` is an empty vector than `*(label.end() - 1)` is undefined behavior (I think). 

[1]: https://github.com/ucbrise/clipper/blob/develop/src/libclipper/src/redis.cpp#L62

mwhittaker:
[Here][1] too.

[1]: https://github.com/ucbrise/clipper/blob/910d1a31c911018203e4d1926d68659684ad9d6b/src/libclipper/src/redis.cpp#L88-L89

dcrankshaw:
Fixed by #104

PR #104, opened 2017-04-03
Fix empty vector problem in reddis.cpp

lucasmoura:
This MR address the following issue #77, which if an empty vector is received, it will look at an invalid position of this vector.

To deal with that, this MR checks if the received vector is empty, and if it is, it returns an empty string. Also, tests were created for both of the functions were this behavior occur.


dcrankshaw:
jenkins test this please

lucasmoura:
@dcrankshaw I believe the code is in the right format now

dcrankshaw:
jenkins test this please

dcrankshaw (approved):
LGTM

PR #106, opened 2017-04-03
Makes run_unittests.sh actually exit when tests fail

dcrankshaw:


dcrankshaw:
This is going to break the build (which is actually already broken) so I'm force merging it

PR #109, opened 2017-04-04
[CLIPPER-126] Remove application command

nishadsingh1:
https://clipper.atlassian.net/projects/CLIPPER/issues/CLIPPER-126?filter=allopenissues

Endpoint, handler, tests, and management library function added for deleting apps.

We could use a 204 http response instead of a 200 when we successfully delete, but then we wouldn't be able to send a "Success!" message in the response.

### DIFF
--- a/management/clipper_manager.py
+++ b/management/clipper_manager.py
@@ -369,6 +369,20 @@ class Clipper:
         r = requests.post(url, headers=headers, data=req_json)
         print(r.text)
 
+    def delete_application(self, name):
+        """Deletes an existing Clipper application.
+
+        Parameters
+        ----------
+        name : str
+            The name of the application.
+        """
+        url = "http://%s:1338/admin/delete_app" % self.host
+        req_json = json.dumps({"name": name})
+        headers = {'Content-type': 'application/json'}
+        r = requests.delete(url, headers=headers, data=req_json)
+        print(r.text)
+
     def get_all_apps(self, verbose=False):
         """Gets information about all applications registered with Clipper.
 
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -253,10 +253,10 @@ class RequestHandler {
   }
 
   void delete_application(std::string name) {
-    std::string predict_endpoint = _get_endpoint_address(name, "predict");
-    std::string update_endpoint = _get_endpoint_address(name, "update");
-    server_.delete_endpoint(predict_endpoint, "POST");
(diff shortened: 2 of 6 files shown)


dcrankshaw:
jenkins test this please

dcrankshaw (changes requested):
This will delete an application from Redis, but it won't actually delete it from the Clipper query processor.

Add an else if statement to the query frontend [here](https://github.com/ucbrise/clipper/blob/bac5ec8b8242fc622b69abe6d9c4581cfdc1cbe8/src/frontends/src/query_frontend.hpp#L109-L123):

```
else if (event_type == "del") {
  std::string name = key;
  delete_application(name);
}
```
Where `delete_application()` deletes the 2 endpoints added [here](https://github.com/ucbrise/clipper/blob/bac5ec8b8242fc622b69abe6d9c4581cfdc1cbe8/src/frontends/src/query_frontend.hpp#L162) and [here](https://github.com/ucbrise/clipper/blob/bac5ec8b8242fc622b69abe6d9c4581cfdc1cbe8/src/frontends/src/query_frontend.hpp#L188).

You'll need to modify `server_http.hpp` as well to add a `delete_endpoint()` method. Look at the [`add_endpoint`](https://github.com/ucbrise/clipper/blob/develop/src/libs/httpserver/server_http.hpp#L271-L293) code to see how endpoints are added. Basically do the same thing but in reverse.

nishadsingh1:
This isn't working yet -- `delete_endpoint` in server_http isn't actually decreasing the number of endpoints. Just pushed to the branch as a progress update

dcrankshaw:
jenkins ok to test

nishadsingh1:
If we run `QueryFrontendTest.TestDeleteApplication` (without the asserts in `server_http.hpp`'s `delete_endpoint`), `EXPECT_EQ(no_apps, rh_.num_applications())` fails because `rh_.num_applications()` is 1 (and `no_apps` is 0). `r_.num_applications()` relies on `server_http.num_endpoints()`. That `rh_.num_applications()` is still 1 after a call to `delete_endpoint` suggests that no endpoint was deleted.

In the latest commit I made asserts in `server_http.hpp`'s `delete_endpoint` that isolate the scope of that problem to just the code in `delete_endpoint`. Do you mind taking a look at them?

rmdort:
Is this ready for merge? @nishadsingh1

nishadsingh1:
This gets more complicated with #226 and isn't working in it's current state. Closing

Issue #73, opened 2017-02-22
Duplicate `models_used_` field.


mwhittaker:
Inside of `datatypes.hpp`, the `Output` class has a [`std::vector<VersionedModelId> models_used_` field](https://github.com/ucbrise/clipper/blob/910d1a31c911018203e4d1926d68659684ad9d6b/src/libclipper/include/clipper/datatypes.hpp#L44). The `Response` class includes an `Output output_` field, but it also includes a [`std::vector<VersionedModelId> models_used_` field](https://github.com/ucbrise/clipper/blob/910d1a31c911018203e4d1926d68659684ad9d6b/src/libclipper/include/clipper/datatypes.hpp#L239-L240)! I'm guessing the repeated `models_used_` was an accident?

dcrankshaw:
Fixed by the change to JSON outputs in #134

PR #110, opened 2017-04-05
Remove models_used_ variable from Response

lucasmoura:
This MR addresses #73, which states that the variable **models_used_** is being used on both **Output** and **Response**.

I have checked for the use of this variable and it seems it is only being used for the **Output** object. Therefore, I have removed the **models_used_** variable from the **Response** class.

### DIFF
--- a/src/frontends/src/query_frontend_tests.cpp
+++ b/src/frontends/src/query_frontend_tests.cpp
@@ -15,7 +15,7 @@ class MockQueryProcessor {
  public:
   MockQueryProcessor() = default;
   boost::future<Response> predict(Query query) {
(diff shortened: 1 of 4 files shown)


dcrankshaw:
@Corey-Zumar is working on rewriting the `Response` type to be more flexible (it will just be JSON) which will take care of this issue as well as part of [CLIPPER-107](https://clipper.atlassian.net/browse/CLIPPER-107), so I'm going to hold off on merging this.

@lucasmoura If you'd like to get more involved in the project send me an email (crankshaw@cs.berkeley.edu) and I can point you to some more issues to work on.

PR #111, opened 2017-04-06
[CLIPPER-139] Simplified python predict function deployment

nishadsingh1:
This commit allows for python developers in an anaconda environment to deploy prediction functions to Clipper without manually creating a container.

The management library has a new function, `deploy_predict_function`, that serializes the developer's anaconda environment and provided prediction function and sends them off to be deployed to a container. This container has an image with some packages pre-installed (see preinstalled_packages.txt). It will create a new environment, load the provided serialized anaconda environment, and run an instance of PredictContainer within that environment. PredictContainer loads in the serialized prediction function and will use it to respond to queries.

In order for this commit to be functional, we need an image of the container to be pushed to clipper/predict_func_container:latest. It would be great if @dcrankshaw could build the container (corresponding dockerfile is named PredictDockerfile) and do that push. Right now, nishadsingh/predict_func_container:latest has this image, so feel free to substitute that for the "clipper/...." address in the clipper management library for testing.

PS – I'm almost certain the naming chosen for containers/etc is really poor...I'm just feeling really unimaginative. I'm definitely open to changing them.


containers/python/python_container.py
```diff
@@ -0,0 +1,96 @@
...
+        predict_path = "{dir}/{predict_fname}".format(
+            dir=path, predict_fname=predict_fname)
+
+        self.predict_func = load_predict_func(predict_path)
+
+        print("Loaded prediction function")
+
+    def predict_ints(self, inputs):
+        if self.input_type != rpc.INPUT_TYPE_INTS:
+            pass
```
dcrankshaw: you should probably print an error in the case of a type mismatch instead of silently passing

containers/python/python_container.py
```diff
@@ -0,0 +1,96 @@
...
+    if "CLIPPER_IP" in os.environ:
+        ip = os.environ["CLIPPER_IP"]
+    else:
+        print("Connecting to Clipper on localhost")
+
+    port = 7000
+    if "CLIPPER_PORT" in os.environ:
+        port = int(os.environ["CLIPPER_PORT"])
+    else:
+        print("Connecting to Clipper with default port: 7000")
```
dcrankshaw: Use string formatting and print the `port` variable instead of hardcoding 7000 into the string

containers/python/pywrencloudpickle.py
```diff
@@ -0,0 +1,878 @@
+"""
```
dcrankshaw, line 1: Did you modify this file at all, or is it straight from PyWren?

dcrankshaw (changes requested):
This looks great. I made a few changes and cleaned some stuff up as I was testing. It ended up just being easier to push to your branch instead of adding a bunch of comments.

Can you add an example of how to use `deploy_predict_function` to the method's documentation? For example, you can add a pre-processing step that centers each input around the mean before making a prediction. Something like the following (double check that this code actually works though):

```py
def center(xs):
    means = np.mean(xs, axis=1)
    return xs - means

centered_xs = center(xs)
model = LogisticRegression()
model.fit(centered_xs, ys)

def centered_predict(inputs):
    centered_inputs = center(inputs)
    return model.predict(centered_inputs)

clipper.deploy_predict_function("example_model", 1, centered_predict, ["example"], "doubles", num_containers=1)
```

dcrankshaw:
jenkins test this please

dcrankshaw:
@nishadsingh1 Make sure you pull my changes before addressing the rest of the review comments

dcrankshaw:
jenkins test this please

dcrankshaw:
When adding the example to the python documentation, use the [Google Pydoc style guide](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

containers/python/pywrencloudpickle.py
nishadsingh1: Pulled straight from Pywren, with the exception of some formatting changes from calling the formatting script. Is there an appropriate way to attribute it / deal with files taken from other repos like this one?

dcrankshaw: Nope. Just need to include the license header which you did.

containers/python/python_container_entry.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+
+conda env create -f=/model/environment.yml -n container-env
+source activate container-env
+python -u /container/python_container.py
```
dcrankshaw: So on further discussion, we don't want to run Python in unbuffered mode. Remove the `"-u"`. Instead, let's flush the buffered output at the end of each iteration of message processing.

Add the following 2 lines on [line 175 of`rpc.py`](https://github.com/ucbrise/clipper/blob/develop/containers/python/rpc.py#L175)

```py
sys.stdout.flush()
sys.stderr.flush()
```

Make sure the flushes are outside the `else` statement like this:

```py
            else:
                received_msg = Message(msg_id_bytes, [])
                response = self.handle_feedback_request(received_msg)
                response.send(self.socket)
                print("recv: %f us" % ((t2 - t1).microseconds))
            sys.stdout.flush()
            sys.stderr.flush()

```

dcrankshaw (changes requested):
Just one comment about the buffered output. Other than that, LGTM.

dcrankshaw (approved):

dcrankshaw:
jenkins test this please

Issue #112, opened 2017-04-06
safe handling of duplicate register_application requests


sueann:
Q: what happens if you try to register the same application twice (same name)?

A: The update will overwrite the old application with the new one. I think the place this will cause problems is you will have multiple query handlers registered to the same REST endpoint and I'm not sure which one will get called. We should probably prevent this from happening (which is a 2 line code change).

dcrankshaw:
Fixed by #115. Clipper now throws an error if you try to register an application with the same name twice.

PR #115, opened 2017-04-07
[CLIPPER-113] Implement model active version tracking and updates

dcrankshaw:
This PR associated a `current_model_version` with each model and makes applications specify candidate model names but not versions. Instead, the current version of a model is deployed. Model versions are automatically updated when a newer model version is deployed. I also implemented a `set_model_version` REST call in the management frontend that allows users to manually set the current model version to any version that has previously been deployed. This allows for model rollback and roll-forward.

I removed the feedback section of the selection policy and the tutorial wording still needs a little bit of work.


dcrankshaw:
Note that in order to test this with the revised tutorial, you will need to build the `query_frontend` and `management_frontend` containers locally. The ones deployed to DockerHub are from the develop branch.

dcrankshaw:
@nishadsingh1 Can you make a version of the [deploy sklearn model image](https://github.com/ucbrise/clipper/blob/develop/examples/tutorial/img/deploy_sklearn_model.png) from the tutorial for the TensorFlow model? In other words, delete the sklearn box from [this image](https://github.com/ucbrise/clipper/blob/develop/examples/tutorial/img/deploy_tf_model.png). With this new model versioning PR, the TF model _replaces_ the sklearn model and I would like to illustrate that.

Also, what did you use to make those images? Can you add whatever powerpoint/google slides file you used to the Clipper Google Drive?

dcrankshaw:
jenkins test this please

src/frontends/src/query_frontend.hpp
```diff
@@ -112,33 +112,66 @@ class RequestHandler {
...
     auto predict_fn = [this, name, input_type, policy, latency_slo_micros,
                        models](std::shared_ptr<HttpServer::Response> response,
                                std::shared_ptr<HttpServer::Request> request) {
       try {
-        auto prediction =
-            decode_and_handle_predict(request->content.string(), name, models,
-                                      policy, latency_slo_micros, input_type);
+        std::vector<VersionedModelId> versioned_models;
+        {
+          std::unique_lock<std::mutex> l(current_model_versions_mutex_);
```
Corey-Zumar, line 164: Do we need to hold this lock for the entire query processing and response creation procedure?

src/frontends/src/query_frontend.hpp
```diff
@@ -165,8 +198,19 @@ class RequestHandler {
         std::shared_ptr<HttpServer::Response> response,
         std::shared_ptr<HttpServer::Request> request) {
       try {
-        auto update = decode_and_handle_update(request->content.string(), name,
-                                               models, policy, input_type);
+        std::vector<VersionedModelId> versioned_models;
+        {
+          std::unique_lock<std::mutex> l(current_model_versions_mutex_);
```
Corey-Zumar, line 203: As above, do we need to hold this lock for the entire update procedure or just until the `versioned_models` map is populated? If the answer is the latter, releasing the lock earlier may avoid resource contention between the `update` and `predict` functions.

src/libclipper/include/clipper/redis.hpp
```diff
@@ -277,6 +298,20 @@ void subscribe_to_application_changes(
     redox::Subscriber& subscriber,
     std::function<void(const std::string&, const std::string&)> callback);
 
+/**
+* Subscribes to changes in model versions.
+*
+* The
```
Corey-Zumar: Nit: check documentation formatting

src/libclipper/src/metrics.cpp
```diff
@@ -1,3 +1,4 @@
+#include <algorithm>
```
Corey-Zumar, line 1: Is this include necessary? (Maybe gcc is throwing an error where clang isn't?, etc)

src/libclipper/src/redis.cpp
```diff
@@ -124,6 +141,34 @@ std::vector<VersionedModelId> str_to_models(const std::string& model_str) {
...
+}
+
+int get_current_model_version(redox::Redox& redis,
+                              const std::string& model_name) {
+  if (send_cmd_no_reply<string>(
+          redis, {"SELECT", std::to_string(REDIS_METADATA_DB_NUM)})) {
+    std::string key = gen_model_current_version_key(model_name);
+    auto result = send_cmd_with_reply<string>(redis, {"GET", key});
+    if (result) {
+      return std::stoi(*result);
```
Corey-Zumar: Should we verify that `result` is a nonnegative integer?

src/libclipper/src/redis.cpp
```diff
@@ -313,41 +366,48 @@ std::unordered_map<std::string, std::string> get_application_by_key(
...
-  subscribe_to_keyspace_changes(REDIS_MODEL_DB_NUM, subscriber,
+  subscribe_to_keyspace_changes(REDIS_MODEL_DB_NUM, "", subscriber,
                                 std::move(callback));
 }
 
 void subscribe_to_container_changes(
     Subscriber& subscriber,
     std::function<void(const std::string&, const std::string&)> callback) {
-  subscribe_to_keyspace_changes(REDIS_CONTAINER_DB_NUM, subscriber,
+  subscribe_to_keyspace_changes(REDIS_CONTAINER_DB_NUM, "", subscriber,
```
Corey-Zumar: Nit: For clarity, defining `const std::string NO_PREFIX = "";` and using it in place of `""` is clearer.

management/clipper_manager.py
```diff
@@ -555,6 +557,37 @@ def inspect_instance(self):
             s = r.text
         return s
 
+    def set_model_version(model_name, model_version, num_containers=0):
```
Corey-Zumar: We should make this the only point through which model rollback / rollforward can be achieved. Currently, we can also call `deploy_model()` while specifying a previously-added version of a model. However, `deploy_model()` has no checks in place to ensure that the model being added is actually equivalent to the old model corresponding to the given version. Hypothetically, version x of a model can therefore change throughout time, and this seems to defeat the purpose of versioning. 

We can fix this by adding checks to `deploy_model()` that fail if a model is deployed with a previously-specified version.

src/management/src/management_frontend_tests.cpp
```diff
@@ -130,4 +134,46 @@ TEST_F(ManagementFrontendTest, TestAddModelMalformedJson) {
...
+  std::string model_path2 = "/tmp/models/m/2";
+  ASSERT_TRUE(add_model(*redis_, model2, InputType::Ints, labels,
+                        container_name, model_path2));
+  VersionedModelId model4 = std::make_pair("m", 4);
+  std::string model_path4 = "/tmp/models/m/4";
+  ASSERT_TRUE(add_model(*redis_, model4, InputType::Ints, labels,
+                        container_name, model_path4));
+
+  ASSERT_FALSE(rh_.set_model_version("m", 11));
+  ASSERT_EQ(get_current_model_version(*redis_, "m"), -1);
```
Corey-Zumar: Wouldn't we expect `get_current_model_version()` to return `4`, corresponding to the last successful call to `add_model()` with `model4`? Is attempting to set an invalid model version supposed to change the model version to `-1`? If so, we should consistently maintain the previous version in the face of an update failure, if possible.

management/clipper_manager.py
Corey-Zumar: Missing `self` in parameter list

Corey-Zumar (changes requested):
Tests pass and functionality seems to be mostly complete. Comments address a couple of concerns involving behavior with `deploy_model()` in conjunction with `set_model_version`. Additionally:

* Line 491 of clipper_manager.py should refer to `Clipper.deploy_model()` instead of `Clipper.add_model()`.

* What’s the intent of the “Warning” image in the tutorial_part_2 notebook? Seems unclear to me.

src/frontends/src/query_frontend.hpp
dcrankshaw: Nope. The lock is in a nested block so it will go out of scope and be released on line 171 before the prediction is made.

src/frontends/src/query_frontend.hpp
dcrankshaw: The lock goes out of scope on line 210.

src/libclipper/include/clipper/redis.hpp
dcrankshaw: fixed

src/libclipper/src/metrics.cpp
dcrankshaw: `std::sort` which we use in this file is defined in `<algorithm>`. I prefer to explicitly include the header dependencies for a file as much as possible and I noticed this one was missing.

src/libclipper/src/redis.cpp
dcrankshaw: Yeah good idea.

src/libclipper/src/redis.cpp
dcrankshaw: Hmm I actually disagree here. Defining NO_PREFIX as a constant seems like treating this as a special case. I'll define a local variable `std::string prefix = ""` in each method instead.

management/clipper_manager.py
dcrankshaw: Yeah this is a really good point.

dcrankshaw: Fixed

src/management/src/management_frontend_tests.cpp
dcrankshaw: I was adding the model directly to Redis so `set_model_version` wasn't being called. I fixed this by calling the managament frontend's `RequestHandler::add_model` with JSON and it now should reflect the expected behavior.

management/clipper_manager.py
dcrankshaw: I added checks to the management frontend's `add_application` and `add_model` methods to reject duplicate adds.

dcrankshaw:
@Corey-Zumar I addressed your comments. I also added a check to prevent duplicate adding of applications or model versions. This fixes #112.

Corey-Zumar (approved):
Awesome, LGTM. Will merge after jenkins is able to build successfully.

dcrankshaw:
@Corey-Zumar BTW the warning in the tutorial is because @nishadsingh1 said he accidentally ran cleanup too early the first time he ran the tutorial. The idea was to have a visual separation from the tutorial code and the cleanup. It's been there for awhile, I just moved the location around slightly.

dcrankshaw:
@Corey-Zumar Hold off on merging this until the tutorial is updated. Should be sometime today.

dcrankshaw:
@Corey-Zumar Tutorial is updated. This PR is good to go once Jenkins passes it.

PR #116, opened 2017-04-09
[CLIPPER-107][CLIPPER-109] Return JSON-formatted responses to prediction queries

Corey-Zumar:
Modifies query frontend and dependent infrastructure to return JSON-formatted responses to prediction queries.

From [CLIPPER-107](https://clipper.atlassian.net/projects/CLIPPER/issues/CLIPPER-107) and the functional spec:

> The format of the response body is a JSON object with the following fields:
> 
> * Output (JSON key name “output”): The prediction JSON object.
> * Default used (JSON key name “default”): A boolean set to true if the model did not return in time and the default output was used. False otherwise.
>
> If there is a non-recoverable error during query execution, an error JSON object will be returned instead and the HTTP response code will be set to the most appropriate error code (a 400 or 500 error code). The format of the error is a JSON object with the following fields:
> * Error (JSON key name “error”): The name of the error.
> * Cause (JSON key name “cause”): A string describing the source of the error.

Note: This PR is currently based on the branch corresponding to #89. As a result, many changes from #89 are redundantly included here.


src/frontends/src/query_frontend.hpp
```diff
@@ -34,6 +34,16 @@ namespace query_frontend {
 const std::string LOGGING_TAG_QUERY_FRONTEND = "QUERYFRONTEND";
 const std::string GET_METRICS = "^/metrics$";
 
+const char* PREDICTION_RESPONSE_KEY_QUERY_ID = "query_id";
+const char* PREDICTION_RESPONSE_KEY_OUTPUT = "output";
+const char* PREDICTION_RESPONSE_KEY_USED_DEFAULT = "default";
+const char* PREDICTION_ERROR_RESPONSE_KEY_ERROR = "error";
+const char* PREDICTION_ERROR_RESPONSE_KEY_CAUSE = "cause";
```
dcrankshaw, line 42: Why are these `char*` instead of `std::string`? (I'm guessing there's a good reason)

src/libclipper/include/clipper/query_processor.hpp
```diff
@@ -6,6 +6,7 @@
 #include <tuple>
 #include <utility>
```
dcrankshaw: Add `#include <stdexcept>` (it's better to explicitly include all the headers needed)

src/libclipper/include/clipper/query_processor.hpp
```diff
@@ -55,6 +56,16 @@ class QueryProcessor {
...
+class PredictError : public std::runtime_error {
+ public:
+  PredictError(const long query_id, const std::string msg);
+  const char* what() const noexcept;
+
+ private:
+  const long query_id_;
+  const std::string msg_;
+};
+
```
dcrankshaw: Let's define this in a separate `exceptions.hpp` file.

src/libclipper/src/query_processor.cpp
```diff
@@ -26,6 +26,17 @@ using std::tuple;
...
+    : std::runtime_error(msg), query_id_(query_id), msg_(msg) {}
+
+const char* PredictError::what() const noexcept {
+  std::stringstream ss;
+  ss << "Failed to render a prediction for query with id " << query_id_
+     << std::endl;
+  ss << "Explanation: " << msg_ << std::endl;
+  return ss.str().data();
+}
+
```
dcrankshaw: Move to `exceptions.cpp`

src/management/src/management_frontend.hpp
```diff
@@ -52,7 +52,7 @@ const std::string APPLICATION_JSON_SCHEMA = R"(
      {"model_name" := string, "model_version" := int}
    ],
    "input_type" := "integers" | "bytes" | "floats" | "doubles" | "strings",
-   "selection_policy" := string,
+   "default_output" := number,
```
dcrankshaw: s/number/float

dcrankshaw (changes requested):
This looks great. We should wait for #89 to be merged first.

Does this require any changes to `clipper_manager.py` or the tutorial (beyond those needed for #89)?

src/frontends/src/query_frontend.hpp
Corey-Zumar: Our utility methods for adding key-value pairs to JSON objects expect `char*` keys. There is no implicit cast from `std::string` to `char*`, so I've defined the constants this way. Alternatively, I could define these with the `std::string` type and call `.data()` to get char* pointers, but this seems less readable.

src/management/src/management_frontend.hpp
Corey-Zumar: Done

Corey-Zumar (commented):
@dcrankshaw Addressed your comments - moved `PredictError` to separate `exceptions.hpp` and `exceptions.cpp` files.

src/libclipper/include/clipper/query_processor.hpp
Corey-Zumar: Done

src/libclipper/include/clipper/query_processor.hpp
Corey-Zumar: Included `stdexept` in `exceptions.hpp`

src/libclipper/src/query_processor.cpp
Corey-Zumar: Done

dcrankshaw:
LGTM. We need to wait on #115 and then #89 before merging this though

dcrankshaw (approved):

src/libclipper/src/query_processor.cpp
```diff
@@ -180,36 +110,22 @@ boost::future<Response> QueryProcessor::predict(Query query) {
...
-
-    } else if (query.selection_policy_ == "UCB") {
-      final_output =
-          combine_predictions<UCBPolicy>(query, outputs, serialized_state);
-    } else {
-      UNREACHABLE();
-    }
+    std::pair<Output, bool> final_output =
+        current_policy->combine_predictions(selection_state, query, outputs);
     std::chrono::time_point<std::chrono::high_resolution_clock> end =
```
dcrankshaw: @Corey-Zumar Can you add a ratio counter here that tracks the fraction of predictions that use the default output?

Corey-Zumar:
@dcrankshaw Rebased on `develop`. Please take another look over the code and confirm that the tutorial runs properly before merging - it's working well for me.

src/libclipper/src/query_processor.cpp
Corey-Zumar: Done

dcrankshaw:
LGTM. Confirming that the tutorial works for me as well.

PR #119, opened 2017-04-11
[CLIPPER-117] List apps and get app info requests served by mgmtfrontend API

nishadsingh1:
The management frontend API now has 2 new GET endpoints:
1. /get_application (returns a JSON object with full info about the requested app)
2. /get_applications (returns a list of names of all apps or, if provided `verbose`==True, a list of full JSON objects of all apps)

The returned JSON objects are in the same format that is requested by /add_application. The intent is that there will now be a uniform externally-facing app info format that is distinct from our redis storage format. See `management_frontend_tests.cpp` for examples.

Python management library now uses these endpoints. Their return types have changed slightly; they now return application objects in the same format requested by /add_application.

The task corresponding to exposing application utilities through the API: https://clipper.atlassian.net/browse/CLIPPER-117

If this approach seems reasonable, we can replicate it for models and containers.


management/clipper_manager.py
```diff
@@ -280,26 +280,26 @@ def register_application(self,
...
-                capture=True)
+        list
+            Returns a list of information about all apps registered to Clipper.
 
-            if len(result.stdout) > 0:
-                return result.stdout
-            else:
-                print("Clipper has no applications registered")
-                return ""
+            If `verbose` == False, the list contains the apps' names.
```
Corey-Zumar: Let's list `verbose` under a separate `Parameters` documentation header instead of `Returns`. See `register_applications()` for reference.

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -65,6 +74,20 @@ rapidjson::Value& check_kv_type_and_return(rapidjson::Value& d,
   return val;
 }
 
+/* Check for matching types else throw exception */
+rapidjson::Value& check_kv_type_is_bool_and_return(rapidjson::Value& d,
```
Corey-Zumar: The reason for defining this method for the `bool` case is that `check_kv_type_and_return` doesn't support checking multiple types, yet we are required to match against `rapidjson::kFalseType` and `rapidjson::kTrueType` for booleans. We could consider modifying `check_kv_type_and_return` to take a vector of types, but  this solution seems fine - just throwing the idea out there.

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -214,6 +246,12 @@ rapidjson::Value& get_object(rapidjson::Value& d, const char* key_name) {
   return object;
 }
 
+rapidjson::Value& get_array(rapidjson::Value& d, const char* key_name) {
```
Corey-Zumar: This is currently unused. Is there a purpose for including it in addition to our type-specific `get_array` methods (`get_int_array`, etc)?

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -335,6 +373,82 @@ void add_object(rapidjson::Document& d, const char* key_name,
...
+  size_t num_elements = string_vec.size();
+  for (size_t i = 0; i < num_elements; i++) {
+    rapidjson::Value v(rapidjson::StringRef(string_vec.at(i).c_str(),
+                                            string_vec.at(i).length()));
+    d.PushBack(v, d.GetAllocator());
+  }
+}
+
+/* Sets `d` to an array containing info from `candidate_models_redis_format`*/
+void set_candidate_models_doc(std::string& candidate_models_redis_format,
```
Corey-Zumar: We should test this method because it depends on a specific redis format that may change in the future. Consider creating a mock list of candidate models adhering to the current format as in `JsonUtilTests#TestParseCandidateModels` and invoking set_candidate_models_doc on this mock.

src/libclipper/include/clipper/json_util.hpp
```diff
@@ -335,6 +373,82 @@ void add_object(rapidjson::Document& d, const char* key_name,
...
+    rapidjson::Document candidate_model_doc(&d.GetAllocator());
+    candidate_model_doc.SetObject();
+    clipper::json::add_string(candidate_model_doc, "model_name", model_name);
+    clipper::json::add_int(candidate_model_doc, "model_version", model_version);
+    d.PushBack(candidate_model_doc, d.GetAllocator());
+  }
+}
+
+/* Sets `d` to an object containing reformatted info from `app_info`*/
+void set_app_info_doc(std::unordered_map<std::string, std::string>& app_info,
```
Corey-Zumar: Write a unit test for this method as well

src/libclipper/src/redis.cpp
```diff
@@ -34,6 +34,17 @@ std::unordered_map<string, string> parse_redis_map(
   return parsed_map;
 }
 
+std::vector<string> parse_redis_vector(
+        const std::vector<string>& redis_data) {
+  return redis_data;
```
Corey-Zumar: Unintentional return statement? Please document this method once it's fixed or remove it if it's unnecessary.

src/management/src/management_frontend.hpp
```diff
@@ -44,6 +51,7 @@ const std::string ADD_MODEL = ADMIN_PATH + "/add_model$";
 const std::string GET_METRICS = ADMIN_PATH + "/metrics$";
 const std::string GET_SELECTION_STATE = ADMIN_PATH + "/get_state$";
 const std::string GET_APPLICATIONS = ADMIN_PATH + "/get_applications$";
```
Corey-Zumar: For clarity, I think we should rename `get_applications` to `get_all_applications`. This creates a more intuitive distinction between the endpoint that retrieves all applications and the endpoint that retrieves a specific application based on user-provided parameters.

dcrankshaw:
jenkins test this please

src/management/src/management_frontend.hpp
```diff
@@ -259,6 +317,67 @@ class RequestHandler {
...
   /**
+   * Creates an endpoint that listens for requests to retrieve info about
+   * registered Clipper applications.
+   *
+   * JSON format:
+   * {
+   *  "verbose" := bool
+   * }
+   *
+   * \return Returns a JSON string that encodes a list with info about
```
Corey-Zumar, line 383: As in `clipper_manager.py`, move `verbose` under a separate `params` documentation section

src/management/src/management_frontend_tests.cpp
```diff
@@ -89,6 +89,270 @@ TEST_F(ManagementFrontendTest, TestAddApplicationMalformedJson) {
...
+  response_doc.SetObject();
+  parse_json(json_response, response_doc);
+
+  rapidjson::Document request_doc;
+  request_doc.SetObject();
+  parse_json(add_app_json, request_doc);
+
+  // The JSON provided for adding the app contains the same name-value
+  // attribute pairs that should be returned by `get_application`.
+  ASSERT_EQ(request_doc, response_doc);
```
Corey-Zumar: This is fine for now. I think it's possible that in the future we will specify JSON flags for an `add_application` call that should not be returned to the user by `get_application`.

management/clipper_manager.py
```diff
@@ -312,25 +312,19 @@ def get_app_info(self, name):
         Returns
         -------
         dict or None
-            Returns a dict with the application info if found. If the application
-            is not registered, None is returned.
+            Returns a dictionary with the specified application's info. This
+            will contain the attribute name-value pairs that were provided to
+            `register_application`. If no application with name `name` is
+            registered with Clipper, None is returned.
```
Corey-Zumar: This method does not return `None` if no application with name `name` is registered. Rather, invoking `get_app_info("nonexistent_name")` returns the following JSON dictionary:

```
{
  "name": "nonexistent_name"
}
```

Corey-Zumar (changes requested):
Looks good in general - noted a few minor errors and commented with requests for documentation & formatting nits. 

When you're done making changes, run `/bin/format_code.sh` in order to pass code format checks necessary for merge.

src/management/src/management_frontend.hpp
nishadsingh1: Are you sure about this? The other methods in management_frontend.hpp have their desired json input schema shown in the same way that it is here.

src/management/src/management_frontend.hpp
nishadsingh1: I like that :)

src/libclipper/src/redis.cpp
nishadsingh1: Thanks for the catch; the whole method is unnecessary.

src/libclipper/include/clipper/json_util.hpp
nishadsingh1: Yep, it's unnecessary – forgot to clean it up.

src/libclipper/include/clipper/json_util.hpp
nishadsingh1: Yep, that could also definitely work. Both of these solutions seem pretty hacky; I'll look into a cleaner fix.

management/clipper_manager.py
nishadsingh1: Great, thank you for the catch; I was manually adding the name back in within the endpoint handling function without checking if redis actually returned a valid app object.

management/clipper_manager.py
nishadsingh1: Cool, makes sense.

nishadsingh1:
Made the recommended changes and:
- Moved function implementations out of `json_util.hpp` to `json_util.cpp`
- Renamed relevant function, most notably set_app_info_doc —> set_json_doc_from_redis_app_metadata. 
- set_json_doc_from_redis_app_metadata now makes use of several helper functions dedicated to transforming specific fields from their redis format to their publicly facing format
- Updated this transformation for the changed candidate model storage
- moved `get_all_applications` logic out of `redis.cpp`. Before, I had a function in `redis.cpp` iterating and making requests to redis. This seems like the sort of thing that shouldn’t exist in an adapter.

Added/updated the following tests:

**Management front-end tests**
TestGetApplicationCorrect
TestGetNonexistantApplicationCorrect
TestGetApplicationMalformedJson

TestGetAllApplicationsVerboseCorrect
TestGetAllApplicationsVerboseNoneRegisteredCorrect

TestGetAllApplicationsNotVerboseCorrect
TestGetAllApplicationsNotVerboseNoneRegisteredCorrect

TestGetAllApplicationsMalformedJson

**Redis tests**
ListApplicationNames
ListApplicationNamesNoneRegistered

**JSON util tests**
TestSetJsonDocFromRedisAppMetadata

Running the code formatting script made some changes in the files I touched. I included those changes here but I'm happy to remove them if the previous formatting was intentional.

nishadsingh1:
I think that something about the way I rebased on develop and merged made this PR include a lot of changes that have already been made. I'm going to scrap it and make a new one – apologies for the confusion.

PR #120, opened 2017-04-13
Moved dockerfiles into same directory as build context

dcrankshaw:
I broke the Docker Hub builds. The dockerfiles need to be in the same directory as the build context for automated builds to work.

This should fix the problem.


Corey-Zumar (approved):
LGTM

PR #122, opened 2017-04-17
Fix example client to use new JSON format

dcrankshaw:


examples/basic_query/example_client.py
```diff
@@ -34,37 +18,45 @@ def predict(host, uid, x):
     print("'%s', %f ms" % (r.text, latency))
 
 
-def add_mnist_app(host):
+def add_example_app(host):
```
Corey-Zumar: To prevent future inconsistencies, it seems like a good idea to import `clipper_manager` and invoke `register_application`. We should do the same with `add_example_model` by calling `clipper_manager.register_external_model`.

Corey-Zumar (changes requested):
Let's use clipper manager in the example client so that we don't have to modify both the example client and clipper manager every time the management API changes

examples/basic_query/example_client.py
dcrankshaw: Yeah that's a good idea. I was thinking about trying to avoid a dependency on the clipper_manager to keep things simple for the example, but the example is illustrating how to query Clipper not how to initialize a model or application.

dcrankshaw:
Addressed your comments and added a note to `examples/basic_query/README` on how to run the example

Corey-Zumar (approved):
Nice, LGTM

PR #123, opened 2017-04-17
[CLIPPER-57] Read existing configuration from Redis on startup

giulio-zhou:
* Create two functions get_all_model_names() and
  get_all_application_names() which call the KEYS command with a
  wildcard (*) argument on the model and application table
* Update query frontend to (1) add existing applications and
  (2) fill internal state with current model name and versions
  on startup

### DIFF
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -149,16 +149,14 @@ class RequestHandler {
     // Read from Redis configuration tables and update models/applications
     // (1) Iterate through applications and set up predict/update endpoints.
     for (std::string app_name :
-            clipper::redis::get_all_application_names(redis_connection_)) {
+         clipper::redis::get_all_application_names(redis_connection_)) {
       // Is there a way/need to not repeat these function calls? (from above)
       auto app_info =
           clipper::redis::get_application_by_key(redis_connection_, app_name);
 
       std::vector<std::string> candidate_model_names =
-          clipper::redis::str_to_model_names(
-              app_info["candidate_model_names"]);
-      InputType input_type =
-          clipper::parse_input_type(app_info["input_type"]);
+          clipper::redis::str_to_model_names(app_info["candidate_model_names"]);
+      InputType input_type = clipper::parse_input_type(app_info["input_type"]);
       std::string policy = app_info["policy"];
       int latency_slo_micros = std::stoi(app_info["latency_slo_micros"]);
 
@@ -167,10 +165,9 @@ class RequestHandler {
     }
     // (2) Update current_model_versions_ with (model, version) pairs.
     for (std::string model_name :
-            clipper::redis::get_all_model_names(redis_connection_)) {
-      auto model_version =
-          clipper::redis::get_current_model_version(redis_connection_,
-                                                    model_name);
+         clipper::redis::get_all_model_names(redis_connection_)) {
+      auto model_version = clipper::redis::get_current_model_version(
+          redis_connection_, model_name);
       // Error handle model_version being < 0?
       std::unique_lock<std::mutex> l(current_model_versions_mutex_);
       current_model_versions_[model_name] = model_version;
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -28,6 +28,7 @@ using clipper::Feedback;
 using clipper::FeedbackQuery;
 using clipper::json::json_parse_error;
 using clipper::json::json_semantic_error;
+using clipper::redis::labels_to_str;
 using HttpServer = SimpleWeb::Server<SimpleWeb::HTTP>;
 
 namespace query_frontend {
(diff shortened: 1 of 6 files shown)


src/frontends/src/query_frontend.hpp
```diff
@@ -145,6 +145,33 @@ class RequestHandler {
             }
           }
         });
+
+    // Read from Redis configuration tables and update models/applications
+    // (1) Iterate through applications and set up predict/update endpoints.
+    for (std::string app_name :
+         clipper::redis::get_all_application_names(redis_connection_)) {
+      // Is there a way/need to not repeat these function calls? (from above)
```
dcrankshaw: Repeating them is fine. There's not much duplicate code.

dcrankshaw:
jenkins test this please

src/frontends/src/query_frontend.hpp
```diff
@@ -145,6 +145,33 @@ class RequestHandler {
...
+      auto app_info =
+          clipper::redis::get_application_by_key(redis_connection_, app_name);
+
+      std::vector<std::string> candidate_model_names =
+          clipper::redis::str_to_model_names(app_info["candidate_model_names"]);
+      InputType input_type = clipper::parse_input_type(app_info["input_type"]);
+      std::string policy = app_info["policy"];
```
dcrankshaw, line 173: This PR is based on an old commit. Update to the latest version. You'll notice that applications no longer have a policy field and now have a `default_output` field that needs to be set.

src/frontends/src/query_frontend.hpp
```diff
@@ -145,6 +145,33 @@ class RequestHandler {
...
+
+      add_application(app_name, candidate_model_names, input_type, policy,
+                      latency_slo_micros);
+    }
+    // (2) Update current_model_versions_ with (model, version) pairs.
+    for (std::string model_name :
+         clipper::redis::get_all_model_names(redis_connection_)) {
+      auto model_version = clipper::redis::get_current_model_version(
+          redis_connection_, model_name);
+      // Error handle model_version being < 0?
```
dcrankshaw: Yeah you should check for it. If it is < 0, log an error then throw a runtime exception.

src/frontends/src/query_frontend.hpp
```diff
@@ -145,6 +145,33 @@ class RequestHandler {
             }
           }
         });
+
+    // Read from Redis configuration tables and update models/applications
+    // (1) Iterate through applications and set up predict/update endpoints.
+    for (std::string app_name :
+         clipper::redis::get_all_application_names(redis_connection_)) {
```
dcrankshaw: Can you save the vector of existing app names and log them in a useful message. Something like "Found 3 existing applications registered in Clipper: app_name_1, app_name_2, app_name_3."

src/frontends/src/query_frontend.hpp
```diff
@@ -145,6 +145,33 @@ class RequestHandler {
...
+    }
+    // (2) Update current_model_versions_ with (model, version) pairs.
+    for (std::string model_name :
+         clipper::redis::get_all_model_names(redis_connection_)) {
+      auto model_version = clipper::redis::get_current_model_version(
+          redis_connection_, model_name);
+      // Error handle model_version being < 0?
+      std::unique_lock<std::mutex> l(current_model_versions_mutex_);
+      current_model_versions_[model_name] = model_version;
+    }
```
dcrankshaw: After you've initialized the `current_model_versions_` map in this for loop, can you log the results similar to the existing applications log message: "Found 3 models deployed to Clipper: model_a@v4, model_b@v7, model_c@v1"

dcrankshaw (changes requested):
This looks good. I tested it locally and it seems to work. We should add unit tests for this though.

Add some tests to the query_frontend that first populate the application/model tables using Redis then create a new query frontend `RequestHandler` object and make sure it reads Redis correctly and adds the right models/versions and applications.

Note that when adding models you'll need explicitly set the current model version as well because the management frontend does that (https://github.com/ucbrise/clipper/blob/develop/src/management/src/management_frontend.hpp#L282-L314).

Make sure to have separate tests for models and applications.

src/frontends/src/query_frontend.hpp
giulio-zhou: Yeah, will do.

src/frontends/src/query_frontend.hpp
giulio-zhou: Updated to include `default_output`.

src/frontends/src/query_frontend.hpp
giulio-zhou: Checked in new implementation with corresponding test.

src/frontends/src/query_frontend.hpp
giulio-zhou: Done.

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw (approved):

PR #126, opened 2017-04-21
[CLIPPER-156] Fix segfault in on_container_ready

dcrankshaw:
I believe this fixes the problem. After the fix I can't reproduce the segfault and logging indicates that this is catching a problem.

The sys path line in `bench_init.py` is hacky, but it's no more hacky than our existing Python path management.


Corey-Zumar (approved):
Fixes the seg fault and some python import errors, lgtm!

PR #127, opened 2017-04-22
[CLIPPER-96] Container batch size estimation

Corey-Zumar:

PR #128, opened 2017-04-22
[CLIPPER-118] Create per-model and per-application metrics

dcrankshaw:


src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -277,24 +324,38 @@ class TaskExecutor {
...
   std::mutex inflight_messages_mutex_;
+  // inflight_messages_ value is vector of <message_id, model, input>
   std::unordered_map<int, std::vector<std::tuple<const long, VersionedModelId,
                                                  std::shared_ptr<Input>>>>
       inflight_messages_;
   std::shared_ptr<metrics::Counter> predictions_counter;
-  std::shared_ptr<metrics::Meter> throughput_meter;
-  std::shared_ptr<metrics::Histogram> latency_hist;
+  std::shared_ptr<metrics::Meter> throughput_meter_;
+  boost::shared_mutex model_queues_mutex_;
```
Corey-Zumar: Good catch. The `model_queues_` map definitely isn't threadsafe.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -353,23 +415,38 @@ class TaskExecutor {
...
 
     inflight_messages_.erase(response.first);
     std::vector<float> deserialized_outputs =
         deserialize_outputs(response.second);
     assert(deserialized_outputs.size() == keys.size());
     int batch_size = keys.size();
     predictions_counter->increment(batch_size);
-    throughput_meter->mark(batch_size);
+    throughput_meter_->mark(batch_size);
     long current_time = std::chrono::duration_cast<std::chrono::milliseconds>(
```
Corey-Zumar: This is a duration in `milliseconds`, but the model prediction latency meter's unit is `microseconds`. This mistake is also present in line 280 of `task_executor_hpp` (milliseconds should be changed to microseconds in the duration cast - I think this is my fault, sorry). In #127, I've modified the `inflight message` entry to include the raw chrono datetime object, rather than a parsed long value - hopefully this helps reduce these issues in the future.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -277,24 +324,38 @@ class TaskExecutor {
...
 
   bool create_model_queue_if_necessary(const VersionedModelId &model_id) {
     // Adds a new <model_id, task_queue> entry to the queues map, if one
     // does not already exist
+    boost::unique_lock<boost::shared_mutex> l(model_queues_mutex_);
     auto queue_added = model_queues_.emplace(std::piecewise_construct,
                                              std::forward_as_tuple(model_id),
                                              std::forward_as_tuple());
-    return queue_added.second;
+    bool queue_created = queue_added.second;
```
Corey-Zumar, line 388: It's worth noting that calls to `register_external_model` in `clipper_manager` won't result in the addition of new model metrics until a container associated with the given model actually connects to Clipper. Do you think this will cause user confusion ("Why aren't there any metrics for my model?")? 

Imo this behavior is fine.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -353,23 +415,38 @@ class TaskExecutor {
...
     long current_time = std::chrono::duration_cast<std::chrono::milliseconds>(
                             std::chrono::system_clock::now().time_since_epoch())
                             .count();
-    for (int batch_num = 0; batch_num < batch_size; ++batch_num) {
-      long send_time = std::get<0>(keys[batch_num]);
-      latency_hist->insert(static_cast<int64_t>(current_time - send_time));
-      cache_.put(std::get<1>(keys[batch_num]), std::get<2>(keys[batch_num]),
-                 Output{deserialized_outputs[batch_num],
-                        {std::get<1>(keys[batch_num])}});
+    if (keys.size() > 0) {
```
Corey-Zumar: `if (batch_size > 0) {`

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -353,23 +415,38 @@ class TaskExecutor {
...
+      const VersionedModelId &cur_model = std::get<1>(keys[0]);
+      auto cur_model_metric_entry = model_metrics_.find(cur_model);
+      if (cur_model_metric_entry != model_metrics_.end()) {
```
Corey-Zumar: This seems like a good place for an `optional`. Instead of checking the predicate `cur_model_metric_entry != model_metrics_.end()` and accessing `model_metrics_entry->second` in two places, we could populate an optional that contains the model metric if it's available and check for the presence of the model metric within the optional in the inner loop. This would be more readable.

src/libclipper/include/clipper/datatypes.hpp
```diff
@@ -27,6 +27,7 @@ enum class RequestType {
 };
 
 size_t versioned_model_hash(const VersionedModelId &key);
+std::string versioned_model_to_str(const VersionedModelId &model);
```
Corey-Zumar, line 30: Nit: return `const std::string` here and for `get_readable_input_type`

Corey-Zumar (changes requested):
Looks good overall. The only substantial problem I caught was the resolution of the timestamps used to calculate per-model latency. This was a preexisting issue that has become more blatant as a result of the metrics updates.

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: I think it's fine for now

src/libclipper/include/clipper/datatypes.hpp
dcrankshaw: Returning const values is a bad idea (http://stackoverflow.com/a/8716406/814642). It prevents rvalue optimization which avoids a copy. But also from an API perspective helper functions should impose minimal restrictions on the calling functions. Taking const parameters when possible is good, because it allows the function to be called with either const or non-const arguments. Conversely, returning non-const values is good, because it allows the calling function to mutate the result if needed.

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: Good idea

dcrankshaw:
@Corey-Zumar Addressed review comments

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -423,7 +499,10 @@ class TaskExecutor {
         std::chrono::duration_cast<std::chrono::milliseconds>(task_latency)
             .count();
     processing_container->update_throughput(1, task_latency_micros);
-    latency_hist->insert(static_cast<int64_t>(task_latency_millis));
+    if (cur_model_metric) {
+      (*cur_model_metric)
+          .latency_->insert(static_cast<int64_t>(task_latency_millis));
```
Corey-Zumar: This should be micros: `._latency_->insert(static_cast<int64_t>(task_latency_micros))`. We can also remove `task_latency_millis` (unused)

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -261,15 +297,11 @@ class TaskExecutor {
           }
 
         });
-    throughput_meter = metrics::MetricsRegistry::get_metrics().create_meter(
-        "model_throughput");
+    throughput_meter_ = metrics::MetricsRegistry::get_metrics().create_meter(
+        "internal:aggregate_model_throughput");
     predictions_counter =
```
Corey-Zumar: `predictions_counter_`

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -391,27 +452,42 @@ class TaskExecutor {
...
     std::unique_lock<std::mutex> l(inflight_messages_mutex_);
     auto keys = inflight_messages_[response.first];
+    boost::shared_lock<boost::shared_mutex> metrics_lock(model_metrics_mutex_);
 
     inflight_messages_.erase(response.first);
     std::vector<float> deserialized_outputs =
         deserialize_outputs(response.second);
     assert(deserialized_outputs.size() == keys.size());
     int batch_size = keys.size();
     predictions_counter->increment(batch_size);
```
Corey-Zumar: This counter is also incremented when prediction tasks are scheduled via `schedule_predictions`. As a result, in the case where a single application is registered, the aggregate predictions count incorrectly exceeds the application prediction count.

Corey-Zumar (changes requested):
Almost ready for merge - found a couple more issues

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: Done

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: Nice catch. Deleting the increment from here and leaving the one in `schedule_predictions`

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: Done

dcrankshaw:
@Corey-Zumar Addressed the remaining comments

Corey-Zumar (approved):
LGTM. Merging after tests pass.

PR #130, opened 2017-04-24
[CLIPPER-162], [CLIPPER-163]: Prevent issues in conda environment creation

nishadsingh1:
This PR is a broader effort to make python prediction function deployment easy and transparent.

Part of deploying a prediction function is supplying a conda environment and creating it in the container. There are two problems that can occur here:
- [CLIPPER-163] a supplied environment can have packages with conflicting dependencies (different versions of numpy, for example)
- [CLIPPER-162] a supplied environment can have packages that don’t exist in the linux-64 conda channel (what the container will use)

This PR’s proposed solution:
- Warn users when they have packages that conflict and require them to resolve those issues before deploying their prediction function
- If no packages conflict, warn users about which packages don’t exist in the linux-64 conda channel and, in the creation of the conda environment on the container, skip installation of those packages

**Errors we are preventing in clipper_manager**
- NoPackagesFoundError: packages aren’t found in the given conda channels.
- UnsatisfiableError: packages have incompatible dependencies in the given conda channels.

**Other possible errors that can happen during package installation. We don’t prevent these directly but they shouldn’t come up.**
- PackageNotFoundError: re-raised in a catch for NoPackagesFoundError, which we prevent
- CondaImportError: re-raised in a catch for UnsatisfiableError, which we prevent
- CondaIOError: independent of package specifications
- CondaOSError: independent of package specifications
- CondaRuntimeError: independent of package specifications
- CondaSystemExit: independent of package specifications
- DirectoryNotFoundError: independent of package specifications
- LockError: independent of package specifications
- TooManyArgumentsError: only can happen when cloning an environment
- PackageNotInstalledError: only can happen when updating an environment
- CondaEnvironmentNotFoundError: can’t happen when creating a new environment
- CondaValueError: only can happen when the environment name isn’t valid, using conda force, no packages are supplied, making an environment revision, or providing multiple incompatible package specification sources. None of these apply to us
- DryRunExit: only can happen when dry-running a command that installs packages

So we can conclude that preventing NoPackagesFoundErrors and UnsatisfiableErrors ensures that we don’t see any crashes that are related to specified packages during `conda env create`.

https://clipper.atlassian.net/browse/CLIPPER-162
https://clipper.atlassian.net/browse/CLIPPER-163


dcrankshaw:
jenkins ok to test

dcrankshaw:
@nishadsingh1 Run `./bin/format_code.sh` to get your python code formatted

management/clipper_manager.py
```diff
@@ -467,28 +474,103 @@ def centered_predict(inputs):
...
+            serialized_function_file.write(serialized_prediction_function)
+        print("Serialized and supplied predict function")
+
         # Deploy function
         return self.deploy_model(name, version, serialization_dir,
                                  default_python_container, labels, input_type,
                                  num_containers)
 
+    def _conda_env_solvable(self, environment_fname, directory):
+        """Returns if the current conda environment is compatibile with container os.
```
dcrankshaw: Typo: "Returns whether" or "Returns true". Also, "compatible" is mis-spelled.

management/clipper_manager.py
```diff
@@ -467,28 +474,103 @@ def centered_predict(inputs):
...
+        environment_fname : str
+            The file name of the exported conda environment file
+        directory : str
+            The path to the diretory containing the environment file
+
+        Returns
+        -------
+        bool
+            Whether or not the (possibly modified) environment file is compatibile
+            with conda on the container os.
```
dcrankshaw: This is a style thing, but it's good to be explicit about what the meaning of true and false is.

Reword this as:
> Returns True if the (possibly modified) environment file is compatible with conda on the container os. Otherwise returns False.

dcrankshaw (changes requested):
Switch to validating environment from a subprocess call to access the conda API from the root conda environment.

management/clipper_manager.py
```diff
@@ -467,28 +469,64 @@ def centered_predict(inputs):
...
+        child = subprocess.Popen(
+            "source deactivate && python check_env.py {environment_fname} {directory} {platform}".
+            format(
+                environment_fname=environment_fname,
+                directory=directory,
+                platform=CONTAINER_CONDA_PLATFORM),
+            stderr=subprocess.PIPE,
+            shell=True)
+        child.communicate()
+        return child.returncode
```
dcrankshaw: `child.returncode` is not a boolean as the method's documentation indicates. You should check the return code before returning. If it is non-zero, print the child process's stdout and stderr and return False. If it is 0, hide the output and return True.

dcrankshaw (changes requested):
Print more diagnostic output (remember, this whole feature is to make it easier for users to debug problems). I added a specific comment inline.

management/clipper_manager.py
nishadsingh1: Thanks for the correction –  I'll change the return value.

Because we've set stderr=subprocess.PIPE, I think we do output what's printed from `check_env.py`, which does give a diagnostic message.

In the case that the user has conflicting package dependencies, we want to return False here (and stop the `deploy_prediction_func` call short).
In the case that they have no conflicting dependencies but at least one package doesn't exist in the linux channel, we want to return True (and not stop the `deploy_prediction_func` call short).

In both cases, we are outputting warning messages. So I think we want to output messages from `check_env.py` even when the return code is zero. Given this, do we want to change the output logic we have (with setting stderr to subprocess.PIPE)?

dcrankshaw: Outputting the error messages either is way a good call. But I wasn't seeing any output from the child process when I tested this. Are you seeing output?

dcrankshaw: Oh also. Remember that when you're printing things in a Python script, it's being printed to stdout, so maybe you want to capture that as well.

nishadsingh1:
Sorry for the commit weirdness...I can make a fresh PR with just one commit containing all the changes once this is functional. I'm still seeing some strange output behavior.

If I open up an interactive python session in terminal and call `deploy_prediction_func`, I _do_ see the output from `check_env` as desired:
```>>> clipper.deploy_predict_function("birds_vs_planes_classifier",2,predict,[],"doubles")
Fetching package metadata .........
Solving package specifications: .
Solving package specifications: ..........
The following packages in your conda environment aren't available in the linux-64 conda channel the container will use:
conda-forge::nltk 3.2.1 py27_0, conda-forge::libsodium 1.0.10 0, conda-forge::cloudpickle 0.2.1 py27
...
We will skip their installation when deploying your function. If your function uses these packages, the container will experience a runtime error when queried.
Removed unavailable packages from supplied environment specifications
...
```
When I use a Jupyter Notebook, I don't see any output. I'll keep trying to look into why this is the case.

dcrankshaw:
Dealing with subprocess input/output can be tricky. I'd guess that Jupyter notebooks set/override some settings that affect this behavior. Googling might be helpful here.

dcrankshaw:
@nishadsingh1 Any updates on this?

nishadsingh1:
I'm now seeing output in Jupyter notebook :) I'm going to make a clean PR with the relevant changes (and without the million random files this one is asking to merge).

PR #131, opened 2017-04-25
Add options to run unittests separately

lucasmoura:
This MR adds the option for running only a subset of Clipper tests. For example, this modification allow users to only execute the frontend tests by running:

```
./bin/run_unittests.sh -f
```

This is also true for both libclipper and management tests.

### DIFF
--- a/bin/run_unittests.sh
+++ b/bin/run_unittests.sh
@@ -4,6 +4,26 @@ set -e
 set -u
 set -o pipefail
 
+function usage {
+    cat <<EOF
+    usage: run_unittests.sh
+
+    This script is used to run Clipper tests. By default, it will run all possible
+    tests
+
+    Options:
+
+    -a, --all                   Run all tests
+    -l, --libclipper            Run tests only for libclipper folder.
+    -m, --management            Run tests only for management folder.
+    -f, --frontend              Run tests only for frotend folder.
+    -j, --java-container        Run tests only for java container folder.
+    -r, --rpc-container         Run tests only for rpc container folder.
+
+$@
+EOF
+}
+
 function clean_up {
     # Perform program exit housekeeping
     # echo Background jobs: $(jobs -l)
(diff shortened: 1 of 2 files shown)


dcrankshaw:
jenkins ok to test

Corey-Zumar:
Awesome!

dcrankshaw:
@nishadsingh1 Can you review this? Run it to make sure all the options work correctly. The code itself looks pretty clean, but take a look and comment on any problems you notice or anything you don't understand.

lucasmoura:
I have updated this MR for the new kind of tests recently added to Clipper. I have also added the new dependencies to the project Vagrantfile.

bin/run_unittests.sh
```diff
@@ -57,22 +77,70 @@ fi
...
 
-echo "Testing container RPC protocol correctness..."
-cd ../../test/
-./test_container_rpc.sh $REDIS_PORT
+case $args in
+    -a | --all )            run_all_tests
+                            ;;
+    -l | --libclipper )     run_libclipper_tests
+                            ;;
+    -m | --management )     run_management_tests
```
nishadsingh1: I think there's an extra space at the end of this line. Do you mind removing it?

bin/run_unittests.sh
```diff
@@ -57,22 +77,70 @@ fi
 
 randomize_redis_port
 
-set -e
```
nishadsingh1: Is there any reason that this was removed? It still seems reasonable to exit the testing script immediately if any subcommand exits with a nonzero status.

bin/run_unittests.sh
```diff
@@ -57,22 +77,70 @@ fi
...
+  ./src/frontends/frontendtests --redis_port $REDIS_PORT
+}
+
+function run_all_tests {
+  run_management_tests
+  redis-cli -p $REDIS_PORT "flushall"
+  run_libclipper_tests
+  redis-cli -p $REDIS_PORT "flushall"
+  run_frontend_tests
+  run_java_container_tests
```
nishadsingh1: I know that in the previous script, we didn't flush redis before running the frontend tests or java container tests. If we intend to ever add tests to java_ ontainer tests or rpc container tests that make use of redis, then it seems pretty safe and costless to add a 'flushall' command between each of these suites.

For now, it's fine to leave this as is because I don't think either of those test suites do use redis. @Corey-Zumar is there a chance they will in the future?

nishadsingh1 (changes requested):
This is really cool -- thanks for taking the time to propose these helpful changes! The requested changes are all pretty minor.

Calling ./run_unit_tests with *, -h, and --help options still results in files being built before just outputting some static usage text. Can we check first to see if any of those arguments were used (before building files), and if so, just print the text in `usage` and exit the script?

bin/run_unittests.sh
nishadsingh1: After `run_management_tests`

bin/run_unittests.sh
lucasmoura: I just removed this command because it is already defined at the beginning of the file, in line 3.

bin/run_unittests.sh
lucasmoura: Done

lucasmoura:
@nishadsingh1 About the help commands, I have created a function that sets the test environment. Now, when the user run the script with the -h command or an invalid flag, this function will not be executed and only the message in the **usage** function will be displayed.

bin/run_unittests.sh
Corey-Zumar: @nishadsingh1 Redis is being used by the rpc container test. Specifically, we subscribe to redis container changes in order to determine when a container is ready for validation. Flushing prior to this test isn't necessary to preserve correct functionality in this case, but it can't hurt.

nishadsingh1: Great, thanks @Corey-Zumar. In that case, @lucasmoura, do you mind adding a `redis-cli -p $REDIS_PORT "flushall"` line in between each of the run_<suite>_tests calls?

nishadsingh1:
@lucasmoura Yep, that makes sense and your code does create that behavior. However, if a user of the run_unittests.sh script just wants to see the usage message, we don't need to build all the files (this can take on the order of minutes). Could we check to see if a `-h`, `--help`, or `*` argument is provided before we build any of the files, and if so, just output the usage method and exit?

bin/run_unittests.sh
lucasmoura: Done

lucasmoura:
@nishadsingh1 I think that's the currently behavior. In order to build the files, it is necessary to call the function **set_test_environment** and this is not done for the **help** or any invalid flag. Therefore, for any of these options, only the **usage** function will be called.

nishadsingh1:
Hey @lucasmoura, apologies for the delay – thanks for making the changes! This LGTM.

nishadsingh1 (approved):

nishadsingh1:
@Corey-Zumar mind merging when you get the chance?

dcrankshaw:
@Corey-Zumar hold off on merging this until #134 goes in

dcrankshaw (changes requested):
Just blocking to make sure PRs get merged in the right order. @lucasmoura @nishadsingh1 no action needed on your part.

dcrankshaw (approved):

PR #133, opened 2017-04-26
[CLIPPER-137] [CLIPPER-155] [CLIPPER-158] Deploy Scala Spark models no repl support

dcrankshaw:
Waiting on a few things:

- [x] Tests
- [ ] Documentation
- [x] #121 will create massive conflicts with this PR, so let's wait to even review until that PR is merged

Creating the pull request so I get a nice diff and can start to review it myself.

PR #121, opened 2017-04-16
[CLIPPER-149] Container reconnection

Corey-Zumar:

PR #134, opened 2017-04-26
[CLIPPER-143] String/JSON Container Output

Corey-Zumar:


Corey-Zumar:
@dcrankshaw This is ready for review. Can you make sure that the tutorial still runs properly? I've verified that it works after rebuilding sklearn container, but I haven't tested with tensorflow due to preexisting import issues on my machine.

containers/java/clipper-java-container/src/main/java/clipper/container/app/ModelContainer.java
```diff
@@ -168,29 +168,61 @@ private void validateRequestInputType(Model<I> model, DataType inputType)
...
+
+    int numOutputs = predictions.size();
+    // Write the number of outputs to the buffer
+    responseBuffer.putInt(numOutputs);
+    int baseStringLengthsPosition = responseBuffer.position();
+    // We will begin writing data after the segment allocated
+    // for storing string lengths. Advance past this segment
+    // for now
+    responseBuffer.position(baseStringLengthsPosition + (4 * numOutputs));
+    for (int i = 0; i < predictions.size(); i++) {
```
dcrankshaw: s/i++/++i

containers/java/clipper-java-container/src/main/java/clipper/container/app/ModelContainer.java
```diff
@@ -168,29 +168,61 @@ private void validateRequestInputType(Model<I> model, DataType inputType)
...
+      SerializableString prediction = model.predict(dataVector);
       predictions.add(prediction);
     }
-    int outputLenBytes = 0;
-    for (FloatVector p : predictions) {
-      outputLenBytes += p.getData().remaining();
+    // At minimum, the output contains an unsigned
+    // integer specifying the number of string
+    // outputs
+    int outputLenBytes = 4;
```
dcrankshaw: All of these `4`s correspond to bytes_per_int I assume? Let's make that a const variable and use that instead to be more self-documenting

containers/java/clipper-java-container/src/main/java/clipper/container/app/data/DataUtils.java
```diff
@@ -13,31 +13,31 @@ private static ByteBuffer getByteBuffer(byte[] bytes) {
     return buffer;
   }
 
-  public static byte[] getBytesFromInts(int[] data) {
+  public static byte[] getBytesFromInts(int... data) {
```
dcrankshaw: Why the use of varargs here? Just a style thing or is there a semantic difference?

containers/java/clipper-java-container/src/main/java/clipper/container/app/data/SerializableString.java
```diff
@@ -7,13 +7,43 @@
...
     super(data);
   }
 
+  /**
+   * Writes the byte representation of the UTF-8
+   * encoded string to the provided buffer
+   *
+   * @return The number of bytes written to the buffer
+   */
+  public int toBytes(ByteBuffer buffer) {
```
dcrankshaw: Let's make this method name more descriptive. Something like `utf8StringBufferToByte` (or whatever the appropriate argument type is).

containers/python/noop_container.py
```diff
@@ -9,20 +9,20 @@ class NoopContainer(rpc.ModelContainerBase):
     def __init__(self):
         pass
 
-    def predict_ints(self, inputs):
-        return np.array([np.sum(x) for x in inputs], dtype='float32')
+    def predict_ints(self, input_item):
+        return str(sum(input_item))
```
dcrankshaw: These strs are all utf-8 encoded?

containers/python/rpc.py
```diff
@@ -168,23 +206,96 @@ def run(self):
...
+
+    def __init__(self, msg_id, num_outputs, total_string_length):
+        self.msg_id = msg_id
+        self.num_outputs = num_outputs
+        print(len(self.output_buffer))
+        self.expand_buffer_if_necessary(
+            total_string_length * MAXIMUM_UTF_8_CHAR_LENGTH_BYTES)
+        self.memview = memoryview(self.output_buffer)
+        struct.pack_into("<I", self.output_buffer, 0, num_outputs)
+        self.string_content_end_position = 4 + (4 * num_outputs)
```
dcrankshaw: Same comment about using a named constant instead of the magic number 4 here

src/frontends/src/query_frontend.hpp
```diff
@@ -386,8 +386,20 @@ class RequestHandler {
...
+      // query response
+      rapidjson::Document json_y_hat;
+      clipper::json::parse_json(query_response.output_.y_hat_, json_y_hat);
+      clipper::json::add_object(json_response, PREDICTION_RESPONSE_KEY_OUTPUT,
+                                json_y_hat);
+    } catch (const clipper::json::json_parse_error& e) {
+      // If the string output is not JSON-formatted, include
+      // it as a raw string in the query response
+      clipper::json::add_string(json_response, PREDICTION_RESPONSE_KEY_OUTPUT,
+                                query_response.output_.y_hat_);
```
dcrankshaw: I don't think this is safe. If we're including it as a raw string, we need to escape it. Otherwise, a response could mess up our JSON structure. It's not quite this bad (since it won't corrupt our system), but relevant XKCD: https://xkcd.com/327/

src/libclipper/src/datatypes.cpp
```diff
@@ -268,6 +269,27 @@ std::vector<ByteBuffer> rpc::PredictionRequest::serialize() {
...
+  char *output_string_data = reinterpret_cast<char *>(
+      bytes.data() + sizeof(uint32_t) + (num_outputs * sizeof(uint32_t)));
+  for (uint32_t i = 0; i < num_outputs; i++) {
+    uint32_t output_length = output_lengths_data[i];
+    std::string output(output_string_data, output_length);
+    outputs.push_back(std::move(output));
+    output_string_data += output_length;
+  }
+  return PredictionResponse(outputs);
+}
```
dcrankshaw, line 291: Is there a unit test for this?

dcrankshaw (changes requested):
Overall looks good. When #121 is merged and this is rebased, I'll check that the tutorial isn't broken.

containers/java/clipper-java-container/src/main/java/clipper/container/app/ModelContainer.java
Corey-Zumar: Done

containers/java/clipper-java-container/src/main/java/clipper/container/app/ModelContainer.java
Corey-Zumar: Done

containers/java/clipper-java-container/src/main/java/clipper/container/app/data/SerializableString.java
Corey-Zumar: Done - renamed to `encodeUTF8ToBuffer`

containers/python/noop_container.py
Corey-Zumar: We UTF-8 encode them before serialization. See `PredictionResponse#add_output` in `rpc.py`.

containers/python/rpc.py
Corey-Zumar: Done

src/frontends/src/query_frontend.hpp
Corey-Zumar: Ah, if I'm not mistaken, the `json::add_string` method should take care of this for us. I'll change the documentation to this effect.

Corey-Zumar (commented):
@dcrankshaw Merged the latest develop and addressed your comments. Will add the requested unit test later today

src/frontends/src/query_frontend.hpp
dcrankshaw: Cool. Can you add a unit test to add_string to make sure strings that could real json formatting (e.g. with unmatched delimiters) are properly escaped?

dcrankshaw: Sorry that was typed on my phone. Strings that could **break** json formatting are properly escaped.

src/libclipper/src/datatypes.cpp
Corey-Zumar: Added one! Renamed `input_test` to `data_serialization_test` and added `data_serialization_test#PredictionResponseDeserialization`.

src/frontends/src/query_frontend.hpp
Corey-Zumar: Done! Added `json_util_test#TestAddStringPreservesJsonIntegrity`

containers/java/clipper-java-container/src/main/java/clipper/container/app/Model.java
```diff
@@ -28,8 +28,12 @@ public DataType getInputType() {
...
-    return DataType.Floats;
+    return DataType.Strings;
   }
 
-  public abstract FloatVector predict(I inputVector);
+  /**
+   * @return A JSON-formatted serializable string to be
+   * returned to Clipper as a prediction result
+   */
+  public abstract SerializableString predict(I inputVector);
```
dcrankshaw, line 38: @Corey-Zumar Note that `Model.predict` should actually take the full batch of input vectors at once, rather than having the RPC system call `predict` in a for-loop. This allows models to apply batch processing optimizations if they want. E.g. this is particularly critical for the tensorflow model when running on a GPU.

I fixed this in the Python code but I'm leaving it as-is for now in the Java container code. I have a fix for this as part of my re-org of the Java codebase in #133.

containers/java/clipper-java-container/src/main/java/clipper/container/app/RPCTestModel.java
```diff
@@ -18,26 +19,34 @@
...
 
   @Override
-  public FloatVector predict(DoubleVector inputVector) {
+  public SerializableString predict(DoubleVector inputVector) {
     double clipperTimestamp = inputVector.getData().get();
     RPCEvent[] eventHistory = container.getEventHistory();
-    List<Float> eventCodes = new ArrayList<>();
-    for (int i = 0; i < eventHistory.length; i++) {
+    // Begin building a JSON array
+    StringBuilder eventCodeJson = new StringBuilder("[");
```
dcrankshaw, line 26: For future reference, don't manually construct JSON. Use a json serialization library.

dcrankshaw:
That may be true, but this is one place where it's critical to the whole purpose of the system to enable these batch-optimizations. Otherwise Clipper is a non-starter for any models that need to run on the GPU or use BLAS libraries. I also don't think it's a huge added burden on the developer.

containers/python/TensorFlowCifarDockerfile
```diff
@@ -2,7 +2,7 @@ FROM clipper/py-rpc:latest
 
 MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
 
-RUN pip install tensorflow
+RUN pip install tensorflow==1.0
```
dcrankshaw, line 5: This is to make sure that the tensorflow version is compatible with our pre-trained model

dcrankshaw (approved):
Okay this looks good to me. Add a comment about the batch-prediction interface and I'll wait for unit tests to pass then this is good to go.

PR #137, opened 2017-04-26
[CLIPPER-150] Remove latency objective default from add_application

Corey-Zumar:
The application-dependent latency objective is critical to the performance of the system. Users should understand the purpose and significance of the latency objective before adding an application. To this end, this PR removes the specification of a default latency objective in Clipper manager and improves documentation to aid user understanding.


management/clipper_manager.py
```diff
@@ -270,9 +266,13 @@ def register_application(self,
...
             by the end of the latency objective.
-        slo_micros : int, optional
+        slo_micros : int
             The query latency objective for the application in microseconds.
-            Default is 20,000 (20 ms).
+            If Clipper cannot process a query within the latency objective,
+            the default output is returned. Therefore, it is recommended that
+            the objective not be set aggressively low unless absolutely necessary.
+            Starting at 40000 (40 ms) is a good for testing, but the optimal
+            the optimal latnecy objective will vary depending on the application.
```
dcrankshaw: Latency is mis-spelled on the last line.

Can you add a note indicating that this is the processing latency between Clipper receiving a request and sending a response. It does not account for network latencies before a request is received or after a response is sent.

dcrankshaw (changes requested):
Good call. Added a comment to clarify the description of the field in the documentation.

management/clipper_manager.py
Corey-Zumar: Fixed typo and added note!

management/clipper_manager.py
```diff
@@ -270,9 +266,17 @@ def register_application(self,
...
-            Default is 20,000 (20 ms).
+            This is the processing latency between Clipper receiving a request 
+            and sending a response. It does not account for network latencies 
+            before a request is received or after a response is sent.
+
+            If Clipper cannot process a query within the latency objective,
+            the default output is returned. Therefore, it is recommended that
+            the objective not be set aggressively low unless absolutely necessary.
+            Starting at 40000 (40 ms) is a good for testing, but the optimal
+            the optimal latency objective will vary depending on the application.
```
dcrankshaw: You've got a couple more typos.

Rephrase and clean up the last sentence a bit:
> Starting at 40000 (40ms) is a good starting value, but the optimal latency objective will vary depending on the application.

Corey-Zumar: Done

dcrankshaw (approved):
LGTM

Issue #141, opened 2017-04-26
is the tutorial supposed to work?


miramar-labs:

Corey-Zumar:
@miramar-labs Yes, what's the issue you're facing?

miramar-labs:
i'm fairly new to jupyter notebooks but I cloned the sources, went through the steps and brought up the first tutorial .. edited cifar_loc to point to a local folder and ran the first cell .. that downloaded the cifar tar.gz. ok .. ran the 2nd cell, got an error .. took a look in extract_cifar.py and that code is looking for a subfolder called 'cifar-10-batches-py' off my download folder .. which wasn't there as the cifar tar.gz hadn't been unzipped .. hmm.. so I manually unzipped the file and got a bit further .. but then the cell that uses cifar_utils to load the data is looking for files that don't exist .. and so at this point i'm a bit confused and am wondering what noob mistake I have made..?

dcrankshaw:
Can you post the full error output you got when you tried to run the 2nd cell?

miramar-labs:
so in cell 1 I set my download folder like so:
cifar_loc = "/Users/mlabs/clipper/examples/tutorial/cifar"
ran cell 1 .. it downloaded cifar-10-python.tar.gz to that folder as expected. However, it is not extracted yet.
ran cell 2 .. (thinking it would extract it) .. but got these errors:
Unpickling data batches...
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
/Users/mlabs/clipper/examples/tutorial/extract_cifar.py in <module>()
     96         sys.exit(1)
     97     extract_to_csv(loc, *extract_cifar(loc), \
---> 98         max_train_datapoints=max_train_datapoints, max_test_datapoints=max_test_datapoints)

/Users/mlabs/clipper/examples/tutorial/extract_cifar.py in extract_cifar(loc)
     22     data, labels = [], []
     23     test_data, test_labels = [], []
---> 24     for filename in os.listdir(batches_dir):
     25         file_path = os.path.join(batches_dir, filename)
     26         if 'data_batch' in filename:

OSError: [Errno 2] No such file or directory: '/Users/mlabs/clipper/examples/tutorial/cifar/cifar-10-batches-py'

nishadsingh1:
@miramar-labs Thanks for altering us to this – there's a PR (#142) up now that attempts to solve the issue you're facing.

miramar-labs:
[...]

PR #142, opened 2017-04-27
[CLIPPER-171] Tutorial unzips downloaded tar file

nishadsingh1:
Should address the issue: https://github.com/ucbrise/clipper/issues/141

The tutorial uses Python 2, so tarfile.extractall() doesn't exist, hence the iteration through the tar object's members: http://stackoverflow.com/questions/31163668/how-do-i-extract-a-tar-file-using-python-2-4

https://clipper.atlassian.net/browse/CLIPPER-171


dcrankshaw:
jenkins test this please

dcrankshaw (approved):
LGTM. Tested and works for me.

PR #146, opened 2017-05-09
Remove unimplemented state_key_hash function and extra semicolons

### LINKED ISSUE #74: Unimplemented `state_key_hash` function.
[`size_t state_key_hash(const StateKey& key)`][1] is declared inside of `persistent_state.hpp`, but it is never implemented and never called. 

[1]: https://github.com/ucbrise/clipper/blob/910d1a31c911018203e4d1926d68659684ad9d6b/src/libclipper/include/clipper/persistent_state.hpp#L23

dcrankshaw:
Fixes #74


Corey-Zumar (approved):
LGTM

PR #147, opened 2017-05-09
[CLIPPER-162] [CLIPPER-163] [CLIPPER-146]: Check conda environment for issues in management library and fix bugs

nishadsingh1:
This is a replacement for #130

`clipper_manager.deploy_prediction_function` now inspects the user's conda environment for solvability within the container's conda channel.

If the environment on the client has packages with conflicting dependencies, we now require them to resolve those issues. Deployment of their prediction function doesn't take place.

If no packages have conflicts, we check if they all exist in the linux-64 conda channel. If some packages do not, we warn the user that those packages will not be installed on the container and update the environment file we provide the container accordingly.

Then, we deploy the function as usual.

Output in the latter case can be long:
screen shot 2017-05-09 at 2 53 53 pm

Should limit the number of missing packages we list to some small number (say, 5) and write out the full list to a file we give the user the path to?


management/clipper_manager.py
```diff
@@ -594,28 +596,67 @@ def centered_predict(inputs):
...
-            serialized_function_file.write(serialized_prediction_function)
-        print("Serialized and supplied predict function")
-
         # Export Anaconda environment
-        subprocess.call(
+        process = subprocess.Popen(
             "PIP_FORMAT=legacy conda env export >> {environment_fname}".format(
                 environment_fname=environment_fname),
             shell=True)
+        process.wait()
```
nishadsingh1, line 604: We want to make sure the environment file gets written out before we check it for problems

management/clipper_manager.py
```diff
@@ -594,28 +596,67 @@ def centered_predict(inputs):
...
+            format(
+                cur_dir=cur_dir,
+                environment_fname=environment_fname,
+                directory=directory,
+                platform=CONTAINER_CONDA_PLATFORM),
+            stdout=subprocess.PIPE,
+            shell=True)
+        out, _ = process.communicate()
+        print(out)
+        return process.returncode == 0
```
dcrankshaw: For some reason Anaconda and Jupyter are doing something funky and this didn't work for me. I found a workaround. Replace the body of this function with the following:

```py
        if "CONDA_PREFIX" not in os.environ:
            print("No Anaconda environment found")
            return False
        
        root_prefix = os.environ["CONDA_PREFIX"].split("envs")[0]
        py_path = os.path.join(root_prefix, "bin", "python")
        process = subprocess.Popen(
            "{py_path} {cur_dir}/check_env.py {environment_fname} {directory} {platform}".
            format(
                py_path=py_path,
                cur_dir=cur_dir,
                environment_fname=environment_fname,
                directory=directory,
                platform=CONTAINER_CONDA_PLATFORM),
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            shell=True)
        out, err = process.communicate()
        print(out)
        print(err)
        return process.returncode == 0
```

dcrankshaw (changes requested):
This is almost there. With the change to how we call `check_env.py`, this does the right thing both when the environment is solvable and when it is not and displays errors in a Jupyter notebook.

Can you write the temp files (`environment.yml`, the serialized model, etc) directly to the `/tmp` directory instead of first writing them to the current directory then copying them? This will avoid polluting the user's working directory.

dcrankshaw:
@nishadsingh1 I know you're busy with finals right now, so I'm going to just make these last couple changes and get this merged.

dcrankshaw:
So I've been testing this more and it doesn't quite seem to work. First of all, the containers take a really really long time to build. Like 10 minutes. And including the additional Python dependencies in the container didn't really reduce the time to create the container. But more importantly, as the code stands I can't get the deployed container to actually make predictions. I'm digging into the issue right now.

dcrankshaw:
Okay I'm not sure what's going on. This actually seems to be working. The containers still take a really long time to build though. We should definitely look into that and see if we can shorten that somehow.

dcrankshaw (approved):

PR #150, opened 2017-05-10
Make unit tests exit on fail

dcrankshaw:
We accidentally removed the option in the unit-tests script that forces the script to exit on the first error. This means that a test can fail and the unit test script can still exit with exit code 0. This patch fixes that.


nishadsingh1:
Just tested locally -- this works for me

Corey-Zumar (approved):
LGTM

PR #151, opened 2017-05-11
Add type checking to python RPC

dcrankshaw:
This adds a few checks to make sure that users' prediction functions return the appropriate type.

### DIFF
--- a/containers/python/rpc.py
+++ b/containers/python/rpc.py
@@ -100,6 +100,11 @@ class EventHistory:
     def get_events(self):
         return self.history_buffer
 
+class PredictionError(Exception):
+    def __init__(self, value):
+        self.value = value
+    def __str__(self):
+        return repr(self.value)
 
 class Server(threading.Thread):
     def __init__(self, context, clipper_ip, clipper_port):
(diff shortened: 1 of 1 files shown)


containers/python/rpc.py
```diff
@@ -119,12 +127,20 @@ def handle_prediction_request(self, prediction_request):
             predict response
         """
         predict_fn = self.get_prediction_function()
-        # outputs = []
         total_length = 0
         outputs = predict_fn(prediction_request.inputs)
+        # Type check the outputs:
+        if not type(outputs) == list:
+            raise PredictionError("Model did not return a list")
+        if not type(outputs[0]) == str:
```
Corey-Zumar: It's possible that we'll get an empty list of outputs and crash due to an out of bounds list access here. We should swap this with the length check in the execution order.

Corey-Zumar (changes requested):

containers/python/rpc.py
dcrankshaw: Good catch

Corey-Zumar (approved):
LGTM

PR #161, opened 2017-05-19
[CLIPPER-155] Restart containers, persistent & remote redis

Corey-Zumar:
Adds support for automatic restarting of containers in the case of failure. Additionally, modifies Clipper manager to support specification of a remote redis server or a path for redis data persistence (if redis was started in a docker container by the manager).

### DIFF
--- a/management/clipper_manager.py
+++ b/management/clipper_manager.py
@@ -116,8 +116,10 @@ class Clipper:
             },
             'services': {
                 'mgmt_frontend': {
-                    'command':
-                    ['--redis_ip=%s' % self.redis_ip, '--redis_port=%d' % self.redis_port],
+                    'command': [
+                        '--redis_ip=%s' % self.redis_ip,
+                        '--redis_port=%d' % self.redis_port
+                    ],
                     'image':
                     'clipper/management_frontend:latest',
                     'ports': [
@@ -129,8 +131,10 @@ class Clipper:
                     }
                 },
                 'query_frontend': {
-                    'command':
-                    ['--redis_ip=%s' % self.redis_ip, '--redis_port=%d' % self.redis_port],
+                    'command': [
+                        '--redis_ip=%s' % self.redis_ip,
+                        '--redis_port=%d' % self.redis_port
+                    ],
                     'depends_on': ['mgmt_frontend'],
                     'image':
                     'clipper/query_frontend:latest',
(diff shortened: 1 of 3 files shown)


dcrankshaw:
jenkins test this please

management/clipper_manager.py
```diff
@@ -799,14 +837,23 @@ def add_container(self, model_name, model_version):
             The name of the model
         model_version : int
             The version of the model
+
+        Returns
+        ----------
+        bool
+            Whether or not the container was added successfully
```
dcrankshaw: Be explicit about when True is returned. E.g "Returns True if the container was added successfully. Otherwise returns False."

management/clipper_manager.py
```diff
@@ -67,6 +68,18 @@ class Clipper:
         The SSH port to use. Default is port 22.
     check_for_docker : bool, optional
         If True, checks that Docker is running on the host machine. Default is True.
+    redis_port : int, optional
+        The port to use for connecting to redis. Default is port 6379.
+    redis_ip : string, optional
+        The ip address of the redis instance that Clipper should use.
+        If unspecified, a docker container running redis will be started
+        on `host` at the port specified by `redis_port`.
```
dcrankshaw, line 81: switch the order of the args redis_ip and redis_port. It's a little more natural to have ip then port.

management/clipper_manager.py
```diff
@@ -82,8 +95,11 @@ def __init__(self,
                  sudo=False,
                  ssh_port=22,
                  check_for_docker=True,
-                 redis_port=REDIS_PORT):
-
+                 redis_port=DEFAULT_REDIS_PORT,
+                 redis_ip=DEFAULT_REDIS_IP,
+                 redis_persistence_path=None,
+                 restart_containers=False):
```
dcrankshaw: Make restart_containers default to true

management/clipper_manager.py
```diff
@@ -67,6 +68,18 @@ class Clipper:
...
         If True, checks that Docker is running on the host machine. Default is True.
+    redis_port : int, optional
+        The port to use for connecting to redis. Default is port 6379.
+    redis_ip : string, optional
+        The ip address of the redis instance that Clipper should use.
+        If unspecified, a docker container running redis will be started
+        on `host` at the port specified by `redis_port`.
+    redis_persistence_path : string, optional
+        The path to which redis data should be persisted. If unspecified,
+        redis will not persist data to disk.
```
dcrankshaw: Should the path already exist, not exist, or does it matter?

dcrankshaw:
jenkins test this please

dcrankshaw (changes requested):
This is causing all kinds of problems for me. Turning on redis-persistence results in an unkillable docker container. I need to reset the Docker daemon in order to kill it. Did you see this?

management/clipper_manager.py
Corey-Zumar: Done

management/clipper_manager.py
Corey-Zumar: Done

management/clipper_manager.py
Corey-Zumar: It doesn't matter. Added a comment to this effect.

management/clipper_manager.py
Corey-Zumar: Done

Corey-Zumar:
@dcrankshaw Addressed your comments. Still unable to reproduce the unkillable container issue.

management/clipper_manager.py
```diff
@@ -67,6 +68,19 @@ class Clipper:
...
+    redis_port : int, optional
+        The port to use for connecting to redis. Default is port 6379.
+    redis_ip : string, optional
+        The ip address of the redis instance that Clipper should use.
+        If unspecified, a docker container running redis will be started
+        on `host` at the port specified by `redis_port`.
+    redis_persistence_path : string, optional
+        The directory path to which redis data should be persisted. If the
+        directory does not already exist, it will be created. If unspecified,
+        redis will not persist data to disk.
```
dcrankshaw: Okay so for some reason providing `/tmp` as redis_persistence_path is breaking docker. As a workaround for now, let's force the user to specify a path that doesn't already exist and verify that in the code.

dcrankshaw (approved):
LGTM. Will merge after #164 goes in and tests pass

PR #162, opened 2017-05-19
Reduce query frontend cpu

### LINKED ISSUE #154: query_frontend - very high CPU usage
I am on Google cloud platform server (n1-highmem-2 - 2 vCPUs, 13 GB memory) . No apps running

query_frontend is using 200% of the CPU

image

Is this normal. Do you need any other logs?

dcrankshaw:
This PR removes all busy waits and dramatically reduces the query_frontend's CPU usage when Clipper is idle. From testing on an ubuntu server, the `clipper/query_frontend` docker container uses 200% CPU while idle (no applications or models registered, no containers connect) before this change, and about 1.5% CPU after this change.

I think this should fix #154.

### DIFF
--- a/src/libclipper/include/clipper/timers.hpp
+++ b/src/libclipper/include/clipper/timers.hpp
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <queue>
 #include <thread>
+#include <condition_variable>
 
 #include <boost/thread.hpp>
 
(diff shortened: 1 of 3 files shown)


dcrankshaw:
jenkins test this please

dcrankshaw:
Hmm I broke something

Corey-Zumar (approved):
LGTM

PR #163, opened 2017-05-19
Fix integration test cleanup

dcrankshaw:
We don't properly clean up the integration test docker containers currently, which is breaking our jenkins builds


Corey-Zumar (approved):
LGTM

PR #168, opened 2017-05-20
Makes integration test length configurable

dcrankshaw:
This reduces the length of the integration tests so our PRBs run faster. It also adds back in the docker container labeling that accidentally got lost in a merge.


Corey-Zumar (approved):
LGTM

PR #171, opened 2017-05-20
Change restart policy to always for all model containers

rmdort:
Currently clipper containers restarts on reboot, but model containers don't. 

Is there a better way of doing this via `query_frontend` ? When `query_frontend` is restarted, it can check for all active applications and load the latest model containers?

### DIFF
--- a/management/clipper_manager.py
+++ b/management/clipper_manager.py
@@ -902,7 +902,7 @@ class Clipper:
             if image_name != EXTERNALLY_MANAGED_MODEL:
                 # Start container
                 add_container_cmd = (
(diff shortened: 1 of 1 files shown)


Corey-Zumar (changes requested):
Thanks for catching this! Can you check the value of the `restart_containers` parameter specified in the initialization of Clipper manager and only append the flag in the add container command if `restart_containers=True`?

dcrankshaw:
jenkins ok to test

Corey-Zumar (approved):
LGTM

PR #173, opened 2017-05-22
Added redis persistence

rmdort:
If `redis_persistence_path` is set, redis will persist data to disk.

I also removed unnecessary check for existence of redis path. Docker will use the volume if it exists.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     labels:
       - "ai.clipper.container.label"
     volumes:
-      - /tmp/clipper_redis_data:/data
+      - redisdata:/data
 
   query_frontend:
     image: "clipper/query_frontend"
@@ -36,3 +36,6 @@ services:
       - "--redis_port=6379"
     labels:
       - "ai.clipper.container.label"
+
+volumes:
+  redisdata:


management/clipper_manager.py
```diff
@@ -154,7 +154,7 @@ def __init__(self,
             self.docker_compost_dict['services']['redis'] = {
                 'image': 'redis:alpine',
                 'ports': ['%d:%d' % (self.redis_port, self.redis_port)],
-                'command': "redis-server --port %d" % self.redis_port,
+                'command': "redis-server --port %d --appendonly yes" % self.redis_port,
```
dcrankshaw: What does the `--appendonly` option do?

dcrankshaw:
jenkins ok to test

rmdort:
I got it from here https://redis.io/topics/persistence#append-only-file

Is there a better alternative?

dcrankshaw:
@Corey-Zumar you were looking into redis persistence. What did you conclude?

Corey-Zumar:
@dcrankshaw The append-only log file seems like a more fault tolerant persistence approach. However, it won't provide a solution for the problems we encountered with specifying a large pre-existing directory.

rmdort:
@dcrankshaw @Corey-Zumar 
I have reverted my changes. I am now using `docker` volumes to persist data.  It works great when clipper is launched via `docker-compose up -d query_frontend`

Docker volumes will persist data even after restart. This volume can be backed up as well. 

But `clipper_manager.py` needs to be changed to add the redis volume. We dont even need `redis_persistence_path` anymore.

If all ok, i will update `clipper_manager.py`. Let me know.

rmdort:
Strange. When i run `clipper.start()` the docker volumes gets prefixed by `tutorial_`

https://www.evernote.com/l/ABonr1F9kO5K1LM3thxk4lUha1pd5X4BN_g

Do you know whats happening?

PS: Docker seems to be adding the prefix https://github.com/docker/compose/issues/3731
Seems like we will have to create a named volume.

dcrankshaw:
@Corey-Zumar Please review this and evaluate if Docker volumes seem like a good fit for our use case.

Corey-Zumar:
@rmdort This looks like a good start, but we need to further investigate the best way to do this w.r.t refactoring Clipper to run on Kubernetes. Please see #216.

PR #180, opened 2017-05-23
Remove conda from clipper_admin pip dependencies

dcrankshaw:
With conda as a dep, when you `pip install clipper_admin` pip installs conda as well which ends up breaking conda with this message:

```
ERROR: The install method you used for conda--probably either `pip install conda`
or `easy_install conda`--is not compatible with using conda as an application.
If your intention is to install conda as a standalone application, currently
supported install methods include the Anaconda installer and the miniconda
installer.  You can download the miniconda installer from
https://conda.io/miniconda.html.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import find_packages
 
 setup(
     name='clipper_admin',
-    version='0.1.1',
+    version='0.1.2',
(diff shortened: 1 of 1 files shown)


dcrankshaw:
Sanity checked by @vsreekanti

PR #185, opened 2017-05-28
Support sudo in local mode

### LINKED ISSUE #179: Clipper_manager does not support sudo in local mode
Clipper manager should support running Docker commands that require sudo in both local and remote mode. It currently only supports this in remote mode. The relevant code is [here](https://github.com/ucbrise/clipper/blob/develop/clipper_admin/clipper_manager.py#L228-L229) and was written this way because Fabric doesn't support running local commands with sudo. We should fix this, but for now one workaround is to [add your user to the docker group](https://askubuntu.com/a/477554).

Corey-Zumar:
Fixes #179

### DIFF
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -237,6 +237,10 @@ class Clipper:
             return run(*args, **kwargs)
 
     def _execute_local(self, *args, **kwargs):
+        if self.sudo:
+            root_args = list(args)
+            root_args[0] = "sudo %s" % root_args[0]
+            args = tuple(root_args)
         # local is not currently capable of simultaneously printing and
         # capturing output, as run/sudo do. The capture kwarg allows you to
         # switch between printing and capturing as necessary, and defaults to
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -226,18 +226,18 @@ class Clipper:
         if not self.sudo:
             return self._execute_standard(*args, **kwargs)
         elif self._host_is_local():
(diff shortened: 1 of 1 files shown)


management/clipper_manager.py
```diff
@@ -227,17 +227,21 @@ def _execute_root(self, *args, **kwargs):
...
-            return self._execute_local(*args, **kwargs)
+            return self._execute_local(False, *args, **kwargs)
         else:
             return run(*args, **kwargs)
 
-    def _execute_local(self, *args, **kwargs):
+    def _execute_local(self, as_root, *args, **kwargs):
+        if self.sudo and as_root:
+            root_args = list(args)
+            root_args[0] = "sudo %s" % root_args[0]
```
dcrankshaw: Is there a reason to concat the first arg with sudo instead of just inserting "sudo" as the first argument in the list? e.g.

```py
root_args = ["sudo"] + list(args)
args = tuple(root_args)
```

dcrankshaw (changes requested):
Just a quick question. But you're also working off an old version of develop. `clipper_manager.py` is now in `clipper_admin` as part of putting together the pip package.

management/clipper_manager.py
Corey-Zumar: The first argument is the entirety of the docker command. For example, we have 
```py
args[0] = "docker-compose up -d query_frontend"
```
when we call
```py
self._execute_root("docker-compose up -d query_frontend")
```

Corey-Zumar:
@dcrankshaw Rebased and addressed your comment.

dcrankshaw (approved):
LGTM

PR #186, opened 2017-05-28
Include default prediction explanation in query response

### LINKED ISSUE #158: Return default prediction explanation in query frontend response JSON
We should return a user-readable explanation in the event that Clipper returns a JSON response containing the default prediction (i.e "no active containers available" when no container is available to field the query)

Corey-Zumar:
Fixes #158

### DIFF
--- a/src/libclipper/src/query_processor.cpp
+++ b/src/libclipper/src/query_processor.cpp
@@ -9,6 +9,7 @@
 #define PROVIDES_EXECUTORS
 #include <boost/thread.hpp>
 #include <boost/thread/executors/basic_thread_pool.hpp>
+#include <boost/optional.hpp>
 
 #include <clipper/containers.hpp>
 #include <clipper/datatypes.hpp>
@@ -68,7 +69,7 @@ boost::future<Response> QueryProcessor::predict(Query query) {
   try {
     tasks = current_policy->select_predict_tasks(selection_state, query, query_id);
   } catch(const NoModelsFoundError& e) {
-    default_explanation = "No models found for query";
+    default_explanation = "No registered models found for query";
   }
 
   log_info_formatted(LOGGING_TAG_QUERY_PROCESSOR, "Found {} tasks",
@@ -126,7 +127,7 @@ boost::future<Response> QueryProcessor::predict(Query query) {
             .count();
 
     Response response{query, query_id, duration_micros, final_output.first,
-                      final_output.second};
+                      final_output.second, default_explanation};
     response_promise.set_value(response);
   });
   return response_future;
--- a/src/libclipper/src/query_processor.cpp
+++ b/src/libclipper/src/query_processor.cpp
@@ -67,6 +67,7 @@ boost::future<Response> QueryProcessor::predict(Query query) {
   boost::optional<std::string> default_explanation;
   std::vector<PredictTask> tasks =
       current_policy->select_predict_tasks(selection_state, query, query_id);
+  
   log_info_formatted(LOGGING_TAG_QUERY_PROCESSOR, "Found {} tasks",
                      tasks.size());
 
--- a/src/libclipper/src/query_processor.cpp
+++ b/src/libclipper/src/query_processor.cpp
@@ -7,9 +7,9 @@
 #include <unordered_map>
 
 #define PROVIDES_EXECUTORS
+#include <boost/optional.hpp>
 #include <boost/thread.hpp>
 #include <boost/thread/executors/basic_thread_pool.hpp>
-#include <boost/optional.hpp>
 
 #include <clipper/containers.hpp>
 #include <clipper/datatypes.hpp>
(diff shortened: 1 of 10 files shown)


src/libclipper/src/query_processor.cpp
```diff
@@ -63,8 +64,14 @@ boost::future<Response> QueryProcessor::predict(Query query) {
...
 
-  std::vector<PredictTask> tasks =
-      current_policy->select_predict_tasks(selection_state, query, query_id);
+  boost::optional<std::string> default_explanation;
+  std::vector<PredictTask> tasks;
+  try {
+    tasks =
+        current_policy->select_predict_tasks(selection_state, query, query_id);
+  } catch (const NoModelsFoundError& e) {
+    default_explanation = "No registered models found for query";
```
dcrankshaw: Add `query_id` to output

src/libclipper/src/query_processor.cpp
```diff
@@ -94,21 +101,29 @@ boost::future<Response> QueryProcessor::predict(Query query) {
...
       if ((*r).is_ready()) {
         outputs.push_back((*r).get());
+        all_tasks_timed_out = false;
       }
     }
+    if (all_tasks_timed_out && !task_futures.empty() && !default_explanation) {
+      default_explanation =
+          "Failed to retrieve a prediction response within the specified "
+          "latency SLO";
+    }
```
dcrankshaw, line 125: Instead of throwing and catching a `NoModelsFoundError`, can you just check if `task_futures.empty()` here and use that to set the `default_explanation`? Also, set the default explanation to "No connected models found for query" instead of "No registered models found for query", because it's possible for a model to be registered with Clipper but no containers connected yet and the user will still see this message.

src/libclipper/src/selection_policies.cpp
```diff
@@ -60,6 +61,7 @@ std::vector<PredictTask> DefaultOutputSelectionPolicy::select_predict_tasks(
     log_error_formatted(LOGGING_TAG_SELECTION_POLICY,
                         "No candidate models for query with label {}",
                         query.label_);
+    throw NoModelsFoundError();
```
dcrankshaw: This isn't the right place to throw the exception. `select_predict_tasks` can request a query from a model that hasn't been registered with Clipper. In practice, this exception will never get thrown because we never allow an application to be registered that doesn't specify a candidate model. The right place to throw the exception for what you're proposing would be in [`TaskExecutor::schedule_predictions`](https://github.com/ucbrise/clipper/blob/727573caabff5fe8f1cdf4cd8ede0998797e7868/src/libclipper/include/clipper/task_executor.hpp#L339).

But I don't think throwing an exception is the right thing to do here, because that will make the TaskExecutor less forward-compatible when we support multiple models for a query and want to allow for the case where only some models have been registered at this point. See my comment above about checking for `task_futures.empty()` in query_processor.cpp for how I think we should handle this.

src/libclipper/test/selection_policies_test.cpp
```diff
@@ -31,9 +32,9 @@ TEST_F(DefaultOutputSelectionPolicyTest,
                                     1000,
                                     DefaultOutputSelectionPolicy::get_name(),
                                     {}};
-  auto zero_models_tasks =
-      policy_.select_predict_tasks(nullptr, zero_candidate_models_query, 0);
-  EXPECT_EQ(zero_models_tasks.size(), (size_t)0);
+  ASSERT_THROW(
+      policy_.select_predict_tasks(nullptr, zero_candidate_models_query, 0),
+      NoModelsFoundError);
```
dcrankshaw: When you remove the `NoModelsFoundError` don't forget to update this test.

dcrankshaw (changes requested):
Thanks for doing this. I have some feedback on how I think we should detect that there aren't any models available that I think will be more forwards-compatible.

src/libclipper/src/query_processor.cpp
Corey-Zumar: The final JSON response already contains `query_id`.

src/libclipper/src/query_processor.cpp
Corey-Zumar: Done

src/libclipper/src/selection_policies.cpp
Corey-Zumar: Thanks for the feedback. Removed the exception and added a check against `task_futures.empty()` instead.

src/libclipper/test/selection_policies_test.cpp
Corey-Zumar: Reverted test.

Corey-Zumar:

src/libclipper/src/query_processor.cpp
```diff
@@ -71,11 +73,12 @@ boost::future<Response> QueryProcessor::predict(Query query) {
 
   vector<boost::future<Output>> task_futures =
       task_executor_.schedule_predictions(tasks);
+  if (task_futures.empty()) {
+    default_explanation = "No connected models found for query";
```
dcrankshaw, line 77: Can you add a logging message here too?

dcrankshaw (changes requested):
Looks good. Just add a logging message where I commented and bring the branch up to date with develop.

src/libclipper/src/query_processor.cpp
Corey-Zumar: Done!

Corey-Zumar:
@dcrankshaw Addressed your comment. This should be ready to go.

dcrankshaw (approved):
LGTM

PR #187, opened 2017-05-30
Clipper Manager Python Tests

Corey-Zumar:
Test via:

```sh
# Execute short tests (total execution time < 1 min)
$ python /management/test/clipper_manager_test.py short
# Or execute long tests (total execution time ~ 10 mins)
$ python /management/test/clipper_manager_test.py long
# Or execute all tests
$ python /management/test/clipper_manager_test.py
```


bin/run_unittests.sh
```diff
@@ -107,6 +107,8 @@ function run_libclipper_tests {
 function run_management_tests {
   echo -e "\nRunning management tests\n\n"
   ./src/management/managementtests --redis_port $REDIS_PORT
+  cd $DIR
+  python ../management/test/clipper_manager_test.py
```
dcrankshaw: `clipper_manager.py` got moved from the `management/` to the `clipper_admin/` directory as part of making the pip package. Move the tests to the same place (`clipper_admin/test`).

management/test/clipper_manager_test.py
```diff
@@ -0,0 +1,267 @@
...
+]
+
+LONG_TEST_ORDERING = [
+    'test_deployed_model_queried_successfully',
+    'test_deployed_predict_function_queried_successfully'
+]
+
+if __name__ == '__main__':
+    run_short = True
+    run_long = True
```
dcrankshaw: There should be a `run_all` option as well

management/test/clipper_manager_test.py
```diff
@@ -0,0 +1,267 @@
...
+        model_info = self.clipper_inst.get_model_info(model_name,
+                                                      model_version)
+        self.assertIsNotNone(model_info)
+        running_containers_output = self.clipper_inst._execute_standard(
+            "docker ps -q --filter \"ancestor=clipper/python-container\"")
+        self.assertIsNotNone(running_containers_output)
+        self.assertGreaterEqual(len(running_containers_output), 1)
+
+
+class ClipperManagerTestCaseLong(unittest.TestCase):
```
dcrankshaw: Can you import `sklearn` here so that the short test case can still be run without sklearn installed?

management/test/clipper_manager_test.py
```diff
@@ -0,0 +1,267 @@
...
+import sys
+import os
+import json
+import time
+import requests
+from sklearn import svm
+cur_dir = os.path.dirname(os.path.abspath(__file__))
+sys.path.insert(0, os.path.abspath('%s/../' % cur_dir))
+import clipper_manager
+
```
dcrankshaw: Can you add a comment at the top of the file briefly describing how these tests are run? Basically just say that at the beginning of each test suite a local Clipper instance is created and the tests issue various commands against the running Clipper instance and verify their results.

This is just so that it's documented that these tests start Docker containers and interact with them, as opposed to being pure Python tests.

dcrankshaw (changes requested):
This is great! It will be really good to have test coverage of `clipper_manager` as we start refactoring it.

bin/run_unittests.sh
Corey-Zumar: Done

management/test/clipper_manager_test.py
Corey-Zumar: Both cases require`sklearn`. It is required in `test_model_deploys_successfully` (short case) and `test_deployed_model_queried_successfully` (long case).

management/test/clipper_manager_test.py
Corey-Zumar: Done

management/test/clipper_manager_test.py
Corey-Zumar: Done!

Corey-Zumar:
@dcrankshaw Addressed your comments.

bin/run_unittests.sh
```diff
@@ -107,6 +107,8 @@ function run_libclipper_tests {
 function run_management_tests {
   echo -e "\nRunning management tests\n\n"
   ./src/management/managementtests --redis_port $REDIS_PORT
+  cd $DIR
+  python ../clipper_admin/test/clipper_manager_test.py all
```
dcrankshaw: The clipper_manager tests should be run in a separate function. You may need to insert a `sleep 5` somewhere between these 2 tests too. I think the managementtests need another couple seconds to fully shut down (and unbind from the ports) which is why the test was failing.

clipper_admin/test/clipper_manager_test.py
```diff
@@ -0,0 +1,280 @@
+import unittest
```
dcrankshaw: Can you move this to `clipper_admin/tests/clipper_manager_tests.py` (just make test plural in both the directory and file name)?

clipper_admin/test/clipper_manager_test.py
```diff
@@ -0,0 +1,280 @@
...
+        elif sys.argv[1] not in ALL_ARGS:
+            args_invalid = True
+    else:
+        args_invalid = True
+
+    if args_invalid:
+        print(
+            "Missing a parameter with value 's'/'short', 'l'/'long', or 'a'/'all' indicating which subset of tests should be executed!"
+        )
+        raise
```
dcrankshaw: Raise a real exception. In fact, can you use a real option parsing library ([optparse](https://docs.python.org/3.1/library/optparse.html#module-optparse)) to deal with options.

clipper_admin/test/clipper_manager_test.py
```diff
@@ -0,0 +1,280 @@
...
+        result = self.clipper_inst.add_container(self.deploy_model_name,
+                                                 self.deploy_model_version)
+        self.assertTrue(result)
+        running_containers_output = self.clipper_inst._execute_standard(
+            "docker ps -q --filter \"ancestor=clipper/noop-container\"")
+        self.assertIsNotNone(running_containers_output)
+        split_output = running_containers_output.split("\n")
+        self.assertGreaterEqual(len(split_output), 2)
+
+    def test_predict_function_deploys_successfully(self):
```
dcrankshaw: This test will fail if not running inside an anaconda environment. I'm not sure if Jenkins runs inside an anaconda environment or not. I'll look into it but in the meantime if you see the test failing on Jenkins for this reason, that's why it's happening.

clipper_admin/test/clipper_manager_test.py
```diff
@@ -0,0 +1,280 @@
...
+
+        time.sleep(30)
+
+        url = "http://localhost:1337/{}/predict".format(self.app_name_2)
+        test_input = [99.3, 18.9, 67.2, 34.2]
+        req_json = json.dumps({'uid': 0, 'input': test_input})
+        headers = {'Content-type': 'application/json'}
+        response = requests.post(url, headers=headers, data=req_json)
+        parsed_response = json.loads(response.text)
+        self.assertNotEqual(parsed_response["output"], self.default_output)
```
dcrankshaw: Can you also check the JSON to make sure the `"default"` field is false?

dcrankshaw (changes requested):
Looks good! Some fairly minor comments.

clipper_admin/test/clipper_manager_test.py
Corey-Zumar: Done!

clipper_admin/test/clipper_manager_test.py
Corey-Zumar: Done!

clipper_admin/test/clipper_manager_test.py
dcrankshaw: Looks like optparse is deprecated. Use argparse instead.

bin/run_unittests.sh
Corey-Zumar: Made the `clipper_manager` tests a separate function (`clipper_admin_tests`) that can be executed separately with option `-c/--clipperadmin`. Inserted a sleep between `management_tests` and `clipper_admin_tests`.

clipper_admin/test/clipper_manager_test.py
Corey-Zumar: Noted! Either working with Jenkins to run the test in an anaconda environment or resolving & merging #194 is a prerequisite for merging this PR.

clipper_admin/test/clipper_manager_test.py
Corey-Zumar: Used argparse. If no arguments are specified, all tests will execute (this is documented in the usage). Removed the exception.

Corey-Zumar:
@dcrankshaw Addressed your comments.

dcrankshaw (approved):
I pushed a couple really minor cleanup changes.

I commented out the deploy_predict_function tests until #194 gets merged because dealing with conda environments in jenkins is a little messy and I don't want to be blocked on that.

Issue #188, opened 2017-06-01
Confusing parts of Clipper Manager API


jegonzal:
The following are a list of peculiarities I found in the python admin API:

- [x] The `clipper_admin` python module contains the `clipper_manager`.  Is the second _clipper_ needed?  
- [ ] The `register_application` function takes a `model` argument but must the model exist?  It seems (from the demo) that the model does not need to be present when registering it in an application.  This should probably raise an error.
- [x] Are `selection_policies` active in this release? (See `inspect_selection_policy`)?
- [ ] What are `external_models` (See `register external models`)
- [x] The use of labels is a bit unclear.

dcrankshaw:
The remaining confusing parts can be tracked in #218

PR #192, opened 2017-06-02
Assorted usability fixes

### LINKED ISSUE #175: Make clipper_manager use version tag not latest versions of Docker containers
In clipper_manager, all the Docker containers use the `latest` tag when they should be used the `0.1` tag.

### LINKED ISSUE #176: Unloading old models
When a new model is deployed, can clipper unload previous versions of that model ?

We deploy at least 10 versions of a model a day. All the old docker containers hogs the machine. Right now we manually stop the old ones. Is it possible to have an API to undeploy?

### LINKED ISSUE #184: Remove uid from prediction queries until personalization is supported
Currently, users must supply a uid as part of the prediction request body but it is never used and must always be 0. We should remove the uid field entirely. The simplest way to do this is to just stop checking for `uid` when we parse the prediction request JSON and hardcode the uid as 0.

Replace [this line](https://github.com/ucbrise/clipper/blob/dd7b42185bc974b42cf104c2503f206477001bcb/src/frontends/src/query_frontend.hpp#L443) with

```cpp
long uid = 0;
```

and update the corresponding schema [here](https://github.com/ucbrise/clipper/blob/dd7b42185bc974b42cf104c2503f206477001bcb/src/frontends/src/query_frontend.hpp#L49-L54)

dcrankshaw:
This PR makes several backwards-compatible usability fixes to make Clipper less confusing. In particular:

+ Fixes #184: Removes uid from prediction query schema which was very confusing, because the uid must be set to 0 in the current version of Clipper. Now we just set it internally.
+ Fixes #176: Adds a method `remove_inactive_containers()` to clean up inactive containers (@rmdort)
+ Addresses several of the issues in #188 
  + imports the `Clipper` class in the `clipper_admin` module, so now users can import Clipper with `from clipper_admin import Clipper`
  + Makes `inspect_selection_policy` private so it doesn't show up in API documentation
  + Makes model labels optional and removes them from the examples
+ Adds a check to `clipper_admin` for python 3 and exits with an error message if Python 3 is detected.


clipper_admin/__init__.py
```diff
@@ -0,0 +1,7 @@
+import sys
+if sys.version_info >= (3, 0):
+    sys.stdout.write(
+        "Sorry, clipper_admin requires Python 2.x, not Python 3.x\n")
```
Corey-Zumar: "Sorry, clipper_admin requires Python 2.x, but you are running Python 3.x" seems clearer

clipper_admin/clipper_manager.py
```diff
@@ -586,16 +596,18 @@ def deploy_predict_function(self,
...
             One of "integers", "floats", "doubles", "bytes", or "strings".
+        labels : list of str, optional
+            A list of strings annotating the model
         num_containers : int, optional
             The number of replicas of the model to create. More replicas can be
             created later as well. Defaults to 1.
 
         Example
         -------
+        Define a feature function ``center()`` and train a model on the featurized input::
```
Corey-Zumar, line 609: Nit: Extra colon after `input`

clipper_admin/clipper_manager.py
```diff
@@ -1031,6 +1047,44 @@ def set_model_version(self, model_name, model_version, num_containers=0):
         for r in range(num_containers):
             self.add_container(model_name, model_version)
 
+    def remove_inactive_containers(self, model_name):
+        # TODO: Test this function
```
Corey-Zumar: I'm assuming this is blocked by #187 and you'd like to merge this PR first. If not, we should use the branch associated with #187 as a base and add a test.

examples/tutorial/tutorial_part_two.ipynb
```diff
@@ -201,7 +199,6 @@
     "    2,\n",
     "    os.path.abspath(\"tf_cifar_model\"),\n",
     "    \"clipper/tf_cifar_container:latest\",\n",
```
Corey-Zumar, line 201: Should the `cifar` and `sklearn` containers reference version `0.1` as opposed to `latest` as well?

clipper_admin/clipper_manager.py
```diff
@@ -1031,6 +1047,44 @@ def set_model_version(self, model_name, model_version, num_containers=0):
...
+        num_containers_removed = 0
+        with hide("output", "warnings", "running"):
+            containers = self._execute_root(
+                "docker ps -aq --filter label={model_container_label}".format(
+                    model_container_label=CLIPPER_MODEL_CONTAINER_LABEL))
+            container_ids = [l.strip() for l in containers.split("\n")]
+            for container in container_ids:
+                # returns a string formatted as "<model_name>:<model_version>"
+                container_model_name_and_version = self._execute_root(
+                    "docker inspect --format \"{{ index .Config.Labels \\\"%s\\\"}}\" %s"
```
Corey-Zumar: This fails in the case where no model containers are actively running.

```
>>> inst.remove_inactive_containers("m1")

Fatal error: local() encountered an error (return code 1) while executing 'docker inspect --format "{{ index .Config.Labels \"ai.clipper.model_container.model_version\"}}" '

================================ Standard error ================================

"docker inspect" requires at least 1 argument(s).
See 'docker inspect --help'.

Usage:  docker inspect [OPTIONS] NAME|ID [NAME|ID...]

Return low-level information on Docker objects

================================================================================

Aborting.
```

clipper_admin/clipper_manager.py
```diff
@@ -1103,10 +1157,6 @@ def _put_container_on_host(self, container_name):
                 "docker images -q {cn}".format(cn=container_name))
 
             if len(local_result.stdout) > 0:
-                saved_fname = container_name.replace("/", "_")
-                subprocess.call("docker save -o /tmp/{fn}.tar {cn}".format(
-                    fn=saved_fname, cn=container_name))
-                tar_loc = "/tmp/{fn}.tar".format(fn=saved_fname)
                 self._execute_put(tar_loc, tar_loc)
```
Corey-Zumar: `tar_loc` is now undefined yet is being referenced. Why were these lines removed?

examples/basic_query/example_client.py
```diff
@@ -24,11 +22,10 @@ def predict(host, uid, x):
 
 if __name__ == '__main__':
     host = "localhost"
-    clipper = cm.Clipper(host, check_for_docker=False)
+    clipper = Clipper(host, check_for_docker=False)
```
Corey-Zumar, line 25: why don't we check for docker here?

Corey-Zumar (changes requested):

jegonzal (commented):
@Corey-Zumar did an excellent job reviewing.  I would approve pending his comments.

clipper_admin/__init__.py
dcrankshaw: Done

clipper_admin/clipper_manager.py
dcrankshaw: That's an intentional pydoc markup symbol (https://docs.python.org/devguide/documenting.html#source-code)

clipper_admin/clipper_manager.py
dcrankshaw: Yeah I want to add a test for this. Let's get #187 merged before this. Leaving the TODO in for now.

clipper_admin/clipper_manager.py
dcrankshaw: Oh weird. They were not supposed to be. Adding them back.

examples/basic_query/example_client.py
dcrankshaw: Because this example doesn't start Clipper, it assumes there is already a Clipper instance running and model connected. We don't check for Docker so that the example works even if you aren't using Docker to run Clipper.

examples/tutorial/tutorial_part_two.ipynb
dcrankshaw: I don't think it matters for this tutorial.

clipper_admin/clipper_manager.py
dcrankshaw: Good catch. Fixed.

clipper_admin/clipper_manager.py
Corey-Zumar: Got it, thanks for clarifying!

examples/tutorial/tutorial_part_two.ipynb
Corey-Zumar: Agreed

examples/basic_query/example_client.py
Corey-Zumar: Got it! The key part is the docker-independent functionality.

Corey-Zumar (approved):
LGTM. Let me know if I should hold off until #187 is merged. If not, we should create an issue reminding us to write a test for `remove_inactive_containers`.

dcrankshaw:
@Corey-Zumar this went through a big round of changes, so you'd better review again

bin/run_unittests.sh
```diff
@@ -123,6 +117,7 @@ function run_frontend_tests {
 function run_integration_tests {
   echo -e "\nRunning integration tests\n\n"
   cd $DIR
+  python ../integration-tests/clipper_manager_tests.py
   python ../integration-tests/light_load_all_functionality.py 2 3
```
Corey-Zumar: Now that we have multiple integration tests, let's rename `light_load_all_functionality` to something more descriptive.

clipper_admin/clipper_manager.py
```diff
@@ -441,8 +447,8 @@ def deploy_model(self,
                      version,
                      model_data,
                      container_name,
-                     labels,
                      input_type,
+                     labels=[DEFAULT_LABEL],
```
feynmanliang: What's the purpose of having all models get a "default" label? Can this be defaulted to `[]`?

clipper_admin/clipper_manager.py
```diff
@@ -1058,6 +1073,45 @@ def set_model_version(self, model_name, model_version, num_containers=0):
...
+        Parameters
+        ----------
+        model_name : str
+            The name of the model whose old containers you want to clean.
+
+        """
+        # Get all Docker containers tagged as model containers
+        num_containers_removed = 0
+        with hide("output", "warnings", "running"):
+            containers = self._execute_root(
```
feynmanliang, line 1089: What is the reasoning for dropping into shell commands over using libraries? e.g. https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.list

src/frontends/src/query_frontend_tests.cpp
```diff
@@ -101,7 +98,7 @@ TEST_F(QueryFrontendTest, TestDecodeCorrectInputDoubles) {
 
 TEST_F(QueryFrontendTest, TestDecodeCorrectInputString) {
   std::string test_json_string =
-      "{\"uid\": 23, \"input\": \"hello world. This is a test string with "
+      "{\"input\": \"hello world. This is a test string with "
       "punctionation!@#$Y#;}#\"}";
```
Corey-Zumar, line 102: Very nit: Typo in "punctuation".

Corey-Zumar (changes requested):
Just requesting that we rename an integration test file + a nit comment. Looks good in general

integration-tests/clipper_manager_tests.py
```diff
@@ -8,7 +8,7 @@
 from argparse import ArgumentParser
 cur_dir = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.abspath('%s/../' % cur_dir))
-import clipper_manager
+from clipper_admin import clipper_manager
```
feynmanliang: AFAICT `clipper_manager` is only used to access the `Clipper` class so we could directly import `Clipper` here

integration-tests/light_load_all_functionality.py
```diff
@@ -5,8 +5,8 @@
 import json
 import numpy as np
 cur_dir = os.path.dirname(os.path.abspath(__file__))
-sys.path.insert(0, os.path.abspath('%s/../clipper_admin/' % cur_dir))
-import clipper_manager as cm
+sys.path.insert(0, os.path.abspath('%s/../' % cur_dir))
+from clipper_admin import clipper_manager as cm
```
feynmanliang: Ditto about importing `Clipper` directly

clipper_admin/clipper_manager.py
dcrankshaw: Simplicity and wanting to use the same commands locally and remotely. As we refactor clipper_manager.py we should definitely look into the Python SDK.

bin/run_unittests.sh
dcrankshaw: Done

src/frontends/src/query_frontend_tests.cpp
dcrankshaw: This isn't a real word anyway. Leaving as is.

integration-tests/clipper_manager_tests.py
dcrankshaw: Ahh yeah good catch. This is a holdover from before the package structure existed.

dcrankshaw:
@Corey-Zumar @feynmanliang addressed your comments. @Corey-Zumar please merge once the tests pass.

Corey-Zumar (approved):
LGTM

dcrankshaw:
jenkins test this please

PR #194, opened 2017-06-02
Make deploy_predict_function work without Anaconda environments

### LINKED ISSUE #190: Support deploy_predict_function running outside Anaconda
If you call `deploy_predict_function` running outside of an Anaconda environment, we should still serialize the function and try to run it in the python container. Basically, we should wrap [this block](https://github.com/ucbrise/clipper/blob/4e893ddd107d5319a1cf41c699bcbcffb0ae2197/clipper_admin/clipper_manager.py#L679-L693) in an if-statement that saves and checks the anaconda environment if present, otherwise prints a warning to the user and just serializes the function.

nishadsingh1:
Fixes #190

There's a ipython notebook with testcases. I think these could eventually be moved into python integration tests -- let me know what the best way to do this is would be (they require checking logs, waiting, etc).

We'd need to update this documentation when this PR goes through: https://github.com/ucbrise/clipper-website/pull/2


nishadsingh1:
Jenkins failures relate to formatting; I'm still running into the problem of the hanging formatting script. I'll figure it out since it'll be needed a lot, but the code itself should be readable for review regardless

clipper_admin/clipper_manager.py
```diff
@@ -636,24 +636,27 @@ def centered_predict(inputs):
...
+            # Confirm that packages installed through conda are solvable
+            # Write out conda and pip dependency files to be supplied to container
+            if not (self._check_and_write_dependencies(
+                    environment_file_abs_path, serialization_dir, conda_dep_fname,
+                    pip_dep_fname)):
+                return False
+
+            print("Supplied environment details")
+        else:
+            print("Anaconda environment was either not found or failed being exported")
```
dcrankshaw: Rephrase to `print("Anaconda environment was either not found or exporting the environment failed.")`

clipper_admin/clipper_manager.py
```diff
@@ -636,24 +636,27 @@ def centered_predict(inputs):
...
+            print(
+                "Your local environment details will not be supplied to and loaded in the container in which your model is deployed."
+            )
```
dcrankshaw: Rephrase to:
"Your function will still be serialized deployed, but may fail due to missing dependencies. In this case, please re-run inside an Anaconda environment. See http://clipper.ai/documentation/python_model_deployment/ for more information."

containers/python/python_container_entry.sh
```diff
@@ -2,12 +2,40 @@
...
-/bin/bash -c "exec python /container/python_container.py"
+CONDA_DEPS_PATH="/model/conda_dependencies.txt"
+PIP_DEPS_PATH="/model/pip_dependencies.txt"
+CONTAINER_SCRIPT_PATH="/container/python_container.py"
+
+test -f $CONDA_DEPS_PATH
+dependency_check_returncode=$?
+
+if [ $dependency_check_returncode -eq 0 ]; then
+	echo "Attempting to run Python container without installing without supplied dependencies."
```
dcrankshaw: Rephrase/fix typo: "First attempting to run Python container without installing supplied dependencies."

containers/python/python_container_entry.sh
```diff
@@ -2,12 +2,40 @@
...
+	  	exit 1
+	  fi
+	else
+		echo "Encountered an ImportError when running Python container."
+		echo "Please supply necessary dependencies through an Anaconda environment and try again."
+		exit 1
+	fi
 fi
+
+echo "Encountered error not related to missing packages. Please refer to the above logs to diagnose."
```
dcrankshaw: rephrase: "Encountered error not related to missing packages. Please refer to the container logs to diagnose."

dcrankshaw (changes requested):
This is failing for me with an ImportError: `No module named clipper_admin.pywrencloudpickle`

Also, can you print out the import error here: https://github.com/ucbrise/clipper/blob/develop/containers/python/python_container.py#L121

Something like:

```py
except ImportError as e:
    print("ImportError: %s" % e)
    sys.exit(IMPORT_ERROR_RETURN_CODE)
```

dcrankshaw:
I did a bit of cleanup to make sure that this will work when the user installed `clipper_admin` via pip. It seems to be working on my computer. I want to wait for #187 to get merged and add tests for `deploy_predict_function` then this should be good to go.

dcrankshaw:
Note that these jenkins tests were intentionally killed by me.

dcrankshaw (approved):
LGTM

PR #197, opened 2017-06-04
Allow model versions to be strings

### LINKED ISSUE #191: Allow model versions to be strings
There's no real reason to enforce that model versions be numeric. We should allow them to be strings, which for example would allow versions to be SHAs (e.g. git hashes).

nishadsingh1:
Resolves #191 

The Management Frontend API now requires strings for model versions. The management library converts all model version inputs to strings, allowing it to accept non-string inputs (like ints). We could add a check to make sure that those inputs are either an int or string; that check doesn't currently exist.

The following tests in clipper_manager_test.py test this change:
test_model_version_sets_correctly
test_external_models_register_correctly
test_model_deploys_successfully


clipper_admin/clipper_manager.py
```diff
@@ -450,7 +450,7 @@ def deploy_model(self,
         ----------
         name : str
             The name to assign this model.
-        version : int
+        version : int | str
```
dcrankshaw: Let's figure out how to document this so it's clear that anything that has a string representation can be used as a version.

clipper_admin/clipper_manager.py
```diff
@@ -552,13 +553,14 @@ def register_external_model(self, name, version, labels, input_type):
...
```
dcrankshaw: Ditto here about documentation.

src/frontends/src/query_frontend.hpp
```diff
@@ -224,18 +224,16 @@ class RequestHandler {
...
         current_model_versions_[model_name] = model_version;
       } else {
-        clipper::log_error_formatted(
-            LOGGING_TAG_QUERY_FRONTEND,
-            "Found model {} with invalid version number {}.", model_name,
-            model_version);
+        clipper::log_error_formatted(LOGGING_TAG_QUERY_FRONTEND,
+                                     "Found model {} with invalid version {}.",
+                                     model_name, model_version);
         throw std::runtime_error("Invalid model version number");
```
dcrankshaw: This error message should just be "Invalid model version"

src/frontends/src/query_frontend.hpp
```diff
@@ -224,18 +224,16 @@ class RequestHandler {
...
       }
-      model_names_with_version.push_back(model_name + "@v" +
-                                         std::to_string(model_version));
+      model_names_with_version.push_back(model_name + "@v" + model_version);
```
dcrankshaw: remove the "v" here, just `model_name + "@" + model_version`

src/frontends/src/query_frontend_tests.cpp
```diff
@@ -283,24 +283,25 @@ TEST_F(QueryFrontendTest, TestReadModelsAtStartup) {
   // Add multiple models (some with multiple versions)
   std::vector<std::string> labels{"ads", "images", "experimental", "other",
                                   "labels"};
-  VersionedModelId model1 = std::make_pair("m", 1);
+  VersionedModelId model1 = std::make_pair("m", "1");
```
dcrankshaw: Let's make some of these version numbers more exotic. Use things that look like git hashes (92248e30a93fa60ae3baa598f21c736e26c1a371), semantic versions ("1.2.3"), and some strings with spaces and punctuation.

src/libclipper/include/clipper/datatypes.hpp
```diff
@@ -11,7 +11,7 @@
 namespace clipper {
 
 using ByteBuffer = std::vector<uint8_t>;
-using VersionedModelId = std::pair<std::string, int>;
+using VersionedModelId = std::pair<std::string, std::string>;
```
dcrankshaw: Can you actually make `VersionedModelId` a class? This will be a bit of extra work, but I've been meaning to do this anyway. The class should have 2 private members (name and id), and should have at least the following methods:
```cpp
std::string get_name() const;
std::string get_id() const;
std::size_t hash() const;
std::string serialize() const;
static VersionedModelId deserialize(std::string);
```

Also make sure you implement the copy and move constructors as `default` ([like this](https://github.com/ucbrise/clipper/blob/develop/src/libclipper/include/clipper/datatypes.hpp#L44-L48))

Then you can replace all the references to `versioned_model_hash` with `VersionedModelId::hash`.

src/libclipper/src/redis.cpp
```diff
@@ -157,46 +155,46 @@ std::vector<VersionedModelId> str_to_models(const std::string& model_str) {
...
 
     return send_cmd_no_reply<string>(redis, cmd_vec);
   } else {
     return false;
   }
 }
 
-int get_current_model_version(redox::Redox& redis,
-                              const std::string& model_name) {
+std::string get_current_model_version(redox::Redox& redis,
```
dcrankshaw: This is a good place to return an option (see `boost::optional`) instead of an empty string.

src/frontends/src/query_frontend_tests.cpp
dcrankshaw: Actually, are there any characters we need to restrict from being part of the model version? I think at some point we use a character delimiter when parsing a model name and version from a string. If the model name or version contains those characters, the parsing will fail.

@nishadsingh1 Can you look into this? If that is the case, we should proactively prevent model names and versions that contain those restricted characters from being registered.

src/libclipper/src/datatypes.cpp
```diff
@@ -21,7 +21,7 @@ std::string versioned_model_to_str(const VersionedModelId &model) {
   std::stringstream ss;
   ss << model.first;
   ss << ":";
-  ss << std::to_string(model.second);
+  ss << model.second;
```
dcrankshaw: This should become a method in the `VersionedModelId` class

src/libclipper/src/redis.cpp
```diff
@@ -145,10 +145,8 @@ std::vector<VersionedModelId> str_to_models(const std::string& model_str) {
         start +
         model_str.substr(start, end - start).find(ITEM_PART_CONCATENATOR);
```
dcrankshaw, line 158: This method will break if `ITEM_PART_CONCATENATOR` or `ITEM_DELIMITER` are characters in the model name or version. We should check to make sure that when models are registered, the name and version don't contain these characters. We should have already been doing this but looks like it got overlooked.

src/libclipper/src/redis.cpp
dcrankshaw: You can grep the codebase to see some examples of returning and checking options.

src/libclipper/test/json_util_test.cpp
```diff
@@ -159,30 +159,30 @@ TEST(JsonUtilTests, TestParseCandidateModels) {
   std::string correct_json = R"(
   {
     "correct_candidate_models": [
-      {"model_name": "sklearn_svm", "model_version": 1},
-      {"model_name": "sklearn_svm", "model_version": 2},
-      {"model_name": "network", "model_version": 3}
+      {"model_name": "sklearn_svm", "model_version": "1"},
+      {"model_name": "sklearn_svm", "model_version": "2"},
+      {"model_name": "network", "model_version": "3"}
```
dcrankshaw: Use some version strings that contain more characters in this test.

src/management/src/management_frontend.hpp
```diff
@@ -458,7 +457,7 @@ class RequestHandler {
     parse_json(json, d);
 
     VersionedModelId model_id = std::make_pair(get_string(d, "model_name"),
```
dcrankshaw: This is the right place to validate that model names and versions don't contain restricted characters.

dcrankshaw: You should add corresponding tests in the `management_frontend_tests.cpp` test suite.

dcrankshaw (changes requested):
This looks great! My comments are mostly about reducing our technical debt as we improve the code base.

dcrankshaw:
To fix the hash problem, write the hash function as demonstrated below (taken from the [C++ documentation](http://en.cppreference.com/w/cpp/utility/hash)) and remove all references to `versioned_model_hash` when defining or constructing `unordered_map`. Note that when defining this, you need to be outside the `clipper` namespace before starting the `namespace std` block.

```cpp
// custom specialization of std::hash can be injected in namespace std
namespace std
{
    template<> struct hash<VersionedModel>
    {
        typedef std::size_t result_type;
        std::size_t operator()(VersionedModel const& vm) const
        {
            std::size_t seed = 0;
            boost::hash_combine(seed, vm.name_);
            boost::hash_combine(seed, vm.id_);
            return seed;
        }
    };
}
```

nishadsingh1:
All tests are passing locally for me -- can we run them again?

dcrankshaw:
The format checker failed. Format the code and push again.

dcrankshaw:
Okay this looks like a real test failure: https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/447/console. The problem is this line: https://github.com/ucbrise/clipper/pull/197/files#diff-b2d2cbaaab09d37b17643a41d168d43fR81

Don't call the constructor like that, just call it like `return VersionedModelId(model_name, model_version);`

nishadsingh1:
Looks like there are more legitimate test failures -- thanks for catching that one. I'll make the changes and ping you again if it doesn't build and I'm unsure why.

clipper_admin/clipper_manager.py
dcrankshaw: Make the `version` method doc consistent with the rest of the methods

clipper_admin/clipper_manager.py
```diff
@@ -1101,14 +1105,16 @@ def set_model_version(self, model_name, model_version, num_containers=0):
         ----------
         model_name : str
             The name of the model
-        model_version : int
+        model_version : int | str
```
dcrankshaw: Make the method doc consistent with the rest of the documentation

src/frontends/src/query_frontend.hpp
```diff
@@ -178,15 +178,15 @@ class RequestHandler {
               event_type);
           if (event_type == "set") {
             std::string model_name = key;
-            int new_version = clipper::redis::get_current_model_version(
+            std::string new_version = clipper::redis::get_current_model_version(
                 redis_connection_, key);
-            if (new_version >= 0) {
+            if (new_version.size() > 0) {
```
dcrankshaw: Why would model version size be 0?

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -364,11 +357,11 @@ class TaskExecutor {
   std::shared_ptr<metrics::Meter> throughput_meter_;
   boost::shared_mutex model_queues_mutex_;
   std::unordered_map<const VersionedModelId, std::shared_ptr<ModelQueue>,
-                     decltype(&versioned_model_hash)>
+                     std::hash<clipper::VersionedModelId>>
```
dcrankshaw: I think the reason you need to provide the hash function explicitly here is because the keys are marked const. Does this work if you remove const?

```cpp
std::unordered_map<VersionedModelId, std::shared_ptr<ModelQueue>> model_queues_;
```

src/libclipper/src/redis.cpp
dcrankshaw: @nishadsingh1 Please return an [`optional`](http://www.boost.org/doc/libs/1_64_0/libs/optional/doc/html/index.html) here instead of an empty string.

src/libclipper/src/task_executor.cpp
```diff
@@ -79,7 +79,8 @@ void PredictionCache::put(const VersionedModelId &model,
 
 size_t PredictionCache::hash(const VersionedModelId &model,
                              size_t input_hash) const {
-  return versioned_model_hash(model) ^ input_hash;
+  size_t model_hash = std::hash<clipper::VersionedModelId>()(model);
+  return model_hash ^ input_hash;
```
dcrankshaw: This should be replaced with `boost::hash_combine`

src/management/src/management_frontend_tests.cpp
```diff
@@ -53,10 +53,26 @@ class ManagementFrontendTest : public ::testing::Test {
...
+    add_string(d, "input_type", input_type);
+    add_string(d, "container_name", container_name);
+    add_string(d, "model_data_path", model_data_path);
+  }
+
   std::string get_app_json_request_string(std::string& name) {
     rapidjson::Document d;
     d.SetObject();
-    add_string(d, "name", name);
+    add_string(d, "name", name);  // WANT TO REMOVE THIS?
```
dcrankshaw: Why would you want to remove this?

dcrankshaw (changes requested):
Looks good. A few more changes.

src/frontends/src/query_frontend.hpp
nishadsingh1: If it's the empty string "" -- we could do the full check for validity, but it might be redundant with the validation in management frontend: https://github.com/ucbrise/clipper/pull/197/files#diff-9d52ffc36b86c3e051c47bd5971c718dR506

dcrankshaw: I want `clipper::redis::get_current_model_version` to return a `boost::optional<std::string>` as noted in a comment below, so the correct check here is whether the `optional` contains a value.

nishadsingh1: Got it, great

dcrankshaw (approved):

dcrankshaw:
LGTM. Will merge once tests pass.

PR #203, opened 2017-06-07
[CLIPPER-155] Fix redis persistence

Corey-Zumar:
This should solve the previously encountered stalling issues that occur when certain pre-existing directories are specified as the redis persistence path. 

Context:
When the `user` parameter in the docker compose dictionary is left unspecified, docker attempts to execute the redis run command as root (w.r.t the docker container, not the host's system), resulting in process attempting to `chown` the volume specified for persistence. In the case where insufficient permissions exist to obtain ownership, errors and/or stalls are encountered.

Solution:
Specify the `uid` and `gid` in `docker_compose_dict` to ensure that the correct privilege level is used when executing redis operations.

### DIFF
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -161,11 +161,14 @@ class Clipper:
         start_redis = (self.redis_ip == DEFAULT_REDIS_IP)
         if start_redis:
             self.docker_compose_dict['services']['redis'] = {
-                'image': 'redis:alpine',
+                'image':
+                'redis:alpine',
                 'ports': ['%d:%d' % (self.redis_port, self.redis_port)],
-                'command': "redis-server --port %d --appendonly yes" % self.redis_port,
+                'command':
+                "redis-server --port %d --appendonly yes" % self.redis_port,
                 # Set the user id and group id to set command execution privileges
-                'user': "%s:%s" % (str(os.getuid()), str(os.getgid())),
+                'user':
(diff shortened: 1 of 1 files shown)


clipper_admin/clipper_manager.py
```diff
@@ -160,37 +160,35 @@ def __init__(self,
...
+            self.docker_compose_dict['services']['redis'] = {
+                'image':
+                'redis:alpine',
                 'ports': ['%d:%d' % (self.redis_port, self.redis_port)],
-                'command': "redis-server --port %d" % self.redis_port,
+                'command':
+                "redis-server --port %d --appendonly yes" % self.redis_port,
+                # Set the user id and group id to set command execution privileges
+                'user':
+                "%s:%s" % (str(os.getuid()), str(os.getgid())),
```
dcrankshaw, line 171: I believe the reason this is failing is that the user running this command is often not the same user and doesn't have the same permissions as the docker user. For example, on Jenkins `os.getuid()` would return the uid for the `jenkins` user, but the Docker daemon runs as `root`.

dcrankshaw:
Thinking about this further, dealing with Redis persistence by mounting a directory from the local filesystem is the wrong way to go about managing Redis persistence. @rmdort had the right idea with using Docker volumes, which can also be migrated. Let's hold off on making this change and instead incorporate a more robust Redis deployment as part of the migration to Kubernetes.

PR #204, opened 2017-06-08
Throughput benchmarking

nishadsingh1:
This PR contains the script for throughput benchmarking. It's not complete; I thought it'd be a good idea to put it up here early for the sake of collaboration/commenting.

The benchmarking script:
1. Asks for some inputs (hyperparams like number of batches of queries / sleep time, location of the CIFAR dataset)
2. Sends queries, in a manner specified by the inputs, to a connected Noop model. It records metrics throughout the process.
3. Outputs those metrics

The PR also refactors some code and makes minor updates to the docs and code for end_to_end_bench.

Remaining TODOs before merging:
1. Support variable batch sizes (having trouble figuring out how to do this)
2. Measure/report the P99 throughput. Right now, only the net throughput of all queries is being reported. Variables that can be measured on a per-query basis (like latency) are placed in a histogram that can report out P99 values. In the case of throughput, it seems like we need to find some way to define a throughput datapoint (maybe throughput over the window of the last X queries) so we can add those to a similar histogram.
3. Update documentation


bench/setup_bench.sh
```diff
@@ -9,10 +9,10 @@ trap "kill 0" EXIT
...
 
-python bench_init.py $cifar_path
+python bench/bench_init.py $cifar_path
 
 export CLIPPER_MODEL_NAME="bench_sklearn_cifar"
 export CLIPPER_MODEL_VERSION="1"
 export CLIPPER_MODEL_PATH="model/"
 
-python ../containers/python/sklearn_cifar_container.py 
+python containers/python/sklearn_cifar_container.py
```
nishadsingh1, line 18: Changes to this file were made so that the usage specified in the end_to_end_bench script, namely that setup_bench.sh should be run from the root clipper folder, works. Before it worked when run from the /bench folder.

shaneknapp:
test this please

shaneknapp:
a 28 min build is much better than ~2 hours.  i think we fixed the network
throughput.

nishadsingh1:
**Instructions on how to run benchmarking:**
1. Pull the nishadsingh1/benchmarking branch

2. Get the model-container up and running:
 - If you want to have the model serving off your machine, run `./bench/setup_same_prediction_bench.sh `
 - If you want the model serving off a docker container, run build the image using `./bin/build_bench_docker_images.sh` and do `docker run clipper/same-prediction-bench`

3. Modify batch size. Modify `FIXED_BATCH_SIZE` in `src/logging_constants.hpp`. Configure for release and `make`.
4. Modify other configuration settings in `config.json`. The fields you might want to change are `batch_delay_micros`, which controls how long to sleep between each issued request, and `reports_path` & `reports_path_verbose` which specify where the output will go.
5. Run `./release/src/benchmarks/noop_bench -f config.json` and break out of the script whenever you feel you have a sufficient number of datapoints in your reports.

I noticed that when I set `FIXED_BATCH_SIZE` >= 13, I could not saturate the model queue, even when setting `batch_delay_micros` to "0". You can check what actual batch size is getting used by looking at the file at `reports_path_verbose` and reading off the information under "model:bench_same_prediction:1:batch_size" entries.

@dcrankshaw

nishadsingh1:
Closing -- all functionality in this branch that isn't now in develop can be found in #253.

PR #195, opened 2017-06-02
R model deployment

dubeyabhi07:
This pull request will integrate R model deployment with clipper.

PR #206, opened 2017-06-08
Deploy containers to Kubernetes

feynmanliang:
* Adds K8s as a hard dependency, using `minikube` for local development
* Removes `docker-compose` dependency (all Clipper resources are provided on the k8s cluster)
* ~~Removes `fabric` dependency (remote administration is done through K8s api now~~ The R container that was added (#195) reintroduced this
* Adds `ClipperK8s` class to handle Kubernetes interface

To test:

* Configure local k8s cluster using `minikube`
```
minikube start --insecure-registry localhost:5000
eval $(minikube docker-env)
```

* Run the following code:
```python
from subprocess import Popen, PIPE
import numpy as np
import clipper_admin

# Sets host to be minikube's virtualbox network interface
clipper_host_public_ip = Popen(['minikube', 'ip'], stdout=PIPE).communicate()[0].strip()

clipper = clipper_admin.Clipper(host=clipper_host_public_ip)
clipper.start() # starts clipper query managers in k8s
```
Clipper and its components should be deployed to minkube
* Run `kubectl get svc` to get the k8s service port mappings. Substitute in the corresponding values into `clipper_manager.py` (this could be improved by making it a constructor argument)

* To deploy to clipper on k8s:
```
clipper.register_application("hello_world", "feature-sum-model", "doubles", "-1.0", 100000)
def feature_sum_function(xs):
      return [str(np.sum(x)) for x in xs]

# serializes predict function, builds docker images, creates k8s deployment which registers with query manager
clipper.deploy_predict_function("feature-sum-model", 5, feature_sum_function, "doubles")
```

* Test the endpoint by POSTing queries to `$(minikube ip):1337/hello_world/predict`

* Cluster resources can be deprovisioned with `clipper.stop_all()`

---
This change is [Reviewable](https://reviewable.io/reviews/ucbrise/clipper/206)


dubeyabhi07:
Review status: 0 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[clipper_admin/clipper_k8s.py, line 16 at r2](https://reviewable.io:443/reviews/ucbrise/clipper/206#-Km_Lzu0CTzSzfg4ZwHC:-Km_Lzu0CTzSzfg4ZwHD:bd3q455) ([raw file](https://github.com/ucbrise/clipper/blob/4355910905bfb005c1eb493478d11f0854e1200d/clipper_admin/clipper_k8s.py#L16)):*
> ```Python
>         # self.initialize_clipper() # NOTE: this allows containers to discover query_manager by DNS rather than IP,
>         #                           # but may couple too tightly to k8s
>         self.initialize_registry() # TODO: check doesn't exist before trying
> ```

initialize_registry() should be after declaring k8s_v1 and k8s_beta

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucbrise/clipper/206)*

dubeyabhi07:
Review status: 0 of 16 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[clipper_admin/clipper_k8s.py, line 100 at r2](https://reviewable.io:443/reviews/ucbrise/clipper/206#-KnA_yM2dFMub5mRTP17:-KnA_yM2dFMub5mRTP18:btwsrnq) ([raw file](https://github.com/ucbrise/clipper/blob/4355910905bfb005c1eb493478d11f0854e1200d/clipper_admin/clipper_k8s.py#L100)):*
> ```Python
>                                                 {
>                                                     'name': 'CLIPPER_IP',
>                                                     'value': '10.0.2.2' # TODO: WTF magic IP that goes to host
> ```

It should be 'value' : 'query-frontend'. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucbrise/clipper/206)*

feynmanliang:
@dcrankshaw ready for review

dcrankshaw:
I'm currently reviewing this. I'm actually refactoring it a bit and adding some additional functionality as well, which I will submit as a pull request *to this branch*. So let's treat this as basically a feature branch for now.

From what I've looked at so far, this looks great. The functionality I'm adding is mostly refactoring the python package and adding plain Docker support so we don't break existing workflows when this gets merged.

dcrankshaw:
Replaced by #251

PR #214, opened 2017-06-15
Base64 decoding for JSON byte data

Corey-Zumar:
This change is [Reviewable](https://reviewable.io/reviews/ucbrise/clipper/214)


.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "bin/yapf"]
 	path = bin/yapf
 	url = https://github.com/google/yapf.git
+[submodule "src/libs/base64"]
```
dcrankshaw: Since this is a header-only library, instead of adding it as a submodule, can you just copy the header into src/libcs/base64.

Also, since the library already has tests, it would be good to port them to our testing infrastructure as well. I took a quick look at them, and it doesn't look like it will be too hard.

src/libclipper/src/json_util.cpp
```diff
@@ -129,6 +131,23 @@ std::string get_string(rapidjson::Value& d, const char* key_name) {
...
+  if (!v.IsString()) {
+    throw json_semantic_error("Input of type " + kTypeNames[v.GetType()] +
+                              " is not of type base64-encoded string");
+  }
+  Base64 decoder;
+  std::string encoded_string = std::string(v.GetString());
+  std::string decoded_string;
+  decoder.Decode(encoded_string, &decoded_string);
+  std::vector<uint8_t> decoded_bytes = std::vector<uint8_t>(
+      decoded_string.data(), decoded_string.data() + decoded_string.length());
```
dcrankshaw: What is the second argument here `decoded_string.data() + decoded_string.length()` for?

src/libclipper/src/json_util.cpp
```diff
@@ -256,7 +275,8 @@ std::shared_ptr<Input> parse_input(InputType input_type, rapidjson::Value& d) {
       return std::make_shared<clipper::SerializableString>(input_string);
     }
     case InputType::Bytes: {
-      throw std::invalid_argument("Base64 encoded bytes are not supported yet");
```
dcrankshaw, line 259: 👍

dcrankshaw (changes requested):
Looks good. Minor comments.

src/libclipper/src/json_util.cpp
Corey-Zumar: This was the second argument of the range constructor, specifying the pointer to the end of the memory to be allocated to the vector. It's cleaner to use `decoded_string.begin()` and `decoded_string.end()` instead - made that change.

.gitmodules
Corey-Zumar: Yep, removed the submodule and just copied the header. Ported the tests to `base64_tests.cpp`.

Corey-Zumar (commented):

dcrankshaw:
jenkins test this please

dcrankshaw:
LGTM. Waiting for the tests to pass.

dcrankshaw (approved):

Issue #218, opened 2017-06-16
Refactor model and application registration and linking


dcrankshaw:
There is some confusion and some limitations around the way that models and applications are registered with Clipper.

- [x] Currently applications specify the name of the model that they will query. This is a little bit awkward and requires the developer registering the application to know the name of the model, even if it hasn't been registered yet. Instead, one proposal is to register model and applications independently. When a user wants to associate a specific model with an application, they could call a `link_model(app_name, model_name)` to associate the two. If we go with this proposal, what are the semantics of `link` and a potential `unlink`. For example, during linking we should type check to make sure that the application and model have compatible input types.

- [ ] Users have expressed some confusion over what `input_type` means and how to use it or set it appropriately. This can be fixed with better documentation and examples.

- [x] For simple applications and demos, we should support a `clipper_admin` method that both registers an application and deploys a model at once. This lets you call a single method to start serving a model.

nishadsingh1:
https://github.com/nishadsingh1/clipper/tree/links is a WIP branch that will address the first checkbox in this issue

dcrankshaw:
Fixed by #226

PR #223, opened 2017-06-20
One-line app registration and model deployment

nishadsingh1:
Addresses the third checkbox in #218

### DIFF
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -321,7 +321,7 @@ class Clipper:
 
     def register_application(self, name, model, input_type, default_output,
                              slo_micros):
-        """Register a new Clipper application.
+        """Register a new Clipper application and returns the response object.
 
         Parameters
         ----------
@@ -345,6 +345,12 @@ class Clipper:
             the objective not be set aggressively low unless absolutely necessary.
             40000 (40ms) is a good starting value, but the optimal latency objective
             will vary depending on the application.
+
+        Returns
+        -------
+        Requests object
+            The response object for the app registration request
+
         """
         url = "http://%s:%d/admin/add_app" % (self.host,
                                               CLIPPER_MANAGEMENT_PORT)
@@ -358,6 +364,7 @@ class Clipper:
         headers = {'Content-type': 'application/json'}
         r = requests.post(url, headers=headers, data=req_json)
         print(r.text)
+        return r
 
     def get_all_apps(self, verbose=False):
         """Gets information about all applications registered with Clipper.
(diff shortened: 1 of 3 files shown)


clipper_admin/clipper_manager.py
```diff
@@ -345,6 +345,12 @@ def register_application(self, name, model, input_type, default_output,
             the objective not be set aggressively low unless absolutely necessary.
             40000 (40ms) is a good starting value, but the optimal latency objective
             will vary depending on the application.
+
+        Returns
+        -------
+        Requests object
+            The response object for the app registration request
```
dcrankshaw: What is a response object?

clipper_admin/clipper_manager.py
```diff
@@ -733,6 +740,80 @@ def centered_predict(inputs):
 
         return deploy_result
 
+    def _register_app_and_check_success(self, app_name, model_name, input_type,
+                                        default_output, slo_micros):
+        response = self.register_application(app_name, model_name, input_type,
+                                             default_output, slo_micros)
+        if response.status_code != 200:
```
dcrankshaw: Don't check for 200, use `if r.status_code == requests.codes.ok:`

clipper_admin/clipper_manager.py
```diff
@@ -733,6 +740,80 @@ def centered_predict(inputs):
...
+            print(
+                "Application registration unsuccessful. Will not deploy model.")
+            return False
+        print("Application registration sucessful! Deploying model.")
+        return True
+
+    def register_app_and_deploy_model(self,
```
dcrankshaw: These are really really long method names.

dcrankshaw (changes requested):
So now that I'm looking at the code, these method signatures still look really complicated. These are basically convenience methods to make things simpler. If the user wants more control, she can call `register_application` and deploy_*` separately.

To simplify the method signatures:
1. Make the model name and app name the same and just have a single name parameter
2. Remove the method for `register_app_and_deploy_model`, just have it for python functions and pyspark models
3. Make the model version arg have a default argument of 1 (`model_version=1`)
4. Make slo_micros default to 100000
5. Make `input_type` the last non-default arg (so the order is name, predict_function, [pyspark model and spark context if deploy_pyspark method], input_type, default args)

nishadsingh1:
Makes sense -- do we want to give `default_output` a default as well? Maybe "-1.0" or something?

nishadsingh1:
Didn't make the change to have a default `default_output` -- happy to add it in. The new function names `activate_predict_function` and `activate_pyspark_model` still don't seem quite right...let me know if better names come to mind.

clipper_admin/clipper_manager.py
```diff
@@ -44,6 +44,8 @@
 CLIPPER_DOCKER_LABEL = "ai.clipper.container.label"
 CLIPPER_MODEL_CONTAINER_LABEL = "ai.clipper.model_container.model_version"
 
+DEFAULT_MODEL_VERSION = 1
+DEFAULT_SLO_MICROS = 10000
```
dcrankshaw: This is 10 ms, make it 100 ms (100000)

clipper_admin/clipper_manager.py
```diff
@@ -345,6 +347,12 @@ def register_application(self, name, model, input_type, default_output,
             the objective not be set aggressively low unless absolutely necessary.
             40000 (40ms) is a good starting value, but the optimal latency objective
             will vary depending on the application.
+
+        Returns
+        -------
+        bool
+            Returns true iff the app reigstration request was successful
```
dcrankshaw: typo in "registration"

clipper_admin/clipper_manager.py
```diff
@@ -735,6 +744,105 @@ def centered_predict(inputs):
...
+    def _register_app_and_check_success(self, name, input_type, default_output,
+                                        slo_micros):
+        if self.register_application(name, name, input_type, default_output,
+                                     slo_micros):
+            print("Application registration sucessful! Deploying model.")
+            return True
+        print("Application registration unsuccessful. Will not deploy model.")
+        return False
+
+    def activate_predict_function(self,
```
dcrankshaw: Activate isn't really the right verb. I can't think of a good shorter name that doesn't conflict with our existing API, so maybe just use the long names from before.

clipper_admin/clipper_manager.py
```diff
@@ -735,6 +744,105 @@ def centered_predict(inputs):
...
+                                  name,
+                                  predict_function,
+                                  default_output,
```
dcrankshaw: Make the default output "None"

clipper_admin/clipper_manager.py
```diff
@@ -735,6 +744,105 @@ def centered_predict(inputs):
...
+        return self.deploy_predict_function(name, model_version,
+                                            predict_function, input_type,
+                                            labels, num_containers)
+
+    def activate_pyspark_model(self,
+                               name,
+                               predict_function,
+                               pyspark_model,
+                               sc,
+                               default_output,
```
dcrankshaw: Make the default output "None"

clipper_admin/clipper_manager.py
```diff
@@ -735,6 +744,105 @@ def centered_predict(inputs):
...
+                                  slo_micros=DEFAULT_SLO_MICROS,
+                                  labels=DEFAULT_LABEL,
+                                  num_containers=1):
+        """Registers an app and deploys provided predict function as a model.
+
+        Parameters
+        ----------
+        name : str
+            The to be assigned to the registered app and deployed model.
+        predict_function : function
```
dcrankshaw, line 773: Don't say "refer to documentation in ..." That is really confusing and leads to hard to read docs. Just copy the text.

As we refactor this documentation, we'll create some module-level documentation about deploying models and the meaning of these fields.

clipper_admin/clipper_manager.py
```diff
@@ -735,6 +744,105 @@ def centered_predict(inputs):
...
+                               slo_micros=100000,
+                               labels=DEFAULT_LABEL,
+                               num_containers=1):
+        """Registers an app and deploys provided spark model.
+
+        Parameters
+        ----------
+        name : str
+            The to be assigned to the registered app and deployed model.
+        predict_function : function
```
dcrankshaw, line 821: Fix docs (remove "refer to...")

dcrankshaw (changes requested):
Mostly there. Just some naming and doc cleanup.

dcrankshaw:
LGTM. Merge with develop (looks like you'll need to do that locally) and I'll do a last pass and get this merged.

dcrankshaw (approved):

PR #225, opened 2017-06-20
Benchmarking utilities

### LINKED ISSUE #222: Fix task latency measurement for better batching behavior
While benchmarking (#204) Clipper, @nishadsingh1 discovered that the latency measurements used to estimate container throughput and therefore batch size include the time a task spends in the model queue waiting to be executed. As a result, we underestimate batch sizes as the queues start to fill up.

@nishadsingh1, you have a fix for this right?

nishadsingh1:
This PR includes some utilities that could be useful for future benchmarking.

Some of the changes here are less obviously extensible to other benchmarking use cases. I'm just going to comment on the them directly to give context for why they exist. I'm happy to scrap or rework the ones that are too case-specific/ won't be useful for future benchmarking efforts.

This PR currently does not include the throughput benchmarking script (https://github.com/nishadsingh1/clipper/blob/benchmark/src/benchmarks/src/noop_bench.cpp). Given that we have a benchmarking example in the source, `end_to_end_bench`, and documentation accompanying it at `bench/README.md`, I wonder if it would be good to include a clean version of the throughput benchmaring script too. It could demonstrate usage of some of the benchmarking utilities not used by `end_to_end_bench` which we could also refer to in the documentation. Alternatively, we could move everything under `bench/` and `src/benchmarks/src/` (including the throughput benchmarking script) to some other branch with updated documentation at `bench/README.md`.

Resolves #222.


nishadsingh1:
I'll explain one group of changes here since they function as a group. This relates to the changes in: `bench/setup_same_prediction_bench_docker.sh`, `bench/setup_noop_bench_docker.sh`, `SamePredictionBenchDockerfile`, `NoopBenchDockerfile`, and `/bin/build_bench_docker_images.sh`.

Our current benchmarking approach relies on Clipper to connect to model-containers that don't explicitly register with it. For this to happen, a model-container must try to connect to Clipper on the default port and with the IP address where Clipper exists. You can see an example of this in `end_to_end_bench` -- we just run `./bin/setup_sklearn_bench.sh`, which spins up a locally running instance of SklearnContainer, and Clipper connects to it automatically in the `end_to_end_bench` script. However, if that SklearnContainer instance was running on a Docker container, its IP would differ from Clipper's and so it wouldn't connect. To allow it to connect, we can set the CLIPPER_IP environment variable on the docker container to the IP in which clipper is actually running.

This is what the new `setup_noop_bench_docker.sh` does for the NoopContainer -- it grabs the local machine's IP address (it assumes it's on an AWS cluster) and sets the environment var CLIPPER_IP to it. NoopBenchDockerfile is just the Dockerfile wrapper to this script.

An identical dynamic exists between `setup_same_prediction_bench_docker.sh` and `SamePredictionBenchDockerfile` for a new model-container SamePredictionContainer.

This is a sort of hacky solution to a very specific problem, so I understand if we'd want to either change the solution itself or not include one in this PR.

bench/setup_sklearn_bench.sh
```diff
@@ -9,10 +9,10 @@ trap "kill 0" EXIT
...
 
-python bench_init.py $cifar_path
+python bench/bench_init.py $cifar_path
 
 export CLIPPER_MODEL_NAME="bench_sklearn_cifar"
 export CLIPPER_MODEL_VERSION="1"
 export CLIPPER_MODEL_PATH="model/"
 
-python ../containers/python/sklearn_cifar_container.py 
+python containers/python/sklearn_cifar_container.py
```
nishadsingh1: In `end_to_end_bench`, we prompt the user to run this script (`bench/setup_sklearn_bench.sh`) from the root directory. This change lets it work from the root directory (before, it would work only if run from within `bench/`).

containers/python/same_prediction_container.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import print_function
```
nishadsingh1: This container is even more noop-y than the noop container -- I am using it for throughput benchmarking to minimize the amount of time actually generating predictions, letting container latency better approximate rpc overhead. This isn't particularly hard to re-implement when needed; it might just be muddying the codebase to have this here. Happy to remove if you think that's best.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+const std::string POISSON_DELAY_PROMPT =
+    "Enter \"true\" if you want the delay between request batches to be drawn "
+    "from a "
+    "poisson distribution: ";
+const std::string MODEL_NAME_PROMPT =
+    "Enter the name of the model you want queried: ";
+const std::string MODEL_VERSION_PROMPT =
+    "Enter the version of the model you want queried: ";
+
+std::string _get_prompt(std::string var) {
```
nishadsingh1: There must be a much more elegant way to implement this. I noticed that switch statements don't work with strings, and creating a static map didn't seem a lot better. Thoughts?

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <sstream>
+#include <unordered_map>
+
+#include "include/bench_utils.hpp"
+
+using namespace clipper;
+
+namespace bench_utils {
+
+const std::string CIFAR_DATA_PATH_PROMPT =
```
nishadsingh1: Only a handful of these (and their counterpart variables here https://github.com/ucbrise/clipper/pull/225/files#diff-5ac4406c77f36f86be1d3ed9ab6795fbR8) are used by `end_to_end_bench`. The rest are used by the throughput benchmarking script that isn't included. Leaving these strings around might reduce busywork in the future, but it also could just add noise. Happy to remove if that seems right.

src/libclipper/include/clipper/app_metrics.hpp
```diff
@@ -0,0 +1,47 @@
+#ifndef CLIPPER_APP_METRICS_HPP
```
nishadsingh1: This has been moved to its own file because it can be used to gather metrics for benchmarking (as is done here https://github.com/nishadsingh1/clipper/blob/benchmark/src/benchmarks/src/noop_bench.cpp#L192).

src/libclipper/include/clipper/containers.hpp
```diff
@@ -31,13 +32,14 @@ class ModelContainer {
...
   double get_average_throughput_per_millisecond();
-  void update_throughput(size_t batch_size, long total_latency);
+  void update_container_stats(size_t batch_size, long total_latency);
   void send_feedback(PredictTask task);
 
   VersionedModelId model_;
   int container_id_;
   int replica_id_;
   InputType input_type_;
+  clipper::metrics::Histogram latency_hist_;
```
nishadsingh1: This histogram (which receives datapoints here: https://github.com/ucbrise/clipper/pull/225/files#diff-628067215a2c6ded52652fff0b43720bR46) is not used now, but I could see it being involved in some selection policies / task discarding in the future. Happy to remove if this isn't the right time to include it.

src/libclipper/include/clipper/metrics.hpp
```diff
@@ -291,6 +291,7 @@ class Histogram : public Metric {
   void insert(const int64_t value);
   const HistogramStats compute_stats();
   static double percentile(std::vector<int64_t> snapshot, double rank);
+  double percentile(double rank);
```
nishadsingh1: Just calls `percentile(snapshot, rank)` with the same snapshot used in `compute_stats()`. It's just included because it might be convenient -- here's an example of its usage: https://github.com/nishadsingh1/clipper/blob/benchmark/src/benchmarks/src/noop_bench.cpp#L130.

bench/bench_init.py
```diff
@@ -6,7 +6,6 @@
 sys.path.append(os.path.abspath("%s/../examples" % cur_dir))
 # sys.path.insert(0, os.path.abspath('%s/../containers/python/' % cur_dir))
```
Corey-Zumar: Nice - while we're cleaning up imports, can this commented import be removed as well?

bench/README.md
```diff
@@ -9,11 +9,13 @@ This benchmark serves an SKLearn model that depends on the CIFAR10 Python datase
...
+The C++ benchmark works by sending CIFAR10 query vectors to the container serving the trained SKLearn model. To achieve this, the **binary dataset** is required. It can also be obtained from [https://www.cs.toronto.edu/~kriz/cifar.html](https://www.cs.toronto.edu/~kriz/cifar.html). You can click [here](https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz) to download directly.
+
+You'll want to unzip the .tar.gz file manually.
 
 ## Optional Configuration Files
 The following benchmark attributes can be loaded via a JSON configuration file:
-- **cifar_data_path**: The path to a **specific binary data file** within the CIFAR10 binary dataset with a name of the form `data_batch_<n>.bin`. (`data_batch_1.bin`, for example)
+- **cifar_data_path**: The path to a **specific binary data file** within the CIFAR10 binary dataset with a name of the form `data_batch_<n>.bin`. (`/Users/.../cifar-100-binary/data_batch_1.bin`, for example)
```
Corey-Zumar: Let's make the path here more generic. Maybe `<path_to_unzipped_cifar_directory>/data_batch_1.bin`.

bench/setup_noop_bench_docker.sh
```diff
@@ -0,0 +1,17 @@
...
+
+trap "exit" INT TERM
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="bench_noop"
+export CLIPPER_MODEL_VERSION="1"
+export CLIPPER_MODEL_PATH="model/"
+
+# Sets CLIPPER_IP to AWS instance's IP
+export CLIPPER_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
Corey-Zumar: From the documentation, it seems like `169.254.169.254` is a local URI that, if queried within instance `X`, provides metadata about instance `X`. Therefore, this should only work if the docker container corresponding to `noop_container` is running on the same host as the `query_frontend` container, right? If so, we should make note of this limitation.

bench/setup_sklearn_bench.sh
Corey-Zumar: Nice!

bin/build_bench_docker_images.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
Corey-Zumar: Any objection to moving this from `/bin` to `/bench`?

containers/python/same_prediction_container.py
Corey-Zumar: Let's rename this to `noop_container` and remove the current version of `noop_container`. Afaik, we don't have any scripts that depend on the nature of the output of the current `noop_container` (if we do, we shouldn't), and this seems implementation to be a more accurate no-op.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+      new_label_vecs.push_back(cifar_double_vec);
+      vecs_map.emplace(label, new_label_vecs);
+    }
+  }
+  return vecs_map;
+}
+
+std::vector<std::vector<double>> concatenate_cifar_datapoints(
+    std::unordered_map<int, std::vector<std::vector<double>>> cifar_data) {
+  std::vector<std::vector<double>> planes_vecs = cifar_data.find(0)->second;
```
Corey-Zumar: Let's define constants `constexpr int CIFAR_PLANE_LABEL = 0`  and `constexpr int CIFAR_BIRD_LABEL = 2` and use them in the calls to `find`.

src/benchmarks/src/bench_utils.cpp
Corey-Zumar: Unfortunately, it doesn't seem like there is a more elegant way to do string matching in C++. However, I think that it would be beneficial to define prompts for a specific benchmark in its own implementation file. This would reduce clutter.

 Towards this end, we should wrap these utility methods in a class with a constructor that accepts a mapping of `desired_vars` to prompt variables. `_get_prompt` can then reference this map.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,174 @@
...
+  std::vector<std::vector<double>> birds_vecs = cifar_data.find(2)->second;
+
+  planes_vecs.insert(planes_vecs.end(), birds_vecs.begin(), birds_vecs.end());
+  return planes_vecs;
+}
+
+std::string get_str(const std::string &key,
+                    std::unordered_map<std::string, std::string> &config) {
+  return config.find(key)->second;
```
Corey-Zumar: I like these convenience methods. Let's check that the entry returned by `config.find(key)` exists before attempting to dereference the pointer. In the case that no such entry exists, we should throw our own readable error.

src/benchmarks/src/include/bench_utils.hpp
```diff
@@ -0,0 +1,71 @@
+#include <unordered_map>
+
+namespace bench_utils {
+
+/*
+ * Keys for config
+ */
+const std::string CIFAR_DATA_PATH = "cifar_data_path";
+const std::string NUM_THREADS = "num_threads";
+const std::string NUM_BATCHES = "num_batches";
```
Corey-Zumar: Again, I think it makes sense to define benchmark-specific keys within individual implementation files.

src/libclipper/include/clipper/app_metrics.hpp
Corey-Zumar: I think we should include this within `metrics.hpp` and the `metrics` namespace instead.

src/libclipper/include/clipper/containers.hpp
Corey-Zumar: This seems useful! Let's keep it. See previous comment about separating it from the `update_throughput` method.

src/libclipper/include/clipper/containers.hpp
```diff
@@ -31,13 +32,14 @@ class ModelContainer {
 
   size_t get_batch_size(Deadline deadline);
   double get_average_throughput_per_millisecond();
-  void update_throughput(size_t batch_size, long total_latency);
+  void update_container_stats(size_t batch_size, long total_latency);
```
Corey-Zumar: I'd rather we separate the critical task of updating container throughput for proper batching from the helpful but non-critical metrics update. Let's call `processing_container->latency_hist_.insert(total_latency)` after `processing_container->update_throughput` within the task executor.

src/libclipper/include/clipper/metrics.hpp
Corey-Zumar: Looks good! Can you add documentation indicating that a snapshot is obtained from the histogram's sampler and then used to calculate the percentile?

Corey-Zumar (changes requested):

bench/bench_init.py
nishadsingh1: Yep

bench/README.md
nishadsingh1: Agreed

bench/setup_noop_bench_docker.sh
nishadsingh1: Yeah, that's exactly right -- thanks for looking into it; I agree we should clarify that.

bin/build_bench_docker_images.sh
nishadsingh1: Sounds good to me

src/benchmarks/src/bench_utils.cpp
nishadsingh1: I agree that `same_prediction_container` is better suited for the name `noop_container` than the current `noop_container` is. I think it'll be useful to keep the old one because, in performing _some_ sort of operation on input data, it ensures that, at the very least, the input data isn't malformed and that the user's system works end to end. Maybe we call it `sum_container` because it sums datapoint entries? Thoughts, @dcrankshaw?

After talking with Dan, it seems like we don't want to support receiving parameters through a prompt anyway – only through JSON config files. So...we can just remove all the corresponding ugly code.

src/benchmarks/src/bench_utils.cpp
nishadsingh1: Sounds good

src/benchmarks/src/include/bench_utils.hpp
nishadsingh1: Yeah I agree. Especially if they don't share prompt text (because the prompt support won't exist)

src/libclipper/include/clipper/containers.hpp
nishadsingh1: After talking to dan, the plan is to revert the extraction of AppMetrics and make a separate BenchMetrics (or something) in bench_util. For now, the two measure seemingly identical things, but that can easily change with what we choose to (or not to) benchmark (say, prediction accuracy, which isn't captured in AppMetrics). 

The change to `update_throughput` sounds good to me.

nishadsingh1:
This new update removes the old `noop_bench` and `end_to_end_bench` and replaces them with `generic_bench`, which contains the intersection of their capabilities. Check out the new README for details. The intention is that this is clean/well documented enough to get merged into `develop`. Other changes we want to keep that don't fit that criteria will get added to a separate benchmarking branch.

Two questions before I request a review:
1. Do we want to keep the the scripts/dockerfiles for putting benchmarking-specific model-containers in docker containers? The reason for their existence is described in an above comment, and the corresponding files are: bench/setup_same_prediction_bench_docker.sh, bench/setup_noop_bench_docker.sh, SamePredictionBenchDockerfile, NoopBenchDockerfile, and /bin/build_bench_docker_images.sh.
2. (As per Corey's comment https://github.com/ucbrise/clipper/pull/225#discussion_r123912396) do we want to rename `noop_container` to `sum_container` (since it sums inputs) and `same_prediction_container` to `noop_container`?

I'll make the corresponding changes and request a review once they're up.

dcrankshaw:
1. Yes. Keep the Dockerfile's for benchmarking containers. They make benchmarking much more convenient.
2. Yes, rename the containers. Can you grep for `noop_container` in the codebase though, because we are changing the semantics of that container? I don't think we have any tests that rely on `noop_container` actually summing the contents, but it would be good to double check.

nishadsingh1:
Alright, it's ready for another review. Thanks a bunch to both of you for giving it a thorough pass once already. Pretty sure it's going to yell at me for formatting, but my local formatting script says all is fine -- would one of you mind pushing a formatting update?

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+      // Select datapoint
+      data_index = indices[(j * request_batch_size + i) % num_datapoints];
+      query_vec = data[data_index];
+      label = labels[data_index];
+
+      if (prevent_cache_hits) {
+        // Modify it to be epoch and thread-specific
+        query_vec[0] += (j * request_batch_size + i) / num_datapoints;
+        query_vec[1] += thread_id;
+      }
```
nishadsingh1: I'm having some trouble here. Empirically, this actually does seem to prevent cache hits, but because all of the input data points are shared (through the variable `data`), it seems like there could be a data race issue. We could make a copy of `data` for each thread, but that feels wasteful. We could put a lock on all of the data or even on each data point, but that feels gratuitous. Modifying the data after the memcpy in `std::make_shared` (by retrieving input->get_data() and and modifying that vector directly) doesn't seem to actually change the submitted `input` variable. Thoughts on how to do this correctly?

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+  for (int j = 0; j < num_batches; j++) {
+    std::vector<boost::future<Response>> futures;
+    std::vector<double> future_labels;
+    for (int i = 0; i < request_batch_size; i++) {
+      // Select datapoint
+      data_index = indices[(j * request_batch_size + i) % num_datapoints];
+      query_vec = data[data_index];
+      label = labels[data_index];
+
+      if (prevent_cache_hits) {
```
nishadsingh1: I'm having some trouble here. In order to prevent us from hitting Clipper's internal cache, we need every submitted datapoint to be unique. Remember that we can cycle over the input data multiple times. The solution here attempts to modify each datapoint by adding info about the thread and epoch it belongs to.

Empirically, this approach actually does seem to prevent cache hits, but there are still at least two problems.

The first is that this solution, which replaces the first and second entry of each datapoint, assumes that there are no two distinct CIFAR datapoints that only differ in their first and/or second entries. If there are, this change won't actually prevent all cache hits. I think this is a pretty minor problem -- there actually aren't any two CIFAR datapoints I've encountered that fulfill this property.

The second: because all of the input data points are shared (through the variable `data`), it seems like there could be a data race issue. We could make a copy of `data` for each thread, but that feels wasteful. We could make a copy of each datapoint within the inner for loop before modifying it, but that feels expensive. We could put a lock on all of the data or even on each data point, but that feels gratuitous. We could just make every vector two entries longer and modify those instead, but then the datapoints wouldn't fit into models that are trained on the original datapoints which would be two entries shorter. Modifying the data after the memcpy in std::make_shared (by retrieving input->get_data() and and modifying that vector directly) doesn't seem to actually change the submitted input variable. Thoughts on how to do this correctly?

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
-- **num_batches**: The number of batches of requests to be sent by each thread
-- **batch_size**: The number of requests to be sent in each batch
-- **batch_delay_millis**: The per-thread delay between batches, in milliseconds
+## Spin up a model container to query
+
+The benchmarking tool will send requests through Clipper to a model. You'll want to spin up this model in a container that implements Clipper's Model-Container RPC interface.
+
+You can also use one of our predefined model-container scripts:
+
+- If you want to have the model serving off your machine
```
Corey-Zumar: "If you want to serve the model directly from your machine" flows better

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+
+  - [`bench/setup_noop_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_noop_bench.sh) runs the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py)
+  - [`bench/setup_sum_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sum_bench.sh) runs the [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py)
+  - [`bench/setup_sklearn_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sklearn_bench.sh) runs the [`sklearn_cifar_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sklearn_cifar_container). If you wish to use this option, remember to download the CIFAR10 python dataset and run
+  ```sh
+  ./setup_bench.sh <path_to_cifar_python_dataset>
+  ```
+where `<path_to_cifar_python_dataset>` is the path to the **directory** containing a parsed CIFAR10 CSV data file with name `cifar_train.data`.
+
+- If you want the model serving off a Docker container, you'll need to build its image. We have some Dockerfiles that define images for some model-containers.
```
Corey-Zumar: "If you want to serve the model from within a docker container, you'll need to build the container's image. We have some Dockerfiles that define images for some model containers."

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+## Run the benchmarking tool
+**These steps are given relative to the main clipper source directory.**
 
-1. Execute the following:
-  ```sh
-  ./setup_bench.sh <path_to_cifar_python_dataset>
-  ```
-where `<path_to_cifar_python_dataset>` is the path to the **directory** containing a parsed CIFAR10 CSV data file with name `cifar_train.data`.
+1. Build the Clipper source for release: 
+`./configure --release`, `cd release && make`
```
Corey-Zumar: Replace comma with `&&`. This can all be one instruction. We should only have to make `generic_bench`, right?

bench/setup_noop_bench.sh
```diff
@@ -0,0 +1,14 @@
...
+set -o pipefail
+
+trap "exit" INT TERM
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="bench_noop"
+export CLIPPER_MODEL_VERSION="1"
+export CLIPPER_MODEL_PATH="model/"
+
+python containers/python/noop_container.py
```
Corey-Zumar: This should be runnable from any directory (currently only works from clipper root). Reference the script's directory and execute `python $SCRIPTDIR/../containers/python/noop_container.py`.

For reference, `build_docker_images.sh` is a good example.

bench/setup_sum_bench.sh
```diff
@@ -0,0 +1,14 @@
...
+set -o pipefail
+
+trap "exit" INT TERM
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="bench_sum"
+export CLIPPER_MODEL_VERSION="1"
+export CLIPPER_MODEL_PATH="model/"
+
+python containers/python/sum_container.py
```
Corey-Zumar: Make this runnable from any directory (see previous comment)

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **num_batches**: The number of batches of requests to be sent by each thread.
+- **request\_batch_size**: The number of requests to be sent in each batch.
+- **request\_batch\_delay_micros**: The per-thread delay between batches, in microseconds. *request\_batch\_delay_micros* and *request\_batch_size* together determine the burstiness of your supplied workload.
+- **poisson_delay**: `"true"` if you wish for the delays between request batches to be drawn from a poisson distribution with mean. `"false"` if you wish for the delay between request batches to be uniform.
+- **prevent\_cache_hits**: `"true"` if you wish for the script to modify datapoints (possibly at the expense of prediction accuracy) in order to prevent hitting Clipper's internal prediction cache. `"false"` otherwise.
+- **latency_objective**: The latency objective for the app that will be created, in microseconds
+- **reports_path**: Path to the file in which you want your benchmarking reports saved
+- **report\_delay_seconds**: The delay between each flush of benchmarking metrics to your reports file, in seconds. At each flush, the metrics will reset.
```
Corey-Zumar: There should be an option to never flush (-1 ?).

NoopBenchDockerfile
```diff
@@ -0,0 +1,11 @@
+FROM clipper/py-rpc:latest
+
+MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
```
dcrankshaw: You can remove the MAINTAINER line. It's apparently deprecated (https://docs.docker.com/engine/reference/builder/#maintainer-deprecated).

NoopBenchDockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+COPY containers/python/rpc.py /containers/python/
```
dcrankshaw: You don't need to copy rpc.py. It's already included by the base image.

SumBenchDockerfile
```diff
@@ -0,0 +1,11 @@
+FROM clipper/py-rpc:latest
+
+MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
+
+COPY containers/python/rpc.py /containers/python/
```
dcrankshaw: You don't need to copy rpc.py

SumBenchDockerfile
```diff
@@ -0,0 +1,11 @@
+FROM clipper/py-rpc:latest
+
+MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
```
dcrankshaw: Delete maintainer line

SumDockerfile
```diff
@@ -0,0 +1,9 @@
+FROM clipper/py-rpc:latest
+
+MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
+
+COPY containers/python/sum_container.py /container/
+
+CMD ["python", "/container/sum_container.py"]
+
```
dcrankshaw, line 6: What's the difference between the `SumBenchDockerfile` and the `SumDockerfile`?

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+### Download CIFAR10 Binary Dataset for Query Execution
+The C++ benchmark works by sending CIFAR10 query vectors to the container serving the trained SKLearn model. To achieve this, the **binary dataset** is required. It can also be obtained from [https://www.cs.toronto.edu/~kriz/cifar.html](https://www.cs.toronto.edu/~kriz/cifar.html). You can click [here](https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz) to download directly.
+
+You'll want to unzip the .tar.gz file manually.
+
+
+### (Optional) Download CIFAR10 Python Dataset for SKLearn Model 
+You can skip this section if you do not intend to use our SKLearn Model for benchmarking.
+
+Our SKLearn model depends on the CIFAR10 Python dataset for training. In order to train the model with this dataset, it must be converted to CSV format. The [CIFAR10 python download utility](https://github.com/ucbrise/clipper/blob/develop/examples/tutorial/download_cifar.py) can be used to obtain the required CSV-formatted dataset.
```
dcrankshaw: You can write this link as `[CIFAR10 python download utility](../examples/tutorial/download_cifar.py)`

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
-The C++ benchmark works by sending CIFAR10 query vectors to the container serving the trained SKLearn model. To achieve this, the **binary dataset** is required. It can also be obtained from [https://www.cs.toronto.edu/~kriz/cifar.html](https://www.cs.toronto.edu/~kriz/cifar.html).
 
-## Optional Configuration Files
-The following benchmark attributes can be loaded via a JSON configuration file:
-- **cifar_data_path**: The path to a **specific binary data file** within the CIFAR10 binary dataset with a name of the form `data_batch_<n>.bin`. (`data_batch_1.bin`, for example)
-- **num_threads**: The number of threads of execution
-- **num_batches**: The number of batches of requests to be sent by each thread
-- **batch_size**: The number of requests to be sent in each batch
-- **batch_delay_millis**: The per-thread delay between batches, in milliseconds
+## Spin up a model container to query
```
dcrankshaw: Replace "Spin up" with "Deploy". This is more consistent with the rest of Clipper documentation nomenclature.

bench/README.md
dcrankshaw: This is a little confusing. Do you mean if you want to run the model locally outside a Docker container?

bench/README.md
dcrankshaw: Fix the wording to: "If you want to deploy the model in a Docker container, you'll need to build the model container Docker image."

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+   "request_batch_size":"<request_batch_size>",
+   "request_batch_delay_micros":"<request_batch_delay_micros>",
+   "poisson_delay":"<true/false>",
+   "prevent_cache_hits":"<true/false>",
+   "latency_objective":"<latency_objective>",
+   "reports_path":"<reports_path>",
+   "reports_delay_seconds":"<reports_delay_seconds>",
+   "model_name":"<model_name>",
+   "model_version":"<model_version>"
 }
```
dcrankshaw: Can you include a `config.json.template` file with exactly this content in `bench/`?

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+  - Create the Docker images for the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py)  and [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py) by running `./bin/build_bench_docker_images.sh`.
+  - Run `docker run <image_id>` on the same host that the benchmarking tool will be run from.
+
 
-To configure these attributes, create a JSON file with the following format and specify its path when the benchmark is executed (see **Steps of Execution** below).
+## Define your benchmarking parameters
+
+Create a JSON configuration file that specifies values for the following parameters:
+
+- **cifar\_data_path**: The path to a *specific binary data file* within the CIFAR10 binary dataset with a name of the form `data_batch_<n>.bin`:  `<path_to_unzipped_cifar_directory>/data_batch_1.bin`, for example.
```
dcrankshaw: Reword the end of this bullet to "... For example, <path...>" (Move the "For example" in front of the example)

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **num_threads**: The number of threads of execution. This should be a positive integer. Each of these threads will access the data in an independent, shuffled order, and will send requests according to *num_batches*, *request\_batch_size*, *request\_batch\_delay_micros*, *poisson_delay*, and *prevent\_cache_hits*, all described below.
```
dcrankshaw: "The number of threads used to send benchmark requests." Your explanation is good, but this will clarify that this parameter shouldn't affect Clipper performance, just the benchmark execution code.

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **num_batches**: The number of batches of requests to be sent by each thread.
```
dcrankshaw: "The total number of request batches to be sent by each thread."

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **request\_batch_size**: The number of requests to be sent in each batch.
+- **request\_batch\_delay_micros**: The per-thread delay between batches, in microseconds. *request\_batch\_delay_micros* and *request\_batch_size* together determine the burstiness of your supplied workload.
+- **poisson_delay**: `"true"` if you wish for the delays between request batches to be drawn from a poisson distribution with mean. `"false"` if you wish for the delay between request batches to be uniform.
```
dcrankshaw: With mean what? Do you use the `request_batch_delay_micros` parameter to set the mean?

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **prevent\_cache_hits**: `"true"` if you wish for the script to modify datapoints (possibly at the expense of prediction accuracy) in order to prevent hitting Clipper's internal prediction cache. `"false"` otherwise.
```
dcrankshaw: Do we measure prediction accuracy in this benchmark? We probably shouldn't. Assuming we don't measure accuracy, delete the comment in parentheses.

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+- **latency_objective**: The latency objective for the app that will be created, in microseconds
+- **reports_path**: Path to the file in which you want your benchmarking reports saved
```
dcrankshaw: Rename to **benchmark_report_path**.

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+   "model_version":"<model_version>"
 }
 ```
 
 If a configuration file is not specified, the benchmark will prompt you for the values of these attributes at runtime.
 
-## Steps of Execution
-**These steps are given relative to the current directory.**
+## Run the benchmarking tool
+**These steps are given relative to the main clipper source directory.**
```
dcrankshaw: s/steps/instructions

bench/README.md
dcrankshaw: The comma is confusing. Change this to `./configure --release && cd release && make generic_bench`

Also, specifying the `make` target speed up compile time and avoid compiling unnecessary binaries.

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
-  ```sh
-  cd .. && ./configure --release && cd release
-  make end_to_end_bench
-  ```
-  
-3. If you created a JSON configuration file above, execute the following:
-  ```sh
-  ./src/benchmarks/end_to_end_bench -f "<path_to_config_.json>"
-  ```
+2. Confirm you have a Redis server running. If you do not, run `redis-server`
```
dcrankshaw: Change the text to:

Confirm you have a Redis server running. If you do not, run `redis-server` or `docker run -d -p 6379:6379 redis:alpine`

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+
+3. Confirm your model-container is running. If it is not, follow the instructions in the "Spin up a model container for querying" section.
```
dcrankshaw: Hyperlink back to the previous section in the document ([instructions on intra-document links](https://gist.github.com/asabaylus/3071099))

dcrankshaw: Also, don't forget you're changing the name of the section referenced here to "Deploy a model container for querying"

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+
+4. Run the tool with your JSON configuration file: `./release/src/benchmarks/generic_bench -f <path_to_your_config_file>`
```
dcrankshaw: s/Run the tool with your JSON configuration file/Run the benchmark

bench/build_bench_docker_images.sh
```diff
@@ -0,0 +1,20 @@
...
+# the script.
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Let the user start this script from anywhere in the filesystem.
+cd $DIR/..
+
+# Build the Clipper Docker images
+# Assume local (if exists) clipper/py-rpc base image or pulled image is correct
+time docker build -t clipper/same-prediction-bench -f SamePredictionBenchDockerfile ./
+time docker build  -t clipper/noop-bench -f NoopBenchDockerfile ./
```
dcrankshaw: Do you need to add the command to build the `SumDockerfile` here too?

Also, SamePredictionBench doesn't exist anymore right? We changed the name?

bench/setup_noop_bench_docker.sh
```diff
@@ -0,0 +1,19 @@
...
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="bench_noop"
+export CLIPPER_MODEL_VERSION="1"
+export CLIPPER_MODEL_PATH="model/"
+
+# Sets CLIPPER_IP to AWS instance's IP.
+# This will only work if the docker container corresponding to noop_container
+# is running on the same host as the clipper/query_frontend container.
+export CLIPPER_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
dcrankshaw: There's no mention of AWS in the README. Should this still be in here?

bench/setup_sum_bench_docker.sh
```diff
@@ -0,0 +1,19 @@
...
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="sum_noop"
+export CLIPPER_MODEL_VERSION="1"
+export CLIPPER_MODEL_PATH="model/"
+
+# Sets CLIPPER_IP to AWS instance's IP.
+# This will only work if the docker container corresponding to sum_container
+# is running on the same host as the clipper/query_frontend container.
+export CLIPPER_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
dcrankshaw: Same comment as above about AWS

containers/python/sum_container.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class SumContainer(rpc.ModelContainerBase):
+    def __init__(self):
+        pass
+
+    def predict_ints(self, inputs):
+        outputs = []
+        for input_item in inputs:
+            outputs.append(str(sum(input_item)))
```
dcrankshaw: We can make this a little more performant (this should get rid of the appends). Do the same for all the `predict_*` functions.

```py
def predict_ints(self, inputs):
    return [str(sum(item)) for item in inputs]
```

src/benchmarks/CMakeLists.txt
```diff
@@ -3,9 +3,9 @@ cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 ##################################
 # Benchmark executable
 
-add_executable(end_to_end_bench src/end_to_end_bench.cpp)
+add_executable(generic_bench src/generic_bench.cpp src/bench_utils.cpp src/include/bench_utils.hpp)
```
dcrankshaw: I don't think there's any need to rename this. It's still doing an end-to-end benchmark. And generic benchmark isn't a very informative name.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+      "app:" + app_name + ":send_rate");
+  num_predictions_ =
+      clipper::metrics::MetricsRegistry::get_metrics().create_counter(
+          "app:" + app_name + ":num_predictions");
+  accuracy_ratio_ =
+      clipper::metrics::MetricsRegistry::get_metrics().create_ratio_counter(
+          "app:" + app_name + ":accuracy_ratio");
+  default_pred_ratio_ =
+      clipper::metrics::MetricsRegistry::get_metrics().create_ratio_counter(
+          "app:" + app_name + ":default_prediction_ratio");
```
dcrankshaw: Let's prefix each of these metrics with "bench" instead of "app"

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
   
-  Otherwise, execute
-  ```sh
-  ./src/benchmarks/end_to_end_bench
-  ```
-  and specify the values of the attributes enumerated in the **Optional Configuration Files** section above. 
+5. Check the logs/output from your model container to confirm it's being queried
+
+6. View the benchmarking reports at your specified **reports_path**.
```
dcrankshaw: Can you add a section describing how to interpret a report? You don't need to explain every metric, but at least explain the bench-specific metrics and some of the model metrics. As a rule of thumb, if it's a metric you've been looking at when doing benchmarking the last couple weeks you should add an explanation.

These explanations don't have to be super detailed, just a sentence or two.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+    log_error(
+        "BENCH",
+        "Please specify the full path of the binary CIFAR-100 data file.");
+    std::terminate();
+  }
+
+  std::ifstream cifar_file(cifar_data_path, std::ios::binary);
+  std::istreambuf_iterator<char> cifar_data(cifar_file);
+  std::unordered_map<int, std::vector<std::vector<double>>> vecs_map;
+  for (int i = 0; i < 10000; i++) {
```
dcrankshaw: Where does 10000 come from?

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+    int label = static_cast<int>(*cifar_data);
+    cifar_data++;
+    std::vector<uint8_t> cifar_byte_vec;
+    cifar_byte_vec.reserve(3072);
```
dcrankshaw: Make a const called NUM_CIFAR_FEATURES = 3072 and use that instead of the magic number.

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+#include <cxxopts.hpp>
+
+#include <clipper/constants.hpp>
+#include <clipper/datatypes.hpp>
+#include <clipper/future.hpp>
+#include <clipper/json_util.hpp>
+#include <clipper/logging.hpp>
+#include <clipper/query_processor.hpp>
+#include <fstream>
+#include <random>
```
dcrankshaw: Move the fstream and random imports up with the other standard library imports (iostream, vector, etc)

src/benchmarks/src/generic_bench.cpp
dcrankshaw: TL;DR: Make a copy of the dataset for each thread to avoid the data race and continue to modify the query_vec as you're doing right now (by modifying the first and second elements).

**Responses to some of your comments:**
So my first comment is don't worry too much about optimizing the benchmark code for anything besides speed. That means extra copies of data aren't a big deal unless you are making the copy in an inner loop where the speed of the memcpy itself is an issue, and we should definitely avoid any locks.

There is definitely a data race here, it seems like the easiest thing to do is just make a copy of the dataset for each thread.

Your assumption about there being no CIFAR data points that differ only in the first and second dimension seems fine. Remember, if there are one or two cache hits it really won't affect anything.

You can't modify the inputs after creating them because `get_data()` returns a `const vector` so you can't modify it.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+constexpr double SKLEARN_PLANE_LABEL = 1;
+constexpr double SKLEARN_BIRD_LABEL = 0;
+
+BenchMetrics::BenchMetrics(std::string app_name) {
+  latency_ = clipper::metrics::MetricsRegistry::get_metrics().create_histogram(
+      "app:" + app_name + ":prediction_latency", "microseconds", 4096);
+  throughput_ = clipper::metrics::MetricsRegistry::get_metrics().create_meter(
+      "app:" + app_name + ":prediction_throughput");
+  send_rate_ = clipper::metrics::MetricsRegistry::get_metrics().create_meter(
+      "app:" + app_name + ":send_rate");
```
dcrankshaw: Rename "send_rate" to "request_throughput"

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+      futures.push_back(std::move(prediction));
+      future_labels.push_back(std::move(label));
+      bench_metrics.send_rate_->mark(1);
+    }
+
+    std::shared_ptr<std::atomic_int> completed =
+        std::make_shared<std::atomic_int>(0);
+    std::pair<boost::future<void>, std::vector<boost::future<Response>>>
+        results = future::when_all(std::move(futures), completed);
```
dcrankshaw: Why are you waiting until all the inputs in a batch have completed? Instead, add a continuation (`.then()`) to each future and update the metrics as each response comes in. That will results in more accurate timings, particularly if we want to see how batch size affects real throughput rate.

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+      }
+      bench_metrics.latency_->insert(r.duration_micros_);
+      bench_metrics.num_predictions_->increment(1);
+      bench_metrics.throughput_->mark(1);
+
+      if (std::stod(r.output_.y_hat_) == future_labels[i]) {
+        bench_metrics.accuracy_ratio_->increment(1, 1);
+      } else {
+        bench_metrics.accuracy_ratio_->increment(0, 1);
+      }
```
dcrankshaw: Let's not measure accuracy in this benchmark. It's sort of irrelevant as this is a physical performance benchmark and we're intentionally modifying the feature vectors.

src/benchmarks/src/bench_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+  num_predictions_ =
+      clipper::metrics::MetricsRegistry::get_metrics().create_counter(
+          "app:" + app_name + ":num_predictions");
+  accuracy_ratio_ =
+      clipper::metrics::MetricsRegistry::get_metrics().create_ratio_counter(
+          "app:" + app_name + ":accuracy_ratio");
```
dcrankshaw: Delete accuracy ratio

src/benchmarks/src/generic_bench.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+  // Assign shuffled datapoint selection indices for each thread
+  for (int j = 0; j < num_threads; j++) {
+    std::vector<int> indices(num_datapoints);
+    for (int i = 0; i < static_cast<int>(num_datapoints); i++) {
+      indices[i] = i;
+    }
+    shuffle(indices.begin(), indices.end(), std::default_random_engine(seed));
+    indices_for_threads.push_back(std::move(indices));
+  }
```
dcrankshaw: What's the purpose of this for loop? What are you shuffling?

src/frontends/src/query_frontend.hpp
```diff
@@ -97,10 +97,13 @@ class AppMetrics {
...
                                       ":default_prediction_ratio")) {}
   ~AppMetrics() = default;
+
   AppMetrics(const AppMetrics&) = default;
+
   AppMetrics& operator=(const AppMetrics&) = default;
 
   AppMetrics(AppMetrics&&) = default;
+
   AppMetrics& operator=(AppMetrics&&) = default;
```
dcrankshaw: Delete these whitespace changes

src/libclipper/include/clipper/metrics.hpp
```diff
@@ -291,6 +291,9 @@ class Histogram : public Metric {
   void insert(const int64_t value);
   const HistogramStats compute_stats();
   static double percentile(std::vector<int64_t> snapshot, double rank);
+  // This method obtains a snapshot from the histogram's sampler and then
```
dcrankshaw: s/sample/reservoir sampler

@Corey-Zumar We're using reservoir sampling right?

src/libclipper/src/containers.cpp
```diff
@@ -26,6 +26,8 @@ ModelContainer::ModelContainer(VersionedModelId model, int container_id,
       container_id_(container_id),
       replica_id_(replica_id),
       input_type_(input_type),
+      latency_hist_("container_prediction_latency", "microseconds",
```
dcrankshaw: Change the name to `"container:" + model.serialiaze() + ":" std::to_string(replica_id) + ":prediction_latency"`

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+  - [`bench/setup_sklearn_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sklearn_bench.sh) runs the [`sklearn_cifar_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sklearn_cifar_container). If you wish to use this option, remember to download the CIFAR10 python dataset and run
+  ```sh
+  ./setup_bench.sh <path_to_cifar_python_dataset>
+  ```
+where `<path_to_cifar_python_dataset>` is the path to the **directory** containing a parsed CIFAR10 CSV data file with name `cifar_train.data`.
+
+- If you want the model serving off a Docker container, you'll need to build its image. We have some Dockerfiles that define images for some model-containers.
+  - Create the Docker images for the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py)  and [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py) by running `./bin/build_bench_docker_images.sh`.
```
dcrankshaw: Should be `./bench/build_bench_docker_images.sh`

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
    "num_batches":"<num_batches>",
-   "batch_size":"<batch_size>",
-   "batch_delay_millis":"<batch_delay_millis>"
+   "request_batch_size":"<request_batch_size>",
+   "request_batch_delay_micros":"<request_batch_delay_micros>",
+   "poisson_delay":"<true/false>",
+   "prevent_cache_hits":"<true/false>",
+   "latency_objective":"<latency_objective>",
+   "reports_path":"<reports_path>",
+   "reports_delay_seconds":"<reports_delay_seconds>",
```
dcrankshaw: This should be `report_delay_seconds`

bench/setup_sum_bench.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -e
+set -u
+set -o pipefail
+
+trap "exit" INT TERM
+trap "kill 0" EXIT
+
+export CLIPPER_MODEL_NAME="bench_sum"
+export CLIPPER_MODEL_VERSION="1"
```
dcrankshaw: Have this take model name and version as arguments.

bench/README.md
```diff
@@ -1,60 +1,104 @@
...
+  - [`bench/setup_sklearn_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sklearn_bench.sh) runs the [`sklearn_cifar_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sklearn_cifar_container). If you wish to use this option, remember to download the CIFAR10 python dataset and run
+  ```sh
+  ./setup_bench.sh <path_to_cifar_python_dataset>
+  ```
+where `<path_to_cifar_python_dataset>` is the path to the **directory** containing a parsed CIFAR10 CSV data file with name `cifar_train.data`.
+
+- If you want the model serving off a Docker container, you'll need to build its image. We have some Dockerfiles that define images for some model-containers.
+  - Create the Docker images for the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py)  and [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py) by running `./bin/build_bench_docker_images.sh`.
+  - Run `docker run <image_id>` on the same host that the benchmarking tool will be run from.
```
dcrankshaw: Make a note that what the models will be named for each model, so the user knows what name and version to put in the benchmark JSON config.

dcrankshaw (changes requested):
Very nice. This works great. My comments are lots of little changes and cleanup, nothing major.

Corey-Zumar (dismissed):
I think Dan covered almost everything. The only major issue I encountered is that `generic_bench` fails to log metrics if the report delay exceeds the length of the benchmark. I'm of the opinion that we should always report metrics at least once.

bench/README.md
Corey-Zumar: I.e only flush when the benchmark terminates

bench/setup_noop_bench_docker.sh
Corey-Zumar: This docker script is AWS-specific, and we should make note of that in the README as well as provide an alternative option to build a container with a custom IP.

bench/README.md
dcrankshaw: Agreed.

nishadsingh1:
@Corey-Zumar @dcrankshaw Thanks a bunch for both of your reviews :)

src/benchmarks/src/generic_bench.cpp
nishadsingh1: I agree that measuring accuracy here is besides the point and complicated by us changing feature vectors. That said, the old `end_to_end_bench` measured accuracy. Should we still include a way to run accuracy checks? I don't see a great reason to.

src/benchmarks/src/generic_bench.cpp
nishadsingh1: `indices[i]` contains an array of integers (running from `0` to `num_datapoints - `) shuffled in a random order. They specify the order in which thread `i` cycles through the data set to send queries.

I thought that each thread should query CIFAR datapoints in a different order (in case the model being tested ran, for some reason, more or less efficiently on a specific datapoint given the ones it processed recently). I'm realizing now that that's not only a strange assumption, but that we also make unique per thread so maybe it's totally irrelevant.

So, we could:
1. Not shuffle at all
2. Since we'll now make copies of the data set for each thread, shuffle the datapoints themselves instead of indices before passing them to each thread.

Do you have a preference? I'm leaning towards the former since I'm now having a hard time thinking of reasons in favor of shuffling.

nishadsingh1:
I think I addressed all your comments. Some highlights:
- removed accuracy reporting / creation + consideration of labels
- no longer shuffle datapoint access order for threads
- each thread gets its own copy of the data
- setting `reports_delay_seconds` to `-1` makes the script report only once the threads making requests have all completed. there's now a guarantee that at least one report is flushed.
- all the setup scripts have been updated to accept as input model names, versions, and IPs to connect to clipper with. the docker images allow for passing these values in as environment variables upon running them. Scripts are more transparent about which values they are using (and when they're resorting to default values).
- changed and updated the documentation, added a section on how to interpret reports

SumBenchDockerfile
```diff
@@ -0,0 +1,14 @@
+FROM clipper/py-rpc:latest
+
+COPY containers/python/sum_container.py /containers/python/
+COPY bench/setup_sum_bench_docker.sh /bench/
+COPY bench/set_bench_env_vars.sh /bench/
+COPY containers/python/rpc.py /containers/python/
```
nishadsingh1: Without copying rpc.py, I get an ImportError from the import statement in `sum_container.`py. Same thing happens for NoopBenchDockerfile with `noop_container.py`. Was having a hard time debugging it -- I'll keep looking into it and push a fix. Figured I'd this push up set of changes and not let this one problem block reviews.

dcrankshaw: Ahh I see the problem. RPCDockerfile copies the `rpc.py` to `/containers`, not `/containers/python` (https://github.com/ucbrise/clipper/blob/develop/RPCDockerfile#L13)

If you fix the import in `sum_container.py` you shouldn't need to copy `rpc.py`

SumDockerfile
dcrankshaw: Ahh I see. We have regular noop and sum containers that can be used as normal containsers, and then we add additional functionality when using them in these benchmarks and call those containers `SumBench` and `NoopBench`

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
 
-To configure these attributes, create a JSON file with the following format and specify its path when the benchmark is executed (see **Steps of Execution** below).
+- `model_version` and `model_name` should match values defined in your JSON configuration file (discussed in the next section)
+- If your model-container will be running on a different host than the benchmarking script, `clipper_ip` should be set to the IP address of the benchmarking script's host. Otherwise, it can be set to`"localhost"`.
+
+- If you want to run the model outside of a Docker container (`model_version` is `1` and `clipper_ip` is `localhost` by default):
+
+  - [`bench/setup_noop_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_noop_bench.sh) runs the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py). The default `model_name` is `"bench_noop"`.
+  
+      ./bench/setup_noop_bench.sh [[<model_name> <model_version> [<clipper_ip>]]]
```
dcrankshaw: Enclose these in triple backticks so they are formatted as code

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
+
+- If you want to run the model outside of a Docker container (`model_version` is `1` and `clipper_ip` is `localhost` by default):
+
+  - [`bench/setup_noop_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_noop_bench.sh) runs the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py). The default `model_name` is `"bench_noop"`.
+  
+      ./bench/setup_noop_bench.sh [[<model_name> <model_version> [<clipper_ip>]]] 
+    
+  - [`bench/setup_sum_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sum_bench.sh) runs the [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py). The default `model_name` is `"bench_sum"`.
+
+       ./bench/setup_sum_bench.sh [[<model_name> <model_version> [<clipper_ip>]]]
```
dcrankshaw: Format as code

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
+      ./bench/setup_noop_bench.sh [[<model_name> <model_version> [<clipper_ip>]]] 
+    
+  - [`bench/setup_sum_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sum_bench.sh) runs the [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py). The default `model_name` is `"bench_sum"`.
+
+       ./bench/setup_sum_bench.sh [[<model_name> <model_version> [<clipper_ip>]]] 
+  
+  - [`bench/setup_sklearn_bench.sh`](https://github.com/ucbrise/clipper/tree/develop/bench/setup_sklearn_bench.sh) runs the [`sklearn_cifar_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sklearn_cifar_container). If you wish to use this option, remember to download the CIFAR10 python dataset first. The default `model_name` is `"bench_sklearn_cifar"`.
+
+    
+      ./bench/setup_sklearn_bench.sh <path_to_cifar_python_dataset> [[<model_name> <model_version> [<clipper_ip>]]]
```
dcrankshaw: Format as code

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
+
+    Note that `<path_to_cifar_python_dataset>` should be the path to the **directory** containing a parsed CIFAR10 CSV data file with name `cifar_train.data`.
+
+- If you want to deploy the model in a Docker container, you'll need to build the model container Docker image (`model_version` is `1` by default):
+  - Create the Docker images for the [`noop_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/noop_container.py)  and [`sum_container`](https://github.com/ucbrise/clipper/blob/develop/containers/python/sum_container.py). The default names for each model are the same as the ones listed above.
+
+       ./bench/build_bench_docker_images.sh
```
dcrankshaw: Format as code

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
+  
+  - Run the container on the same host that the benchmarking tool will be run from:
+  
+        docker run [-e MODEL_NAME=<nondefault_model_name>] [-e MODEL_VERSION=<nondefault_model_version>] [-e IP=<non-aws-clipper-ip>] <image_id>
```
dcrankshaw: Format as code

bench/README.md
```diff
@@ -1,60 +1,175 @@
...
+Under `meters`:
+
+- `model:<model_name>:<model_version>:prediction_throughput`: The rate at which your model-container is serving requests
+- `bench:cifar_bench:request_throughput`: The rate at which the benchmarking tool is sending requests.
+- `bench:cifar_bench:prediction_throughput`: The rate at which the whole Clipper system is serving requests.
+
+Under `histograms` (each of these entries lists the unit of measurement, number of datapoints, min, max, mean, standard deviation, and p50, p95, and p99 values):
+
+- `bench:cifar_bench:prediction_latency`: Statistics on latency measured from when the benchmarking script sends a request to when it receives a response.
+- `model:<model_name>:<model_version>:prediction_latency`: Statistics on latency measured from when Clipper sends datapoints to your model-container to when it receives a response. The value will always be lower than that of the former metric.
```
dcrankshaw: Remove the last sentence under this bullet. The model prediction latency can definitely be higher than prediction latency, because predictions will return early if the SLO is set to be lower than the model latency.

src/benchmarks/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 ##################################
 # Benchmark executable
 
-add_executable(end_to_end_bench src/end_to_end_bench.cpp)
+add_executable(end_to_end_bench src/end_to_end_bench.cpp src/bench_utils.cpp src/include/bench_utils.hpp)
```
dcrankshaw: Put the `bench_utils.hpp` file in `src/`. In this case, nothing outside of `src/benchmarks/src` should need to include `bench_utils` so it's cleaner to just keep everything in one directory.

src/benchmarks/src/end_to_end_bench.cpp
```diff
@@ -1,227 +1,261 @@
...
+        bench_metrics.latency_->insert(r.duration_micros_);
+        bench_metrics.num_predictions_->increment(1);
+        bench_metrics.throughput_->mark(1);
+      });
     }
-    std::this_thread::sleep_for(std::chrono::milliseconds(batch_delay_millis));
+
+    delay_micros =
+        draw_from_poisson ? distribution(generator) : batch_delay_micros;
+    std::this_thread::sleep_for(std::chrono::microseconds(delay_micros));
```
dcrankshaw: This is fine, but one thing to note is that there is a minimum amount of time a thread can sleep for, even if you set sleep_time to 0. Just as an educational aside.

dcrankshaw (changes requested):
Mostly there. Just some formatting and cleanup.

SumDockerfile
nishadsingh1: Sorry, I meant to respond to this -- that's exactly it.

SumBenchDockerfile
nishadsingh1: Sounds good! Thanks for the catch.

bench/README.md
nishadsingh1: Good point; that didn't occur to me.

src/benchmarks/src/end_to_end_bench.cpp
nishadsingh1: That is good to know -- we can just skip the sleep code if the `delay_micros` is <= 0

dcrankshaw (approved):

PR #226, opened 2017-06-22
App/model links (add only)

nishadsingh1:
This PR addresses the third checkbox in #218. It only allows for adding links -- I'll work on a PR that allows for removing links after this gets edited and merged.

Given that users cannot yet remove links and that we limit apps to having at most 1 linked model, the only real difference this PR introduces to Clipper users is that app's candidate models (we now call them linked models) can be specified after app registration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,6 @@ Success!
 >>> clipper.get_all_apps()
 [u'hello_world']
 
-# Let Clipper know that queries to your app should be served by
-# a model with name "feature_sum_model".
->>> clipper.link_model_to_app("hello_world", "feature_sum_model")
-Success!
-
 # Define a simple model that just returns the sum of each feature vector.
 # Note that the prediction function takes a list of feature vectors as
 # input and returns a list of strings.
@@ -68,6 +63,11 @@ Success!
 
 # Deploy the model, naming it "feature_sum_model" and giving it version 1
 >>> clipper.deploy_predict_function("feature_sum_model", 1, feature_sum_function, "doubles")
+
+# Let Clipper know that queries to your app should be served by
+# your newly deployed model.
+>>> clipper.link_model_to_app("hello_world", "feature_sum_model")
+Success!
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -46,15 +46,19 @@ Checking if Docker is running...
 >>> clipper.start()
 Clipper is running
 
-# Register an application called "hello_world" that will query a model
-# called "feature_sum_model". This will create a prediction REST endpoint at
-# http://localhost:1337/hello_world/predict
+# Register an application called "hello_world". This will create
+# a prediction REST endpoint at http://localhost:1337/hello_world/predict
 >>> clipper.register_application("hello_world", "feature_sum_model", "doubles", "-1.0", 100000)
 Success!
 
 # Inspect Clipper to see the registered apps
 >>> clipper.get_all_apps()
 [u'hello_world']
+
+# Let Clipper know that queries to your app should be served by
+# a model with name "feature_sum_model".
+>>> clipper.link_model_to_app("hello_world", "feature_sum_model")
+Success!
 
 # Define a simple model that just returns the sum of each feature vector.
 # Note that the prediction function takes a list of feature vectors as
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Clipper is running
 
 # Register an application called "hello_world". This will create
 # a prediction REST endpoint at http://localhost:1337/hello_world/predict
->>> clipper.register_application("hello_world", "feature_sum_model", "doubles", "-1.0", 100000)
+>>> clipper.register_application("hello_world", "doubles", "-1.0", 100000)
 Success!
 
 # Inspect Clipper to see the registered apps
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -379,7 +379,7 @@ class Clipper:
             The name of the model to link to the application
         """
         url = "http://%s:%d/admin/add_app_links" % (self.host,
-                                      CLIPPER_MANAGEMENT_PORT)
+                                                    CLIPPER_MANAGEMENT_PORT)
         req_json = json.dumps({
             "app_name": app_name,
             "model_names": [model_name]
@@ -395,16 +395,18 @@ class Clipper:
         ----------
         app_name : str
             The name of the application
+
+        Returns
+        -------
+        list
+            Returns a list of the names of models linked to the app
         """
         url = "http://%s:%d/admin/get_app_links" % (self.host,
-                                      CLIPPER_MANAGEMENT_PORT)
-        req_json = json.dumps({
-            "app_name": app_name
-        })
+                                                    CLIPPER_MANAGEMENT_PORT)
+        req_json = json.dumps({"app_name": app_name})
         headers = {'Content-type': 'application/json'}
         r = requests.post(url, headers=headers, data=req_json)
-        print(r.text)
-
+        return r.json()
 
     def get_all_apps(self, verbose=False):
         """Gets information about all applications registered with Clipper.
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -369,7 +369,9 @@ class Clipper:
         return r.status_code == requests.codes.ok
 
     def link_model_to_app(self, app_name, model_name):
-        """Allows the model to be used by app
+        """
+        Allows the model with `model_name` to be used by the app with `app_name`.
+        The model and app should both be registered with Clipper.
 
         Parameters
         ----------
@@ -391,11 +393,10 @@ class Clipper:
         })
         headers = {'Content-type': 'application/json'}
         r = requests.post(url, headers=headers, data=req_json)
-        print(r.text)
         return r.status_code == requests.codes.ok
 
     def get_linked_models(self, app_name):
-        """Retrieves the models linked to the app
+        """Retrieves the models linked to the specified application
 
         Parameters
         ----------
@@ -405,14 +406,19 @@ class Clipper:
         Returns
         -------
         list
-            Returns a list of the names of models linked to the app
+            Returns a list of the names of models linked to the app.
+            If no models are linked to the specified app, None is returned.
         """
         url = "http://%s:%d/admin/get_linked_models" % (
             self.host, CLIPPER_MANAGEMENT_PORT)
         req_json = json.dumps({"app_name": app_name})
         headers = {'Content-type': 'application/json'}
         r = requests.post(url, headers=headers, data=req_json)
-        return r.json()
+        if r.status_code == requests.codes.ok:
+            return r.json()
+        else:
+            print(r.text)
+            return None
 
     def get_all_apps(self, verbose=False):
         """Gets information about all applications registered with Clipper.
@@ -429,7 +435,7 @@ class Clipper:
         -------
         list
             Returns a list of information about all apps registered to Clipper.
-            If no apps are registered with Clipper, an empty list is returned.
+            If no apps are registered with Clipper, None is returned.
         """
         url = "http://%s:1338/admin/get_all_applications" % self.host
         req_json = json.dumps({"verbose": verbose})
@@ -928,7 +934,7 @@ class Clipper:
         -------
         list
             Returns a list of information about all apps registered to Clipper.
-            If no models are registered with Clipper, an empty list is returned.
+            If no models are registered with Clipper, None is returned.
         """
         url = "http://%s:1338/admin/get_all_models" % self.host
         req_json = json.dumps({"verbose": verbose})
@@ -989,7 +995,7 @@ class Clipper:
         -------
         list
             Returns a list of information about all apps registered to Clipper.
-            If no containerss are registered with Clipper, an empty list is returned.
+            If no containerss are registered with Clipper, None is returned.
         """
         url = "http://%s:1338/admin/get_all_containers" % self.host
         req_json = json.dumps({"verbose": verbose})
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -377,6 +377,11 @@ class Clipper:
             The name of the application
         model_name : str
             The name of the model to link to the application
+
+        Returns
+        -------
+        bool
+            Returns true iff the model link request was successful
         """
         url = "http://%s:%d/admin/add_model_links" % (self.host,
                                                       CLIPPER_MANAGEMENT_PORT)
@@ -387,6 +392,7 @@ class Clipper:
         headers = {'Content-type': 'application/json'}
         r = requests.post(url, headers=headers, data=req_json)
         print(r.text)
+        return r.status_code == requests.codes.ok
 
     def get_linked_models(self, app_name):
         """Retrieves the models linked to the app
@@ -784,14 +790,24 @@ class Clipper:
 
         return deploy_result
 
-    def _register_app_and_check_success(self, name, input_type, default_output,
-                                        slo_micros):
-        if self.register_application(name, name, input_type, default_output,
-                                     slo_micros):
-            print("Application registration sucessful! Deploying model.")
-            return True
-        print("Application registration unsuccessful. Will not deploy model.")
-        return False
+    def _register_app_link_model_check_success(self, name, input_type,
+                                               default_output, slo_micros):
+        if not self.register_application(name, input_type, default_output,
+                                         slo_micros):
+            print(
+                "Application registration unsuccessful. Will not deploy model or link it to app."
+            )
+            return False
+        print("Application registration sucessful! Linking it to model.")
+
+        time.sleep(1)
+        if not self.link_model_to_app(name, name):
+            print(
+                "Linking model to app was unsuccessful. Will not deploy model.")
+            return False
+
+        print("Linking model to app was successful! Deploying model.")
+        return True
 
     def register_app_and_deploy_predict_function(
             self,
@@ -831,7 +847,7 @@ class Clipper:
             The number of replicas of the model to create. More replicas can be
             created later as well.
         """
-        if not self._register_app_and_check_success(
+        if not self._register_app_link_model_check_success(
                 name, input_type, default_output, slo_micros):
             return False
 
@@ -891,7 +907,7 @@ class Clipper:
             The number of replicas of the model to create. More replicas can be
             created later as well.
         """
-        if not self._register_app_and_check_success(
+        if not self._register_app_link_model_check_success(
                 name, input_type, default_output, slo_micros):
             return False
 
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -322,7 +322,7 @@ class Clipper:
             self._execute_root("docker-compose up -d query_frontend")
             print("Clipper is running")
 
-    def register_application(self, name, model, input_type, default_output,
+    def register_application(self, name, input_type, default_output,
                              slo_micros):
         """
         Submits a request to register a new Clipper application and returns whether or
@@ -332,8 +332,6 @@ class Clipper:
         ----------
         name : str
             The name of the application.
-        model : str
-            The name of the model this application will query.
         input_type : str
             One of "integers", "floats", "doubles", "bytes", or "strings".
         default_output : string
@@ -361,7 +359,6 @@ class Clipper:
                                               CLIPPER_MANAGEMENT_PORT)
         req_json = json.dumps({
             "name": name,
-            "candidate_model_names": [model],
             "input_type": input_type,
             "default_output": default_output,
             "latency_slo_micros": slo_micros
@@ -370,6 +367,44 @@ class Clipper:
         r = requests.post(url, headers=headers, data=req_json)
         print(r.text)
         return r.status_code == requests.codes.ok
+
+    def link_model_to_app(self, app_name, model_name):
+        """Allows the model to be used by app
+
+        Parameters
+        ----------
+        app_name : str
+            The name of the application
+        model_name : str
+            The name of the model to link to the application
+        """
+        url = "http://%s:%d/admin/add_app_links" % (self.host,
+                                      CLIPPER_MANAGEMENT_PORT)
+        req_json = json.dumps({
+            "app_name": app_name,
+            "model_names": [model_name]
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        print(r.text)
+
+    def get_linked_models(self, app_name):
+        """Retrieves the models linked to the app
+
+        Parameters
+        ----------
+        app_name : str
+            The name of the application
+        """
+        url = "http://%s:%d/admin/get_app_links" % (self.host,
+                                      CLIPPER_MANAGEMENT_PORT)
+        req_json = json.dumps({
+            "app_name": app_name
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        print(r.text)
+
 
     def get_all_apps(self, verbose=False):
         """Gets information about all applications registered with Clipper.
--- a/clipper_admin/clipper_manager.py
+++ b/clipper_admin/clipper_manager.py
@@ -804,15 +804,15 @@ class Clipper:
                 "Application registration unsuccessful. Will not deploy model or link it to app."
             )
             return False
-        print("Application registration sucessful! Linking it to model.")
+        print("Application registration sucessful! Deploying model.")
+        return True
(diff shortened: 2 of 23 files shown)


nishadsingh1:
Looks like it's a formatting error: https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/463/consoleFull
"Found 1 C++ style violations.
Run format_code.sh to fix."

My code formatting script doesn't recommend any changes. Would one of you mind pushing up a formatting fix?

nishadsingh1:
Thanks @dcrankshaw

nishadsingh1:
"Build timed out (after 120 minutes). Marking the build as aborted.
Build was aborted" -- is this normal / should we just restart the build?

nishadsingh1:
jenkins test this please

nishadsingh1:
jenkins test this please

nishadsingh1:
jenkins test this please

nishadsingh1:
jenkins test this please

shaneknapp:
test this please

shaneknapp:
test this please

nishadsingh1:
Somehow, tests passed this time without any functional changes (just added print statements):
```
[ RUN      ] RedisTest.SubscriptionDetectModelLinksAdd
[12:08:01.529][info]       [REDIS] Successfully issued command "FLUSHALL"
[12:08:01.529][info]       [REDIS] Successfully issued command "CONFIG SET notify-keyspace-events AKE"
[12:08:01.529][info]       [REDIS] Successfully issued command "SELECT 5"
[12:08:01.529][info]       [REDIS] Successfully issued command "HMSET my_app_name input_type doubles policy exp3_policy default_output 1.0 latency_slo_micros 10000"
[12:08:01.529][info]       [REDIS] SUBSCRIPTION STRING: __keyspace@7__:*
[12:08:02.030][info]   [REDISTEST] Before add_mode_links
[12:08:02.030][info]       [REDIS] Successfully issued command "SELECT 7"
[12:08:02.031][info]       [REDIS] Successfully issued command "SADD my_app_name model1"
[12:08:02.031][info]       [REDIS] MESSAGE: sadd
[12:08:02.031][info]   [REDISTEST] Received event
[12:08:02.031][info]   [REDISTEST] Confirmed event is SADD
[12:08:02.031][info]   [REDISTEST] Got lock in subscription callback
[12:08:02.031][info]   [REDISTEST] Before cv notify
[12:08:02.031][info]   [REDISTEST] After cv notify
[12:08:02.031][info]       [REDIS] MESSAGE: sadd
[12:08:02.031][info]   [REDISTEST] Received event
[12:08:02.031][info]   [REDISTEST] Confirmed event is SADD
[12:08:02.031][info]   [REDISTEST] Got lock in subscription callback
[12:08:02.031][info]   [REDISTEST] Before cv notify
[12:08:02.031][info]   [REDISTEST] After cv notify
[12:08:02.031][info]       [REDIS] Successfully issued command "SADD my_app_name model2"
[12:08:02.031][info]   [REDISTEST] After add_mode_links
[12:08:02.031][info]   [REDISTEST] Got lock outside of subscription callback
[12:08:02.031][info]   [REDISTEST] After waiting on cv
[       OK ] RedisTest.SubscriptionDetectModelLinksAdd (1527 ms)
```

Maybe I'll wait on getting this PR into its final format before worrying about why Jenkins tests are failing/passing on this code (since these occasional failures seem to be specific to this PR)

src/libclipper/src/redis.cpp
```diff
@@ -438,6 +449,22 @@ bool add_application(redox::Redox& redis, const std::string& appname,
   }
 }
 
+bool add_model_links(redox::Redox& redis, const std::string& appname,
+                     const std::vector<std::string>& model_names) {
+  if (send_cmd_no_reply<string>(
+          redis, {"SELECT", std::to_string(REDIS_APP_MODEL_LINKS_DB_NUM)})) {
+    for (auto model_name : model_names) {
```
nishadsingh1, line 461: We're making multiple requests to Redis here instead of sending the whole set of `model_names` at once. This is because I couldn't find a way to add multiple elements to a set using the available `send_cmd_no_reply` infrastructure in one call. Let me know if there's a way to do this.

It's possible that no such way exists and sending a request for each new linked model is a bad practice. In this case, we could just have this REDIS_APP_MODEL_LINKS table store a mapping between app names and a string. The string would be a serialized list of model names (maybe with one of our predefined list delimiters). Then, whenever we want to fetch model links, we'd just deserialize the list. When adding new ones, we'd fetch the list, deserialize it, add to it, and re-serialize, and send it to Redis.

For now it's possible this isn't a big deal because we enforce that `model_names` has no more than 1 entry.

README.md
```diff
@@ -46,16 +46,20 @@ Checking if Docker is running...
 >>> clipper.start()
 Clipper is running
 
-# Register an application called "hello_world" that will query a model
-# called "feature_sum_model". This will create a prediction REST endpoint at
-# http://localhost:1337/hello_world/predict
+# Register an application called "hello_world". This will create
+# a prediction REST endpoint at http://localhost:1337/hello_world/predict
 >>> clipper.register_application("hello_world", "feature_sum_model", "doubles", "-1.0", 100000)
```
Corey-Zumar: This no longer conforms to the method prototype of `register_application`. We should remove `"feature_sum_model"`.

clipper_admin/clipper_manager.py
```diff
@@ -371,6 +368,52 @@ def register_application(self, name, model, input_type, default_output,
         print(r.text)
         return r.status_code == requests.codes.ok
 
+    def link_model_to_app(self, app_name, model_name):
+        """Allows the model to be used by app
```
Corey-Zumar: How does "Allows the model with `model_name` to be used by the app with `app_name`" sound?

We should also note that `app_name` should be a preexisting application that has already been registered with clipper, while the model does not have to be registered.

clipper_admin/clipper_manager.py
```diff
@@ -371,6 +368,52 @@ def register_application(self, name, model, input_type, default_output,
...
+        """
+        url = "http://%s:%d/admin/add_model_links" % (self.host,
+                                                      CLIPPER_MANAGEMENT_PORT)
+        req_json = json.dumps({
+            "app_name": app_name,
+            "model_names": [model_name]
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        print(r.text)
```
Corey-Zumar, line 396: Remove extraneous print statement

src/management/src/management_frontend.hpp
```diff
@@ -414,13 +472,89 @@ class RequestHandler {
...
+      // We asserted earlier that `model_names` has size 1
+      std::string new_model_name = model_names[0];
+
+      if (std::find(existing_linked_models.begin(),
+                    existing_linked_models.end(),
+                    new_model_name) != existing_linked_models.end()) {
+        return "Success!";
+      } else {
+        std::stringstream ss;
+        ss << "There already exist links to different models for app "
```
Corey-Zumar: Nit: `A model with name <model_name> is already linked to <app_name>` is cleaner.

clipper_admin/clipper_manager.py
```diff
@@ -371,6 +368,52 @@ def register_application(self, name, model, input_type, default_output,
...
+        return r.status_code == requests.codes.ok
+
+    def get_linked_models(self, app_name):
+        """Retrieves the models linked to the app
```
Corey-Zumar: Replace "app" with "specified application".

examples/tutorial/tutorial_part_one.ipynb
```diff
@@ -220,14 +220,12 @@
    "outputs": [],
    "source": [
     "app_name = \"cifar_demo\"\n",
-    "model_name = \"birds_vs_planes_classifier\"\n",
-    "# If the model doesn't return a prediction in time, predict\n",
-    "# label 0 (bird) by default\n",
+    "# If the model (which we will later link to our apo) doesn't\n",
```
Corey-Zumar: Nit: typo. `apo` >> `application`

examples/tutorial/tutorial_part_one.ipynb
```diff
@@ -253,6 +251,52 @@
...
+   "metadata": {},
+   "source": [
+    "## Link your app to a model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Clipper applications don't generate predictions themselves; models do. Let Clipper know that your `\"cifar_demo\"` app should use the `\"birds_vs_planes_classifier\"` model to serve predictions once it is deployed.\n",
```
Corey-Zumar: "In order to generate predictions, a Clipper application needs to be linked to a model. Let Clipper know that your..." is clearer.

examples/tutorial/tutorial_part_one.ipynb
```diff
@@ -197,11 +197,11 @@
    "source": [
     "# Create an application\n",
     "\n",
-    "In order to query Clipper for predictions, you need to create an application. Each application specifies a name, a set of models it can query, the query input datatype, the selection policy, and a latency service level objective. Once you register an application with Clipper, the system will create two REST endpoints: one for requesting predictions and for providing feedback.\n",
+    "In order to query Clipper for predictions, you need to create an application. Each application specifies a name, the query input datatype, the selection policy, and a latency service level objective. Once you register an application with Clipper, the system will create two REST endpoints: one for requesting predictions and for providing feedback.\n",
```
Corey-Zumar: While we're here, let's remove the reference to the "feedback" endpoint, as we don't currently support it or incorporate it as part of the tutorial.

examples/tutorial/tutorial_part_one.ipynb
```diff
@@ -253,6 +251,52 @@
...
+   "source": [
+    "model_name = \"birds_vs_planes_classifier\"\n",
+    "clipper.link_model_to_app(app_name, model_name)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can view which models your app is linked to by running the below code."
```
Corey-Zumar: Nit: `below code` >> `code below`

src/frontends/src/query_frontend.hpp
```diff
@@ -156,16 +156,34 @@ class RequestHandler {
...
+          clipper::log_info_formatted(
+              LOGGING_TAG_QUERY_FRONTEND,
+              "APP LINKS EVENT DETECTED. App name: {}, event_type: {}",
+              app_name, event_type);
+          if (event_type == "sadd") {
+            clipper::log_info_formatted(LOGGING_TAG_QUERY_FRONTEND,
+                                        "New model link detected for app: {}",
+                                        app_name);
+
+            boost::optional<std::vector<std::string>> linked_model_names =
```
Corey-Zumar: Why do we reference `linked_model_names` as an`optional` here when `clipper::redis::get_linked_models` returns a vector?

src/frontends/src/query_frontend.hpp
```diff
@@ -156,16 +156,34 @@ class RequestHandler {
...
+                clipper::redis::get_linked_models(redis_connection_, app_name);
+            if (linked_model_names) {
```
Corey-Zumar: If `linked_model_names` does not need to be an optional, we can instead check whether or not it's empty.

src/libclipper/include/clipper/redis.hpp
```diff
@@ -182,6 +180,14 @@ std::vector<std::string> get_all_model_names(redox::Redox& redis);
 std::vector<VersionedModelId> get_all_models(redox::Redox& redis);
 
 /**
+ * Looks up which models are linked to app with name `app_name`
+ * \return Returns a vector of model names. If no models are linked
+ * to the specified app, then am empty vector will be returned.
```
Corey-Zumar: Nit: Typo `am` >> `an`

clipper_admin/clipper_manager.py
```diff
@@ -371,6 +368,52 @@ def register_application(self, name, model, input_type, default_output,
...
+        req_json = json.dumps({
+            "app_name": app_name,
+            "model_names": [model_name]
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        print(r.text)
+        return r.status_code == requests.codes.ok
+
+    def get_linked_models(self, app_name):
```
Corey-Zumar, line 399: This results in `ValueError: No JSON object could be decoded` in the case that `app_name` is not a registered application.

src/frontends/src/query_frontend.hpp
Corey-Zumar: This method may not be necessary (see my comment below about whether or not it's a good idea to store link information locally)

src/frontends/src/query_frontend.hpp
```diff
@@ -248,9 +268,20 @@ class RequestHandler {
...
-  void add_application(std::string name, std::vector<std::string> models,
-                       InputType input_type, std::string policy,
-                       std::string default_output, long latency_slo_micros) {
+  void set_linked_models_for_app(std::string name,
+                                 std::vector<std::string> models) {
+    std::unique_lock<std::mutex> l(linked_models_for_apps_mutex_);
+    linked_models_for_apps_[name] = models;
+  }
+
+  std::vector<std::string> get_linked_models_for_app(std::string name) {
```
Corey-Zumar, line 278: What's the rationale behind storing a copy of `linked_models_for_apps_` rather than querying Redis when a request is made for the data? These kinds of queries are probably infrequent and won't present an excessive load on the database. Unless there's a good reason, we should rely on Redis as a source of truth.

src/libclipper/src/json_util.cpp
```diff
@@ -384,6 +384,33 @@ std::string to_json_string(rapidjson::Document& d) {
   return buffer.GetString();
 }
 
+void set_string_array(rapidjson::Document& d,
+                      const std::vector<std::string>& str_array) {
+  d.SetArray();
+  for (auto str : str_array) {
```
Corey-Zumar: This should be `for (auto const &str: str_array) {`. The reason being that using `auto` alone iterates by value, which I'm pretty sure incurs a copy.

src/libclipper/include/clipper/redis.hpp
```diff
@@ -251,12 +257,20 @@ std::vector<std::pair<VersionedModelId, int>> get_all_containers(
...
                      const std::string& default_output,
                      const long latency_slo_micros);
 
 /**
+ * Adds links between the specified app and models. This will not
+ * overwrite existing links.
+ *
+ * \return Returns true if the add was successful.
+ */
+bool add_model_links(redox::Redox& redis, const std::string& appname,
```
Corey-Zumar: Nit: `appname` >> `app_name`

src/libclipper/src/redis.cpp
```diff
@@ -209,6 +205,24 @@ boost::optional<std::string> get_current_model_version(
...
 }
 
+std::vector<std::string> get_linked_models(redox::Redox& redis,
+                                           const std::string& app_name) {
+  std::vector<std::string> linked_models;
+  if (send_cmd_no_reply<string>(
+          redis, {"SELECT", std::to_string(REDIS_APP_MODEL_LINKS_DB_NUM)})) {
+    auto result =
+        send_cmd_with_reply<std::vector<string>>(redis, {"SMEMBERS", app_name});
+    linked_models = *result;
```
Corey-Zumar: We should check that `result` is present before attempting to dereference it from the `optional` container.

src/libclipper/src/redis.cpp
Corey-Zumar: @dcrankshaw Any advice here?

It's true that this is fine for now because of the restriction that an application may only have one backing model. However, it would be nice to write this efficiently for future use cases when the restriction is loosened.

src/management/src/management_frontend.hpp
```diff
@@ -64,13 +68,25 @@ const std::string GET_CONTAINER = ADMIN_PATH + "/get_container$";
...
    "input_type" := "integers" | "bytes" | "floats" | "doubles" | "strings",
    "default_output" := string,
    "latency_slo_micros" := int
   }
 )";
 
+const std::string ADD_MODEL_LINKS_JSON_SCHEMA = R"(
+  {
+    "app_name" := string,
+    "model_names" := := [string]
```
Corey-Zumar: Nit: redundant `:=`

Corey-Zumar (changes requested):
Functionality is almost complete! I really like the new link interface. Most of my comments are cosmetic.

nishadsingh1:
- [x] Make sure models exist before linking them

nishadsingh1:
Thanks @Corey-Zumar for the review -- the latest commit should address your comments and enforce that models are registered with Clipper before linking them to any apps.

Tests are passing locally for me; they're failing on Jeknins because the spark scala container isn't building correctly (fix attempt at #248), though of course there may be legitimate failures too that will show up later.

Even without the passing Jenkins tests, I think this is still ready for another pass-over.

nishadsingh1:
Looks like it's hanging on `RedisTest. SubscriptionDetectModelLinksAdd`, even when print statements are added back in.

shaneknapp:
i just checked the worker that this test ran on and it's got a TON zombie
processes.  this could be causing issues.

i'll need to reboot this machine to get rid of them.

clipper_admin/clipper_manager.py
```diff
@@ -371,6 +368,58 @@ def register_application(self, name, model, input_type, default_output,
...
+            Returns true iff the model link request was successful
+        """
+        url = "http://%s:%d/admin/add_model_links" % (self.host,
+                                                      CLIPPER_MANAGEMENT_PORT)
+        req_json = json.dumps({
+            "app_name": app_name,
+            "model_names": [model_name]
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
```
Corey-Zumar, line 395: On second thought, let's re-add the call to `print(r.text)`. It's very helpful for user understanding - sorry!

src/libclipper/src/redis.cpp
```diff
@@ -438,6 +454,22 @@ bool add_application(redox::Redox& redis, const std::string& appname,
   }
 }
 
+bool add_model_links(redox::Redox& redis, const std::string& appname,
```
Corey-Zumar, line 457: Currently, this method allows users to link an application of input type A to a model of input type B. We should verify that the input types are the same and fail if they aren't. Let's add a test for this as well.

src/libclipper/src/redis.cpp
Corey-Zumar: Resolved. This function is not on any critical paths, so making a few extra database queries will not have any significant impact on performance.

Corey-Zumar (changes requested):
Almost there - just missing some input type verification during the linking process

src/libclipper/src/redis.cpp
nishadsingh1: Great catch :) I think it's correct to do this in `redis.cpp`, but we could also do it in the management front end. Just want to confirm that `redis.cpp` is the right place for it - thoughts?

clipper_admin/clipper_manager.py
nishadsingh1: Np at all

src/libclipper/src/redis.cpp
nishadsingh1: Actually, I'm going to redistribute the validation responsibilities.
Management frontend:
- Confirm that the user supplied only one model_name
- Make sure that there will only be one link

Redis:
- Confirm that an app with specified app name exists
- Confirm that a model exists for each specified model name
- Confirm input types match

It seems like we'd want the checks that guarantee important logic to exist at the single entry point to the database (redis.cpp). Let me know if that sounds reasonable.

--------------------
Edit:  It looks like we aren't propagating errors up from any of the functions in `redis.cpp`, so there isn't a standard way to handle them (and access their specific error messages) from the management frontend function.

We should definitely still let the user know when their specified app/model doesn't exist or if the input types don't match, instead of returning some nondescriptive error message in all cases. In order to give the appropriate error message, we need the management frontend function to make the checks that are under the 'Redis' section above.

Should we just duplicate the checks like so?:

Management frontend:
- Confirm that the user supplied only one model_name
- Make sure that there will only be one link
- Confirm that an app with specified app name exists
- Confirm that a model exists for each specified model name
- Confirm input types match

Redis:
- Confirm that an app with specified app name exists
- Confirm that a model exists for each specified model name
- Confirm input types match

I think this is probably okay since adding model links isn't on a critical path.

--------------------
Edit 2: We don't seem to do much validation at all in `redis.cpp`. Is this intentional / does this mean that we should only do this validation in the management frontend?

nishadsingh1:
**Problematic scenario 1:**
- Register a model with name 'M', version '1', and input_type 'ints'.
- Register an app with name 'A' and input_type 'ints'
- Add a model link between 'M' and 'A'. Input types match, both apps exist, only one model name was specified and the app had no previous links: this request is successful.
- Register a model with name 'M', version '2', and input_type 'doubles'.

Queries to 'A' route to the latest version of 'M'. In this case, that would send requests to 'M:2', which has an incompatible input_type. Under the hood, registering 'M:2' set the current version of 'M' to '2'. This was the source of the problem.

**Problematic scenario 2**
- Register a model with name 'M', version '1', and input_type 'ints'.
- Register a model with name 'M', version '2', and input_type 'doubles'.
- Register an app with name 'A' and input_type 'doubles'
- Add a model link between 'M' and 'A'. Input types match, both apps exist, only one model name was specified and the app had no previous links: this request is successful.
- Set the current model version to '1'. 

Queries to 'A' route to 'M:1' which has an incompatible input_type.

So, it seems like we need to make sure that every time we set the current model version of a model, we're not creating some conflict with an existing linked application with respect to input type.

**Where to do this validation?**
If we put it in management frontend, we'd need to do the check in both `add_model` and `set_model_version`.

If we put it in redis.cpp's, we'd only need to do the check in `set_current_model_version`.

I like the latter option. If every time we change a model's model version, shouldn't we just have the check be in the function (or on a guaranteed path to the function ) that makes the change? Otherwise we run the risk of forgetting to run these checks with each usage of redis' `set_current_model_version` and creating bugs.

------
Update: we'll have all checks be in Management Frontend for now and move them to Redis in another PR

src/libclipper/src/redis.cpp
nishadsingh1: We resolved to just have all the checks be in the Management Frontend and to move them to Redis in another PR

nishadsingh1:
Updates on the latest change:
- Addressed Corey's comments
- Made sure that input types matched for new links
- Made sure that when adding a new model, there exist no previous versions were linked to apps under different input types
- Made sure that when setting a new model version, there exist no links to applications under a the current model's input type that would be incompatible under the version change
- In the process of the latter two points, rewrote the frontend set_model_version code to match the form of our other frontend endpoints (plus, the change was necessary to avoid some uglier code changes to accommodate for the above checks)

src/management/src/management_frontend.hpp
```diff
@@ -511,23 +643,52 @@ class RequestHandler {
...
+
+    if (!existing_model_data.empty()) {
       std::stringstream ss;
-      ss << "Error model " << model_id.get_name() << ":" << model_id.get_id()
+      ss << "Error model " << model_name << ":" << model_version
          << " already exists";
       throw std::invalid_argument(ss.str());
     }
+
+    // Need to fux with this
```
Corey-Zumar: Remove comment. Are there outstanding changes that need to be made?

src/management/src/management_frontend.hpp
```diff
@@ -836,8 +1042,35 @@ class RequestHandler {
...
+   *  "model_version" := string,
+   * }
+   */
+  std::string set_model_version(const std::string& json) {
+    rapidjson::Document d;
+    parse_json(json, d);
+    std::string model_name = get_string(d, "model_name");
+    std::string new_model_version = get_string(d, "model_version");
+
+    check_updated_model_consistent_with_app_links(model_name,
```
Corey-Zumar: The `check_updated_model_consistent_with_app_links` method is being invoked with `new_model_version` as the second parameter, as opposed to an `input_type` value.

Also, `attempt_model_version_update` calls the same method again. Does it need to be called in both places?

Corey-Zumar (changes requested):
one more issue & doc cleanup. ready to merge when these changes are made and tests pass.

Corey-Zumar (approved):
Looks good. Will merge when tests pass. Looks like there's an issue with pyspark model deployment.

dcrankshaw:
Actually hold off on merging this. This is not backwards compatible with the 0.1 release and will cause problems due to #244

dcrankshaw:
Actually you can go ahead and merge it. I turned off the automated builds on docker hub until we can address #244. I'll prioritize that and push out a new 0.1.x release that manages docker versions properly.

PR #228, opened 2017-06-24
[CLIPPER-227] Fix EWMA behavior for meters

### LINKED ISSUE #227: Histogram metric computes 1,5,15 min rates incorrectly
There is a bug in the way the the `metrics::Histogram` computes the running rates. For example, with a steady workload of ~850 qps, the histogram reports the following rates:

```
              "internal:aggregate_model_throughput": {
                  "unit": "events per second",
                  "rate": "862.38770176081448",
                  "rate_1min": "4009.4179095954541",
                  "rate_5min": "11650.076397861654",
                  "rate_15min": "13916.713327875257"
              }
          }
```

Corey-Zumar:
Fixes #227

### DIFF
--- a/src/libclipper/src/metrics.cpp
+++ b/src/libclipper/src/metrics.cpp
@@ -22,7 +22,6 @@ namespace clipper {
 namespace metrics {
 
 constexpr long MICROS_PER_SECOND = 1000000;
-constexpr long CLOCKS_PER_MILLISECOND = CLOCKS_PER_SEC / MICROS_PER_SECOND;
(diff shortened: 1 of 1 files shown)


dcrankshaw (approved):
This looks fine. What was the problem?

Corey-Zumar:
The issue was that marks would compound until they were all included in the EWMA at once upon invocation of `get_n_minute_rate_seconds`. This resulted in much higher, inaccurate EWMA rates. This was fixed by ensuring that we tick every time a mark occurs.

PR #232, opened 2017-06-27
RPC container content fix

### LINKED ISSUE #231: Clipper RPC Layer fails to fully process content messages under certain conditions
In the case that a container content (prediction response) message is received by Clipper for an unrecognized container, the message is never fully read from the socket. Subsequent messages are then corrupted when unread fields from the partially-read content message are mistakenly polled from the socket later on. This may occur when Clipper restarts after issuing a high-latency prediction request.

Corey-Zumar:
Fixes #231

### DIFF
--- a/containers/python/rpc.py
+++ b/containers/python/rpc.py
@@ -302,7 +302,7 @@ class Server(threading.Thread):
 
                         t4 = datetime.now()
 
-                        response.send(socket)
+                        response.send(socket, self.event_history)
 
                         print("recv: %f us, parse: %f us, handle: %f us" %
                               ((t2 - t1).microseconds, (t3 - t2).microseconds,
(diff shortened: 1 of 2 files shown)


containers/python/rpc.py
```diff
@@ -394,13 +394,14 @@ def add_output(self, output):
...
 
-    def send(self, socket):
+    def send(self, socket, event_history):
         socket.send("", flags=zmq.SNDMORE)
         socket.send(
             struct.pack("<I", MESSAGE_TYPE_CONTAINER_CONTENT),
             flags=zmq.SNDMORE)
         socket.send(self.msg_id, flags=zmq.SNDMORE)
         socket.send(self.output_buffer[0:self.string_content_end_position])
+        event_history.insert(EVENT_HISTORY_SENT_CONTAINER_CONTENT)
```
Corey-Zumar, line 404: This missing event history log did not contribute to #231 , but I've corrected the logging issue while addressing the bug.

dcrankshaw (approved):
LGTM

PR #237, opened 2017-06-29
Save module dependencies not in conda/pip

### LINKED ISSUE #207: Save module dependencies when serializing Python closures
Deploying Python closures to Clipper will break when the user is using multi-module Python programs and tries to capture functions defined in other modules (that are defined locally, not as part of installed pip packages). We should port PyWren's [module dependency code](https://github.com/pywren/pywren/blob/master/pywren/serialize/module_dependency.py) to Clipper to automatically capture included modules when possible.

nishadsingh1:
Resolves #207 

Seems to work when using prediction functions of both the following forms (where `test_util.py` is some module not in conda or pip that has a `COEFFICIENT=<int>` variable):

```
import test_util
func = lambda x : len(x) * test_util.COEFFICIENT
```

```
import test_util as tu
func = lambda x : len(x) * tu.COEFFICIENT
```


containers/python/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM clipper/py-rpc:latest
 MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
```
nishadsingh1: Building this no longer works, but I'm also not sure why this file is around in the first place.

nishadsingh1:
Looks like something's still going wrong with this PR; not read for review yet.

containers/python/Dockerfile
Corey-Zumar: This won't work because we can't reference files outside of the docker file's build context. The context directory (the one containing the docker file) must reference `clipper_admin` as a subdirectory.

dcrankshaw: Right. This file should be deleted.

nishadsingh1:
jenkins test this please

nishadsingh1:
This redundantly includes the change in #248 -- I wanted to see if the other changes in this PR would pass tests on Jenkins. Should be ready for review.

nishadsingh1:
jenkins test this please

shaneknapp:
test this please

nishadsingh1:
jenkins test this please

nishadsingh1:
jenkins test this please

clipper_admin/clipper_manager.py
```diff
@@ -578,13 +584,119 @@ def _save_python_function(self, name, predict_function):
...
+
+    def _export_local_modules(self, func_modules, serialization_dir,
+                              conda_deps_fname, pip_deps_fname,
+                              local_modules_folder_name,
+                              local_modules_list_fname):
+        """
+        Supplies modules in `func_modules` not already accounted for by Anaconda or pip to container.
+        Copies those modules to a location specified by `serialization_dir` and `local_modules_folder_name`
+        Enumerates the local modules copied in a file whose location is specified by `serialization_dir` and
+        `local_module_deps_fname`.
```
dcrankshaw: By "location is specified by serialization_dir and local_module_deps_fname" you mean it will be at `serialization_dir/local_module_deps_fname`?

clipper_admin/module_dependency.py
```diff
@@ -0,0 +1,275 @@
+"""
+BSD Licensed
+From 
+https://github.com/cloudpipe/multyvac-fork/blob/master/multyvac/util/module_dependency.py
+"""
```
dcrankshaw, line 5: Was this file copied without modification or should I review it?

containers/python/python_container.py
```diff
@@ -17,9 +19,22 @@ def load_predict_func(file_path):
         return cloudpickle.load(serialized_func_file)
 
 
+def load_modules(modules_list_path):
+    with open(modules_list_path, 'rb') as f:
+        module_fnames = pickle.load(f)
+        for m in module_fnames:
+            importlib.import_module(m)
```
dcrankshaw: Interesting. Why do you need to manually import these?

integration-tests/util/nested_test_util.py
```diff
@@ -0,0 +1,8 @@
+"""
```
dcrankshaw: This isn't really a "test_util". Can you rename this to `nested_mock_module`?

integration-tests/util/test_util.py
```diff
@@ -0,0 +1,12 @@
+"""
```
dcrankshaw: Rename this to `mock_module`?

integration-tests/clipper_manager_tests.py
```diff
@@ -8,10 +8,13 @@
 from argparse import ArgumentParser
 cur_dir = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.abspath('%s/../' % cur_dir))
+sys.path.insert(0, os.path.abspath('%s/util/' % cur_dir))
 import clipper_admin.clipper_manager
 Clipper = clipper_admin.clipper_manager.Clipper
+sys.path.insert(0, os.path.abspath(cur_dir))
```
dcrankshaw: Remove these path insert calls. See my comment below about adding __init__.py files.

integration-tests/util/test_util.py
```diff
@@ -0,0 +1,12 @@
...
+A local module that is used by a predict function. This file is used to
+test the functionality of exporting modules not found in conda or pip.
+"""
+
+import os
+import sys
+cur_dir = os.path.dirname(os.path.abspath(__file__))
+sys.path.insert(0, os.path.abspath(cur_dir))
```
dcrankshaw: Instead of adding the path manually, add __init__.py files to create a proper module system and import normally. I think you'll need to add the following __init__.py files:

```
integration_tests/__init__.py
integration_tests/util/__init__.py
```

integration-tests/clipper_manager_tests.py
dcrankshaw: To expand on this a bit: The reason we add the parent directory to the path explicitly is because we don't want to make the entire Clipper repo a Python package. But it's okay for the integration tests subdirectory to be a python package.

dcrankshaw (changes requested):
Looks great. Just a few minor comments.

clipper_admin/module_dependency.py
nishadsingh1: Copied without modification

containers/python/python_container.py
nishadsingh1: Ah, you're right -- looks like we don't need to import, just make the importable by adding their enclosing folder to sys.path.

clipper_admin/clipper_manager.py
nishadsingh1: Yep -- making that explicit

nishadsingh1:
There's now a different mechanism for importing modules from packages (`import util.mock_module`) and importing modules directly by having them in your path. That's why the integration test suites import the two modules in different ways.

In `clipper_manager_tests.py`, there's only one test that queries the python-container, `test_deployed_and_linked_predict_function_queried_successfully`. This test now makes use of both local module importing mechanisms. There isn't a separate tests for querying a python-container that uses these because the bulk of the interesting work being done by the `deploy_predict_func` feature is package management.

By contrast, the old pyspark-container tests remain intact and a new one has been added to test the local module importing. The `deploy_pyspark_model` feature needs to coordinate loading a pyspark model into the container, and testing that behavior independent of package management might clarify where bugs exist if/when they pop up.

clipper_admin/clipper_manager.py
```diff
@@ -628,13 +630,136 @@ def _save_python_function(self, name, predict_function):
...
+                        print(
+                            "Could not identify the location of the root package {package} for module {module}. Will not supply it to the container.".
+                            format(
+                                package=module_name_split[0],
+                                module=module_name))
+            else:
+                mda.ignore(module_name)
+        module_paths = mda.get_and_clear_paths()
+
+        def path_already_captured(path):
```
nishadsingh1, line 736: Let's say your predict function makes use of some package-based imported module: `import my_package.my_module`. Right now, we add the path to `.../my_package/` to `paths_to_copy` and plan on recursively copying the whole folder to the container later. The location of my_package on the container would be: `/model/modules/my_package`. These will be accessible for import because `/model/modules` will be added to container's sys.path in the container code.

We also add that module name, `my_package.my_module`, to our Module Dependency Analyzer to find all the files that the module is dependent on; it's possible that some don't exist in the folder `.../my_package/` and we still want to copy them to the container. For each file with path `.../<file_name>`, its copy location on the container will be `/model/modules/<file_name>`, and these will be accessible to directly import because `/model/modules` will be added to container's sys.path in the container code.

In order to get the paths to all these files, we run `module_paths = mda.get_and_clear_paths()`, `module_paths` will certainly have files contained in `my_package/...`. To avoid copying multiple versions of these files, we check here if each is already captured (using `path_already_captured`), and don't add it to the set of lists we'll copy if it is.

This seems good in most cases because it prevents us from doing a bunch of unnecessary copies. However, there's a case (seems kind of unlikely) where it will cause an ImportError. Let's say that your prediction function makes use of two modules: one imported via package (`import my_package.my_module`), and one imported directly by adding it's enclosing folder to your path and running (`import my_module_2`). Let's say that `my_module_2` resides under `my_package/`.

Now, we'll run into an ImportError in the container because it won't be able to import the module with name `my_module_2`; it expects the module file to be in it's path (probably at `/model/modules/my_module_2.py`, but we prevented it from getting copied there because we were already going to copy it to `/model/modules/my_package/my_module_2.py`.

I'm not sure how much we care about this case. If we don't think it's costly to have a bunch of unnecessary file copies, then I think we should remove the code responsible for avoiding them (lines 736-744 here). Do you have any thoughts?

nishadsingh1:
Forgot to ping you earlier @dcrankshaw -- this is now ready for review again.

clipper_admin/clipper_manager.py
dcrankshaw: We discussed this in person, but to re-iterate, let's keep the de-duplication logic (that seems useful).

Let's support the case when users import packages but not try to support when users add the enclosing folder to `sys.path`. That's sort of an anti-pattern in Python (despite the fact we make occasional use of it for good reasons within this repo) and it's okay to not support it.

dcrankshaw (approved):
LGTM

dcrankshaw:
jenkins test this please

PR #243, opened 2017-06-30
Update Python container to correctly run "pip install cryptography"

dcrankshaw:
If the user has a version of the cryptography library installed with pip, this change allows the model container to build it from source as the pip package requires. The cryptography library has C extensions, so it requires a compiler to build it.

In particular, this case is important to support because the `clipper_admin` pip package installs the cryptography as a dependency (by way of the fabric dependency).

### DIFF
--- a/PythonContainerDockerfile
+++ b/PythonContainerDockerfile
@@ -2,6 +2,8 @@ FROM clipper/py-rpc:latest
 
 MAINTAINER Dan Crankshaw <dscrankshaw@gmail.com>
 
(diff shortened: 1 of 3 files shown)


nishadsingh1:
Checking this now. Also, is there a reason we don't preinstall `clipper_admin`? We could make a preinstalled pip package list whose elements we install after the conda ones.

Corey-Zumar (approved):
LGTM

dcrankshaw:
jenkins test this please

Corey-Zumar:
jenkins test this please

Issue #245, opened 2017-07-05
Error when using predict (Connection aborted)


rmdort:
We are getting this error occasionally with clipper when using `/predict`

We are sending around 50 prediction requests to clipper (not parallel)

````
ConnectionError: ('Connection aborted.', BadStatusLine("''",))
````
Logs doesnt show anything. What might be the cause?

**Comment by dcrankshaw** [comment]
@Corey-Zumar can you look into this?

**Comment by Corey-Zumar** [comment]
@rmdort Can you provide some information about the latency SLO of your application and approximate input size?

**Comment by Corey-Zumar** [comment]
@dcrankshaw The issue seems to be that we fail to handle exceptions occurring within the context of the future returned by `decode_and_handle_predict()` within the query frontend. As a result, no HTTP response is returned and the connection eventually times out. [httplib](https://docs.python.org/2/library/httplib.html) then throws the insufficiently descriptive connection aborted error for Python 2.7. 

@rmdort This implies that there's an additional error occurring that we'll have trouble debugging until we can handle future exception propagation

**Comment by Corey-Zumar** [comment]
@rmdort After #246 is merged, try replicating the error again and monitor Clipper's logs. Let me know if you see one of the following log messages:
```
[CLIPPER] Unexpected error: The associated promise has been destructed prior to the associated state becoming ready.
```
or
```
[QUERYPR...] Unexpected error: The associated promise has been destructed prior to the associated state becoming ready.
```

rmdort:
ok noted. Latency is set to 320ms. We are facing another issue where query_frontend fails to find the model. 

````
[REDIS] Error with command "GET model-staging:0:0":
[QUERYPR...] No selection state found for query with user_id: 0 and label: model-staging
````

This is happening atleast once a day now. We had to re-deploy the model to fix it.

Corey-Zumar:
@rmdort #246 was merged. Please let me know what kind of log output you're encountering. Will look into the Redis issue

rmdort:
BadStatusLine error seems to be fixed.

But i am getting errors `No connected models` when using `predict`. This happens very very often. Why does the app disconnect from the model?

I always have to re-deploy the model to get it running again.

Any logs i should look at?

Corey-Zumar:
@rmdort Can you send log output for a container that disconnects in the way you describe? Full log output from the query frontend would also be very helpful.

rmdort:
Sorry for the late reply. The containers are always running, but clipper response says `No connected models...` 

[...]

PR #246, opened 2017-07-05
Handling for internal server errors

Corey-Zumar:
If the thread of execution responsible for delivering an HTTP prediction response encounters an unexpected error, this PR ensures that we respond with error code 500 instead of timing out.

### DIFF
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -1,9 +1,12 @@
 #include <cassert>
 #include <iostream>
+#include <sstream>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
+
+#include <boost/exception_ptr.hpp>
 
 #include <clipper/config.hpp>
 #include <clipper/constants.hpp>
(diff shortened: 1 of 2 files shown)


src/frontends/src/query_frontend.hpp
```diff
@@ -287,6 +289,18 @@ class RequestHandler {
             request->content.string(), name, versioned_models, policy,
             latency_slo_micros, input_type);
         prediction.then([response, app_metrics](boost::future<Response> f) {
+          if (f.has_exception()) {
+            try {
+              boost::rethrow_exception(f.get_exception_ptr());
```
dcrankshaw, line 294: Why are we re-throwing the exception here? This will immediately be caught by the catch statement right?

dcrankshaw (changes requested):
Why are you rethrowing the exceptions inside a `try` block?

dcrankshaw:
Does this fix #245?

src/frontends/src/query_frontend.hpp
Corey-Zumar: Yes. If the boost future encounters a fatal exception, it will store it in an exception pointer tied to the future. `rethrow_exception` allows us to obtain the original exception that was thrown and log / handle it appropriately. This flow is a documented way of performing exception propagation when using boost futures / threads.

Corey-Zumar:
@dcrankshaw This fixes the `Connection Aborted` behavior cited in #245. Instead, fatal errors will return HTTP code 500. However, I doubt this fixes the underlying cause of the exception being thrown in #245.

dcrankshaw (approved):

dcrankshaw:
Sounds good. I figured you got that from somewhere, just wanted to double check. I'll merge once the tests pass.

PR #248, opened 2017-07-07
Updated libzmq3 to libzmq5

nishadsingh1:
Attempt to solve breaking spark scala container build.


nishadsingh1:
I think it may take longer to build. Building it locally took ~4.5 min; I can't remember how long it took before but I don't think it was that long.

dcrankshaw (approved):
lgtm

PR #249, opened 2017-07-10
C++ Model Container

Corey-Zumar:


dcrankshaw:
Looks great! Let me know when it's ready for review.

Corey-Zumar:
@dcrankshaw This is ready for review.

Corey-Zumar:
@dcrankshaw Actually, hold off on reviewing this. There's a better way to go about input parsing from a container perspective due to the restriction of inputs to a single type. As a result, it doesn't make sense to rely on Clipper's `Input` datatype subclasses.

Corey-Zumar:
@dcrankshaw This is now ready for review.

containers/container_implementation.md
```diff
@@ -170,7 +170,7 @@ RPC requests sent from Clipper to model containers are divided into two categori
 ### Serializing Prediction Requests
 1. All requests begin with a 32-bit integer header, sent as a single ZeroMQ message. The value of this integer will be 0, indicating that the request is a prediction request.
 
-2. The second ZeroMQ message contains the size of the input header, in bytes, as a 64-bit long. This is the size of the content of the third ZeroMQ message.
+2. The second ZeroMQ message contains the size of the input header, in bytes, as a 32-bit integer. This is the size of the content of the third ZeroMQ message.
```
dcrankshaw, line 173: Were these typos or did you change the RPC implementation?

src/container/include/container/container_parsing.hpp
```diff
@@ -0,0 +1,107 @@
...
+struct input_parser_type {
+  static const bool is_string_parser = false;
+};
+
+template <>
+struct input_parser_type<char> {
+  static const bool is_string_parser = true;
+};
+
+template <typename D>
```
dcrankshaw: What is this template parameter D? Can you add a comment that documents what the meaning of the template type is?

src/container/include/container/container_parsing.hpp
```diff
@@ -0,0 +1,107 @@
...
+  const std::vector<Input<D>> get_primitive_inputs(
+      const std::vector<int>& input_header, long input_content_size) {
+    // For a prediction request containing `n` inputs, there are `n - 1` split
+    // indices
+    int num_splits = input_header[1] - 1;
+    std::vector<Input<D>> inputs;
+    int prev_split = 0;
+    // Iterate from the beginning of the input data content
+    // (the first two elements of the header are metadata)
+    for (int i = 2; i < num_splits + 2; i++) {
```
dcrankshaw: nit: s/i++/++i

src/container/include/container/container_parsing.hpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  const std::vector<Input<D>> get_primitive_inputs(
+      const std::vector<int>& input_header, long input_content_size) {
+    // For a prediction request containing `n` inputs, there are `n - 1` split
+    // indices
+    int num_splits = input_header[1] - 1;
+    std::vector<Input<D>> inputs;
+    int prev_split = 0;
+    // Iterate from the beginning of the input data content
```
dcrankshaw: I'm confused by the term "input data content" here. Does this meaning iterate from the beginning of the header section that specifies where the input data splits are?

src/container/include/container/container_parsing.hpp
dcrankshaw: Why do we iterate to `i < num_splits + 2`? Why the 2 here?

src/container/include/container/container_parsing.hpp
```diff
@@ -0,0 +1,107 @@
...
+  const Input<D> construct_input(std::vector<D>& data_buffer, size_t data_start,
+                                 size_t data_end) {
+    D* input_data = data_buffer.data() + data_start;
+    size_t length = data_end - data_start;
+    Input<D> input(input_data, length);
+    return input;
+  }
+
+  const std::vector<Input<D>> get_string_inputs(
+      const std::vector<int>& /*input_header*/, long input_content_size) {
```
dcrankshaw, line 101: Just to clarify, we don't need the input_header here because the strings are null-terminated strings?

src/container/include/container/container_rpc.hpp
```diff
@@ -0,0 +1,400 @@
...
+        "tcp://" + clipper_ip + ":" + std::to_string(clipper_port);
+    log_info_formatted(
+        LOGGING_TAG_CONTAINER,
+        "Starting container RPC with clipper ip: {} and port: {}", clipper_ip,
+        clipper_port);
+    serving_thread_ = std::thread(
+        [this, clipper_address, &model, model_name, model_version]() {
+          serve_model(model, model_name, model_version, clipper_address);
+        });
+    serving_thread_.detach();
```
dcrankshaw: Don't detach the serving thread, join the thread in the RPC destructor. I know we do this in `RPCService`

src/container/include/container/container_rpc.hpp
```diff
@@ -0,0 +1,400 @@
...
+  void send_container_metadata(std::string& model_name, int model_version,
+                               InputType model_input_type,
+                               zmq::socket_t& socket) const;
+
+  void log_event(rpc::RPCEvent event) const;
+
+  template <typename D>
+  void serve_model(Model<Input<D>>& model, std::string model_name,
+                   int model_version, std::string clipper_address) {
+    zmq::context_t context(1);
```
dcrankshaw: What's context? And what does the 1 mean here?

src/container/include/container/container_rpc.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+            switch (request_type) {
+              case RequestType::PredictRequest:
+                handle_predict_request(model, input_parser, socket,
+                                       input_header_buffer, output_buffer,
+                                       msg_id);
+                break;
+
+              case RequestType::FeedbackRequest:
+                // Do nothing for now
```
dcrankshaw: Should we do nothing or throw an exception?

src/container/include/container/container_rpc.hpp
```diff
@@ -0,0 +1,400 @@
...
+                break;
+
+              default: break;
```
dcrankshaw: Probably want to throw an exception or at least log something if we hit the default case

src/container/include/container/container_rpc.hpp
```diff
@@ -0,0 +1,400 @@
...
+            }
+
+          } break;
+
+          case rpc::MessageType::NewContainer:
+            log_event(rpc::RPCEvent::ReceivedContainerMetadata);
+            log_error_formatted(
+                LOGGING_TAG_CONTAINER,
+                "Received erroneous new container message from Clipper!");
+          default: break;
```
dcrankshaw: Throw an exception or log something if we hit the default case

src/libclipper/include/clipper/datatypes.hpp
```diff
@@ -17,6 +17,7 @@ using QueryId = long;
 using FeedbackAck = bool;
 
 enum class InputType {
+  Invalid = -1,
```
dcrankshaw, line 20: I'm assuming you needed to add this for a reason? Why do we need an invalid type?

dcrankshaw (changes requested):
Excellent. Overall looks great. I mostly had some questions about parts of the implementation I didn't fully understand.

containers/container_implementation.md
Corey-Zumar: I didn't realize that a C++ long is equivalent to a uint32_t, so the correct input header input type has always been  32-bit integer and the docs need this correction.

src/container/include/container/container_parsing.hpp
Corey-Zumar: Exactly. See the bottom of the "Serializing Prediction Requests" section of [container_implementation.md](container_implementation.md)

src/libclipper/include/clipper/datatypes.hpp
Corey-Zumar: I added the `Invalid` type to for the purpose of defining the `input_trait` struct in `container_rpc.hpp`. This trait allows us to infer the `InputType` enum categorization for the based on the type of `Input<D>` that a user's model is parameterized with (`ByteVector`, `DoubleVector`).

src/container/include/container/container_rpc.hpp
Corey-Zumar: Fixed

src/container/include/container/container_rpc.hpp
Corey-Zumar: The context is responsible for initializing and providing IO threads used by ZMQ socket s(See http://api.zeromq.org/2-1:zmq-cpp  # Context). 

The `1` indicates that our context has a single IO thread. I've added documentation to this effect.

src/container/include/container/container_parsing.hpp
Corey-Zumar: Added the following docstring

```
/**
 * Parses raw data of type `D` to create `Input<D>` objects that will be
 * used by models to render predictions.
 *
 * @tparam D A primitive datatype corresponding to a valid Input<D> object
 * (ex: D == `uint8_t` is valid because ByteVector == Input<uint8_t> is a valid Input object)
 */
```

src/container/include/container/container_parsing.hpp
Corey-Zumar: Good catch. This should say "input header content" instead of "input data content". Fixed.

src/container/include/container/container_parsing.hpp
Corey-Zumar: Done

src/container/include/container/container_rpc.hpp
Corey-Zumar: Added an exception

src/container/include/container/container_rpc.hpp
Corey-Zumar: Done - added an exception

src/container/include/container/container_rpc.hpp
Corey-Zumar: Added an exception

Corey-Zumar (commented):

src/container/include/container/container_parsing.hpp
Corey-Zumar: Added the following comment and variable definitions to make things clearer:

```
    // The first two elements of the header indicate the type of input data
    // and the number of content split indices. Therefore, actual splits
    // begin at header index 2 and continue through header index `num_splits` +
    // 2
    int splits_begin_index = 2;
    int splits_end_index = num_splits + 2;
```
The loop is now
```
for (int i = splits_begin_index; i < splits_end_index; ++i) {
    ....
}
```

Corey-Zumar:
@dcrankshaw Build is failing due to code format violations that the `format_code` script isn't catching on my machine. Can you try formatting when you get a chance?

dcrankshaw (approved):
This looks good to me. I formatted the code so as long as the tests pass I'll merge this.

dcrankshaw:
jenkins test this please

PR #250, opened 2017-07-12
Make management frontend check for input type object equality

dcrankshaw:
#226 has a bug in the model/app compatibility check when linking. It checks for input type compatibility through string comparison, but there are multiple valid string representations for each input type. This PR fixes it to compare by `clipper::InputType` object equality which is a canonical representation of input type.

The added unit test fails without the fix and passes with the fix.


nishadsingh1 (approved):
Thanks for catching the bug -- this LGTM.

nishadsingh1:
jenkins test this please

PR #251, opened 2017-07-13
Refactor and clean up Clipper admin python package

dcrankshaw:
This pull request proposes a new implementation of the Clipper admin Python package for managing and deploying a Clipper cluster. The new implementation provides a new API and an internal re-organization of the code. Aside from general cleanup, this implementation makes several improvements. Details and requests for feedback will be coming soon.


dcrankshaw:
@feynmanliang This is finally ready for another round of work on the k8s support. I'd like to treat this branch as a "feature branch" and submit PRs against it until the package is fully ready to be merged.

### Some context
The package has been re-organized pretty substantially and the new package is in `clipper_admin_v2/`. The relevant piece for you is that the container management functionality has been factored out into a [`ContainerManager`](https://github.com/dcrankshaw/clipper/blob/clipper_admin_v2/clipper_admin_v2/clipper_admin/container_manager.py) interface and there are currently two implementations: a pure Docker implementation in `clipper_admin_v2/clipper_admin/docker/docker_container_manager.py` and the K8s implementation you wrote in `clipper_admin_v2/clipper_admin/k8s/k8s_container_manager.py`.

### TODOs for you
[k8s_container_manager.py](https://github.com/dcrankshaw/clipper/blob/clipper_admin_v2/clipper_admin_v2/clipper_admin/k8s/k8s_container_manager.py) should be the only file you have to modify, I stuck some TODOs in there for you. I tried to document what the semantics of each function should be, but you can also look at how [`clipper_admin.py`](https://github.com/dcrankshaw/clipper/blob/clipper_admin_v2/clipper_admin_v2/clipper_admin/clipper_admin.py) uses the ContainerManager interface to see what the functions should do.

If you have any feedback or ideas on how to improve the rest of the package, feel free to implement them or make suggestions, but don't feel obligated to.

### Testing
You'll know that your implementation is correct and complete when the integration tests pass with the Kubernetes container manager. Note that you'll need to rebuild the Clipper docker containers locally before running the tests.

You can run them with

```
./bin/build_docker_images.sh
./bin/run_unittests -i
```

I've modified the integration tests to use the k8s container manager, but if you'd like to see how they run with just Docker, you can change [line 22](https://github.com/dcrankshaw/clipper/blob/clipper_admin_v2/integration-tests/test_utils.py#L22)  in `test_utils.py` to the following to run them with the Docker container manager.

```py
SERVICE = "docker"
```

feynmanliang:
I'll look at this today./

feynmanliang:
The commit history on the k8s work is lost (presumably because you did a merge against `ucbrise:develop` for 9416786). We could have kept that if this PR used `feynmanliang:k8s` as the base, but since some merge conflicts were resolved the additional commits don't cleanly cherrypick. I don't think it's worth the effort trying to recover them, so lets leave this as is.

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,87 @@
...
+    # It seems like in k8s, you might create a DeploymentService
+    # when a model is deployed, and then just add replicas to it
+    # when  calling `add_replica()`. In this case, the two methods
+    # would require different behavior.
+    @abc.abstractmethod
+    def deploy_model(self, name, version, input_type, repo):
+        return
+
+    @abc.abstractmethod
+    def add_replica(self, name, version, input_type, repo):
```
feynmanliang: Why not `set_number_replicas` to improve referential transparency and also support decrementing?

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,87 @@
...
+    def start_clipper(self):
+        return
+
+    # TODO(feynmanliang): Do we need a separate deploy_model method?
+    # It seems like in k8s, you might create a DeploymentService
+    # when a model is deployed, and then just add replicas to it
+    # when  calling `add_replica()`. In this case, the two methods
+    # would require different behavior.
+    @abc.abstractmethod
+    def deploy_model(self, name, version, input_type, repo):
```
feynmanliang: Not sure I understand the comment, but yes you can start a Deployment with 0 replicas and modify that later. However, you'll need this in the superclass because running on K8s requires first creating the Deployment to manage rolling updates/restarting pods/etc whereas when running locally there is no Deployment to create.

Also, in the old CM, this method registers the model with clipper (`self._publish_new_model`). I suspect you'll want to do that here as well (it's currently missing in the Docker CM).

clipper_admin_v2/clipper_admin/container_manager.py
feynmanliang: Also, why do we need anything more than just the `name` if we're operating on an already deployed model?

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,488 @@
...
+DEFAULT_LABEL = ["DEFAULT"]
+
+logger = logging.getLogger(__name__)
+
+
+class ClipperException(Exception):
+    pass
+
+
+def start_clipper(cm):
```
feynmanliang: Any reason why this isn't encapsulated in a class? Seems weird to have the end user maintain the reference to the `ContainerManager`, manage its lifecycle, and pass it into all these functions.

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ContainerManager(object):
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, clipper_public_hostname):
+        self.public_hostname = clipper_public_hostname
+
+    @abc.abstractmethod
+    def start_clipper(self):
```
feynmanliang: Seems strange that the `cm` is responsible for starting clipper, but `cm`'s constructor takes Clipper hostname as an argument (which isn't known until after Clipper is started).

I would've made`start_clipper` concrete and have it delegate to an abstract `_start_clipper` that is overridden by implementations of the abc which returns the hostname, then have `start_clipper` (this method) set `self.public_hostname` with the return.

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,488 @@
...
+        logger.info(e.msg)
+        return False
+
+
+def register_application(cm, name, input_type, default_output, slo_micros):
+    url = "http://{host}/admin/add_app".format(host=cm.get_admin_addr())
+    req_json = json.dumps({
+        "name": name,
+        "input_type": input_type,
+        "default_output": default_output,
```
feynmanliang: This missing a field

```
ClipperException: Received error status code: 400 and message: Error parsing JSON: JSON object does not have required key: candidate_model_names. Expected JSON schema:
  {
   "name" := string,
   "candidate_model_names" := [string],
   "default_output" := string,
   "latency_slo_micros" := int
  }
```

clipper_admin_v2/README.rst
```diff
@@ -0,0 +1,132 @@
...
+To file a bug or request a feature, please file a GitHub issue. Pull requests are welcome. Additional help and instructions
+for contributors can be found on our website at <http://clipper.ai/contributing>.
+
+## The Team
+
++ Dan Crankshaw (@dcrankshaw)
++ Joey Gonzalez (@jegonzal)
++ Corey Zumar (@Corey-Zumar)
++ Nishad Singh (@nishadsingh1)
++ Alexey Tumanov (@atumanov)
```
feynmanliang: Add me?

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,87 @@
+import abc
+
+# Constants
+
+CLIPPER_QUERY_PORT = 1337
```
feynmanliang: These must not be hardcoded for k8s compatibility, where there is no guarantee of the actual port bound by a service. Since we don't ever use these without having a reference to a CM, consider making this a property of a running CM instead

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,488 @@
...
+    else:
+        docker_client = docker.from_env()
+    logger.info("Building model Docker image at {}".format(model_data_path))
+    docker_client.images.build(
+            path=model_data_path,
+            tag=repo)
+
+    logger.info("Pushing model Docker image to {}".format(repo))
+    docker_client.images.push(repository=repo)
+    cm.deploy_model(name, version, input_type, repo, num_replicas=num_replicas)
```
feynmanliang: This fails because `num_replicas` is not part of CM's public interface (it is provided in the Docker CM). If we want to expose it, we should add it to the CM's public API.

clipper_admin_v2/clipper_admin/k8s/k8s_container_manager.py
```diff
@@ -0,0 +1,240 @@
...
+                 redis_ip=None,
+                 redis_port=6379,
+                 registry=None,
+                 registry_username=None,
+                 registry_password=None):
+        super(K8sContainerManager, self).__init__(clipper_public_hostname)
+        config.load_kube_config()
+        self._k8s_v1 = client.CoreV1Api()
+        self._k8s_beta = client.ExtensionsV1beta1Api()
+        self.redis_port = redis_port
```
feynmanliang: These are duplicated across both CM implementations. Since redis and a registry are both mandatory for a CM to do things (e.g. deploy docker images), this could be initialized in the superclass (via the `super` constructor and `**kwargs`).

feynmanliang:
I sent you a PR with the TODOs (and some of the comments ^^) at https://github.com/dcrankshaw/clipper/pull/3

dcrankshaw:
Awesome. Thanks for the comments. I know that the commit history was lost. I wanted to rebase this PR on the latest version of `develop` and there were a ton of merge conflicts that I had to keep addressing over and over. I made the choice to blow away commit history to make the rebase a lot cleaner. Sorry about any inconvenience.

clipper_admin_v2/clipper_admin/container_manager.py
dcrankshaw: Those are both good points. The reasons for both choices are historical.

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: You're right, that does seem weird. At first I was thinking that plain functions would be cleaner and you could potentially wrap this in a CLI, but given that there's still this container management state that needs to be tracked, we should create a class and track it for the user.

clipper_admin_v2/clipper_admin/container_manager.py
dcrankshaw: Got it. That makes sense.

feynmanliang:
Oops I think my PR had a problem: https://github.com/dcrankshaw/clipper/pull/3/files#diff-6272c59650820d22a2a09dfd5eba47f3L123 shouldn't have been removed

dcrankshaw:
Got it. I'll add it back in. I fixed a couple other problems before merging as well, so no worries.

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+DEFAULT_LABEL = ["DEFAULT"]
+
+logger = logging.getLogger(__name__)
+
+
+class ClipperException(Exception):
+    pass
+
+
+class ClipperConnection(object):
```
dcrankshaw: Can we come up with a better name than ClipperConnection?

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+            self.cm.start_clipper()
+            while True:
+                try:
+                    url = "http://{host}/metrics".format(host=self.cm.get_query_addr())
+                    requests.get(url, timeout=5)
+                    break
+                except RequestException as e:
+                    logger.info("Clipper still initializing.")
+                    time.sleep(1)
+            logger.info("Clipper is running")
```
dcrankshaw: Does `start_clipper` need to call `self.connect()`?

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+            return True
+        except ClipperException as e:
+            logger.info(e.msg)
+            return False
+
+    def connect(self):
```
dcrankshaw: Let's set a boolean here that prevents users from calling other methods if they haven't connected yet. This bool can be set in `start_clipper` and in `connect`, and then we can provide nice error messages.

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        logger.info(r.text)
+        if r.status_code != requests.codes.ok:
+            msg = "Received error status code: {code} and message: {msg}".format(
+                code=r.status_code, msg=r.text)
+            logger.error(msg)
+            raise ClipperException(msg)
+
+    def _resolve_docker_repo(self, image, registry):
```
dcrankshaw: this is out of date, I will be changing this.

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        if registry is None:
+            registry = self.cm.get_registry()
+        if registry is None:
+            repo = image
+        else:
+            repo = "{registry}/{image}".format(registry=registry, image=image)
+        return repo
+
+    def build_and_deploy_model(self,
```
dcrankshaw: Extract the `build` part from this function and make a public API method that let's users build a model container and push to a registry without deploying to Clipper.

dcrankshaw: Rename `build_and_deploy_model` to a shorter and cleaner name

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        image_name = "{name}:{version}".format(name=name, version=version)
+        repo = self._resolve_docker_repo(image_name, registry)
+        if "DOCKER_API_VERSION" in os.environ:
+            docker_client = docker.from_env(
+                version=os.environ["DOCKER_API_VERSION"])
+        else:
+            docker_client = docker.from_env()
+        logger.info(
+            "Building model Docker image at {}".format(model_data_path))
+        docker_client.images.build(path=model_data_path, tag=repo)
```
dcrankshaw: Delete the dockerfile we just wrote to keep filesystem clean

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+                     version,
+                     input_type,
+                     model_data_path,
+                     base_image,
+                     labels=None,
+                     registry=None,
+                     num_replicas=1):
+        version = str(version)
+
+        with open(model_data_path + '/Dockerfile', 'w') as f:
```
dcrankshaw: rename the Dockerfile to a random string to avoid overwriting existing dockerfiles and set the `dockerfile` argument in `docker_client.images.build`

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        logger.info("Pushing model Docker image to {}".format(repo))
+        docker_client.images.push(repository=repo)
+        self.deploy_model(name, version, input_type, repo, labels, num_replicas)
+
+
+    def deploy_model(self,
+                     name,
+                     version,
+                     input_type,
+                     repo,
```
dcrankshaw: rename `repo` to `image`

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        if version is None:
+            model_info = self.get_all_models(verbose=True)
+            for m in model_info:
+                if m["is_current_version"]:
+                    version = m["model_version"]
+        assert version is not None
+        version = str(version)
+        return self.cm.get_num_replicas(name, version)
+
+    def set_num_replicas(self, name, version=None, num_replicas=1):
```
dcrankshaw: Does this method name adequately convey that replicas will be added or removed (which is a fairly heavyweight operation)?

dcrankshaw: Remove the default value from `num_replicas` and force the user to be explicit

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+                current_version = m["model_version"]
+        assert current_version is not None
+        self.cm.stop_models(
+            model_name=model_name, keep_version=current_version)
+
+    def get_query_addr(self):
+        return self.cm.get_query_addr()
+
+    def stop_deployed_models(self):
+        self.cm.stop_models()
```
dcrankshaw: add a `def stop_models(self, names)` method that takes a list of model names and stops all versions of each of those models

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        model_name : str
+            The name of the model whose old containers you want to clean.
+        """
+        model_info = self.get_all_models(verbose=True)
+        current_version = None
+        for m in model_info:
+            if m["is_current_version"]:
+                current_version = m["model_version"]
+        assert current_version is not None
+        self.cm.stop_models(
```
dcrankshaw: Change the container manager `stop_models` method to take a dict of <model_name: list(model_version)> pairs. For each model. this will stop each version in that model's list.

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,86 @@
...
+    @abc.abstractmethod
+    def start_clipper(self):
+        return
+
+    @abc.abstractmethod
+    def connect(self):
+        return
+
+    @abc.abstractmethod
+    def deploy_model(self, name, version, input_type, repo):
```
dcrankshaw: rename repo to image

clipper_admin_v2/clipper_admin/container_manager.py
```diff
@@ -0,0 +1,86 @@
...
+            is ignored if model_name is empty. The typical use case for this argument is to remove
+            old versions of a model but keep the currently active version.
+        """
+        return
+
+    @abc.abstractmethod
+    def stop_clipper(self):
+        pass
+
+    def get_registry(self):
```
dcrankshaw: delete this method

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+    CLIPPER_INTERNAL_MANAGEMENT_PORT)
+
+DOCKER_NETWORK_NAME = "clipper_network"
+
+logger = logging.getLogger(__name__)
+
+
+class DockerContainerManager(ContainerManager):
+    def __init__(self,
+                 clipper_public_hostname,
```
dcrankshaw: This argument doesn't make sense, it should be set in `start_clipper` or provided in `connect()`

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                 clipper_query_port=1337,
+                 clipper_management_port=1338,
+                 clipper_rpc_port=7000,
+                 redis_ip=None,
+                 redis_port=6379,
+                 registry=None,
+                 registry_username=None,
+                 registry_password=None,
```
dcrankshaw: Delete registry related arguments

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class DockerContainerManager(ContainerManager):
+    def __init__(self,
+                 clipper_public_hostname,
+                 clipper_query_port=1337,
+                 clipper_management_port=1338,
+                 clipper_rpc_port=7000,
+                 redis_ip=None,
+                 redis_port=6379,
```
dcrankshaw: Move connection-related arguments to be set in `start_clipper` and `connect`

clipper_admin_v2/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,539 @@
...
+        else:
+            repo = "{registry}/{image}".format(registry=registry, image=image)
+        return repo
+
+    def build_and_deploy_model(self,
+                     name,
+                     version,
+                     input_type,
+                     model_data_path,
+                     base_image,
```
dcrankshaw: Add an additional metadata field that will be provided as a string environment variable to each container when run. This can be used to pass additional model metadata (such as object store addresses, etc) that need to be resolved at runtime and can't be put into the image.

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+            logger.info(
+                "{nw} network already exists".format(nw=DOCKER_NETWORK_NAME))
+        container_args = {
+            "network": DOCKER_NETWORK_NAME,
+            "labels": {
+                CLIPPER_DOCKER_LABEL: ""
+            },
+            "detach": True,
+        }
+        self.extra_container_kwargs.update(container_args)
```
dcrankshaw: This overwrites any labels the user provided. We should extend the labels list and overwrite the network.

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if self.redis_ip is None:
+            logger.info("Starting managed Redis instance in Docker")
+            self.redis_ip = "redis"
+            self.docker_client.containers.run(
+                'redis:alpine',
+                "redis-server --port %d" % self.redis_port,
```
dcrankshaw: This should always start redis on the default port, and just remap 6379 to the provided redis port in the `ports` argument.

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                name="redis",
+                ports={'%s/tcp' % self.redis_port: self.redis_port},
+                **self.extra_container_kwargs)
+
+        cmd = "--redis_ip={redis_ip} --redis_port={redis_port}".format(
+            redis_ip=self.redis_ip, redis_port=self.redis_port)
+        self.docker_client.containers.run(
+            'clipper/management_frontend:latest',
```
dcrankshaw: These should not pull the `latest` tagged versions from docker hub.

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+            cmd,
+            name=self.mgmt_frontend_name,
+            ports={
+                '%s/tcp' % CLIPPER_INTERNAL_MANAGEMENT_PORT:
+                self.clipper_management_port
+            },
+            **self.extra_container_kwargs)
+        self.docker_client.containers.run(
+            'clipper/query_frontend:latest',
```
dcrankshaw: change version tag

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def stop_models(self, model_name=None, keep_version=None):
+        containers = self.docker_client.containers.list(
+            filters={"label": CLIPPER_MODEL_CONTAINER_LABEL})
+        for c in containers:
+            c_name, c_version = c.labels[CLIPPER_MODEL_CONTAINER_LABEL].split(
+                ":")
+            if model_name is not None and model_name == c_name:
+                if keep_version is None or keep_version != c_version:
+                    c.stop()
```
dcrankshaw: Well I'm going to rewrite this function anyway, but this misses the case where model_name is None and we want to remove all model containers

clipper_admin_v2/clipper_admin/deployers/python.py
```diff
@@ -0,0 +1,131 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_endpoint(clipper_conn,
+                    name,
+                    input_type,
+                    func,
+                    default_output="None",
+                    version=1,
+                    slo_micros=100000,
```
dcrankshaw: Bump default slo_micros to 3 seconds

clipper_admin_v2/clipper_admin/deployers/python.py
```diff
@@ -0,0 +1,131 @@
...
+
+    clipper_conn.link_model_to_app(name, name)
+
+
+def deploy_python_closure(clipper_conn,
+                          name,
+                          version,
+                          input_type,
+                          func,
+                          base_image="clipper/python-closure-container",
```
dcrankshaw: This should have a version tag associated with it

clipper_admin_v2/clipper_admin/deployers/python.py
```diff
@@ -0,0 +1,131 @@
...
+            1,
+            centered_predict,
+            "doubles",
+            num_containers=1)
+    """
+
+    serialization_dir = save_python_function(name, func)
+    logger.info("Python closure saved")
+    # Deploy function
+    deploy_result = clipper_conn.deploy_model(name, version, input_type,
```
dcrankshaw: I don't think `deploy_model` returns anything anymore?

clipper_admin_v2/clipper_admin/deployers/pyspark.py
```diff
@@ -0,0 +1,151 @@
...
+
+def create_endpoint(clipper_conn,
+                    name,
+                    input_type,
+                    func,
+                    pyspark_model,
+                    sc,
+                    default_output="None",
+                    version=1,
+                    slo_micros=100000,
```
dcrankshaw: bump default slo

dcrankshaw (commented):
Comments from joint review with @Corey-Zumar

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: Addressed

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: I think it's fine

dcrankshaw: Done

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
dcrankshaw: In retrospect I don't think this comment makes sense. Any docker specific arguments should be provided to init, so that `start_clipper` and `connect` can cleanly inherit from the generic `ContainerManager` interface. I'm leaving this as is.

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
dcrankshaw: See above comment

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/deployers/python.py
dcrankshaw: Done

clipper_admin_v2/clipper_admin/deployers/pyspark.py
dcrankshaw: Done

Corey-Zumar:
@dcrankshaw Is this ready for review again?

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw:
@Corey-Zumar This is ready for a review. There are a couple documentation loose ends I need to wrap up (the quickstart in the README and the tutorial and examples), but other than that this should be ready.

shaneknapp:
test this please

shaneknapp:
test this please

PySparkContainerDockerfile
```diff
@@ -1,9 +1,7 @@
 ARG CODE_VERSION=0.2-rc1
```
Corey-Zumar: This should be `develop`, right?

PyClosureContainerDockerfile
```diff
@@ -0,0 +1,14 @@
+ARG CODE_VERSION=0.2-rc1
```
Corey-Zumar: `develop`

bin/run_unittests.sh
```diff
@@ -138,10 +138,13 @@ function run_integration_tests {
...
   fi
-  python ../integration-tests/clipper_manager_tests.py
+  python ../integration-tests/clipper_admin_tests.py
+  python ../integration-tests/many_apps_many_models.py 2 3
   python ../integration-tests/deploy_pyspark_models.py
   python ../integration-tests/deploy_pyspark_pipeline_models.py
-  python ../integration-tests/many_apps_many_models.py 2 3
+  python ../integration-tests/kubernetes_integration_test.py
+  # TODO: uncomment
+  # python ../integration-tests/deploy_R_models.py
```
Corey-Zumar: Assuming this was commented out due to impending R support?

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,256 @@
...
+                 docker_network="host",
+                 extra_container_kwargs={}):
+        """
+        Parameters
+        ----------
+        docker_ip_address : str, optional
+            The public hostname or IP address at which the Clipper Docker
+            containers can be accessed via their exposed ports. If running the containers
+            on a network other than "host", you may need to set this explicitly for ``ClipperConnection``
+            access the Clipper containers.
```
Corey-Zumar: Nit: Typo, should be `...you may need to set this explicitly for ``ClipperConnection`` **to** access the Clipper containers`

clipper_admin/setup.py
```diff
@@ -0,0 +1,38 @@
+from setuptools import setup
```
Corey-Zumar, line 1: `python setup.py install` fails due to the absence of a `requirements.txt` file (Python 2.7).

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,256 @@
...
+            if c_name in models and c_version in models[c_name]:
+                c.stop()
+
+    def stop_all_model_containers(self):
+        containers = self.docker_client.containers.list(
+            filters={"label": CLIPPER_MODEL_CONTAINER_LABEL})
+        for c in containers:
+            c.stop()
+
+    def stop_all(self):
```
Corey-Zumar, line 254: If `redis_ip` was initially `None`, we don't reset it when we stop all containers. As a result, subsequent calls to `start_clipper` fail to start redis.

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,256 @@
...
+
+        if self.redis_ip is None:
+            logger.info("Starting managed Redis instance in Docker")
+            redis_container = self.docker_client.containers.run(
+                'redis:alpine',
+                "redis-server --port %s" % self.redis_port,
+                ports={'%s/tcp' % self.redis_port: self.redis_port},
+                labels=self.common_labels,
+                **self.extra_container_kwargs)
+            self.redis_ip = redis_container.name
```
Corey-Zumar: In the case where we're using the docker host network, this should be `self.redis_ip = localhost`. Otherwise, the frontend containers fail to connect to redis.

Corey-Zumar: Also, this seems to be failing in general for other docker networks. The frontend containers cannot find redis using the container name as the ip address.

clipper_admin/docs/conf.py
```diff
@@ -42,13 +42,13 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
-import recommonmark
-from recommonmark.parser import CommonMarkParser
-from recommonmark.transform import AutoStructify
-
-source_parsers = {'.md': CommonMarkParser}
+# import recommonmark
```
Corey-Zumar, line 45: Remove commented code

clipper_admin/docs/conf.py
```diff
@@ -167,15 +168,23 @@
...
 def setup(app):
-    app.add_config_value('recommonmark_config', {
-        'url_resolver': lambda url: github_doc_root + url,
-        'auto_toc_tree_section': 'Contents',
-        'enable_eval_rst': True,
-        'enable_auto_doc_ref': True,
-    }, True)
-    app.add_transform(AutoStructify)
+    app.connect("autodoc-skip-member", skip)
+    # app.add_config_value('recommonmark_config', {
```
Corey-Zumar, line 184: Remove commented code

containers/python/python_closure_container.py
```diff
@@ -3,10 +3,9 @@
 import os
 import sys
 import numpy as np
-np.set_printoptions(threshold=np.nan)
 
-sys.path.append(os.path.abspath("/lib/"))
-from clipper_admin import cloudpickle
+# sys.path.append(os.path.abspath("/lib/"))
```
Corey-Zumar: Should this be removed or uncommented?

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -0,0 +1,256 @@
...
+        self.extra_container_kwargs = extra_container_kwargs
+
+        # Merge Clipper-specific labels with any user-provided labels
+        if "labels" in self.extra_container_kwargs:
+            self.common_labels = self.extra_container_kwargs.pop("labels")
+            self.common_labels.update({CLIPPER_DOCKER_LABEL: ""})
+        else:
+            self.common_labels = {CLIPPER_DOCKER_LABEL: ""}
+
+    def start_clipper(self, query_frontend_image, mgmt_frontend_image,
```
Corey-Zumar: In the case that the docker network is `host`, this fails to forward the proper ports. As a result, requests sent to `localhost:1337/metrics` by `ClipperConnection` fail and `ClipperConnection.start()` never returns.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,995 @@
...
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        logger.info(r.text)
+        if r.status_code != requests.codes.ok:
+            msg = "Received error status code: {code} and message: {msg}".format(
+                code=r.status_code, msg=r.text)
+            logger.error(msg)
+            raise ClipperException(msg)
+
+    def build_and_deploy_model(self,
```
Corey-Zumar: In addition to this method and `deploy_model`, I still think it's a good idea to have a `build_model` method defined within this module that performs only the first operation documented below. This will be useful for data scientists who want to deploy their models without access to (or a need for) a running Clipper instance.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,995 @@
...
+        :py:exc:`clipper.ClipperException`
+        """
+
+        if not self.connected:
+            raise UnconnectedException()
+        version = str(version)
+
+        _validate_versioned_model_name(name, version)
+
+        with open(model_data_path + '/Dockerfile', 'w') as f:
```
Corey-Zumar: As far as I can tell, couldn't we still potentially be overwriting a dockerfile here? There was a comment indicating that this was fixed, but I'm not sure:

Comment:
"rename the Dockerfile to a random string to avoid overwriting existing dockerfiles and set the `dockerfile` argument in `docker_client.images.build`"

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,995 @@
...
+        ------
+        :py:exc:`clipper.UnconnectedException`
+        """
+        if not self.connected:
+            raise UnconnectedException()
+        if version is None:
+            model_info = self.get_all_models(verbose=True)
+            for m in model_info:
+                if m["is_current_version"]:
+                    version = m["model_version"]
```
Corey-Zumar: Nit: We can insert a break after we've found the most recent version

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,995 @@
...
+        Raises
+        ------
+        :py:exc:`clipper.UnconnectedException`
+        """
+        if not self.connected:
+            raise UnconnectedException()
+        if version is None:
+            model_info = self.get_all_models(verbose=True)
+            for m in model_info:
+                if m["is_current_version"]:
```
Corey-Zumar: I think this needs to be filtered by the supplied model name. Otherwise, we're setting `version` to be the current version of the last model contained in the dictionary.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,995 @@
...
+        :py:exc:`clipper.UnconnectedException`
+        :py:exc:`clipper.ClipperException`
+        """
+        if not self.connected:
+            raise UnconnectedException()
+        if version is None:
+            model_info = self.get_all_models(verbose=True)
+            for m in model_info:
+                if m["is_current_version"]:
+                    version = m["model_version"]
```
Corey-Zumar: This logic needs to be changed in the same way as lines 453-459. We should define a `get_current_model_version(name)` method to minimize duplicate code.

Corey-Zumar (changes requested):
This is awesome - very intuitive API. There are a few functionality issues to address (see comments). 

Notes:

* Issues with DockerContainerManager are preventing full testing of ClipperConnection. I will test the full API once these issues have been addressed. 

* Unless the user looks at the source for logging behavior, they may not have any log output by default. Logging does not work until `logging.basicConfig()` is called. We should call it for the user in some fashion. 

* Resolve all outstanding TODOs or comment indicating that they’ll be addressed in a subsequent PR

PySparkContainerDockerfile
dcrankshaw: Fixed

PyClosureContainerDockerfile
dcrankshaw: Fixed

bin/run_unittests.sh
dcrankshaw: Decided to delete the weird R/Python container. I deleted the commented out line.

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Yeah, turns out setting `docker_network=host` causes some weirdness. I reverted back to creating our own network and we throw an exception if the user tries to set the network to "host". This should work now.

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: See above comment, "host" network is no longer supported.

dcrankshaw: I think this should all work now

clipper_admin/setup.py
dcrankshaw: Dependencies are specified in  `setup.py`. Run `pip install -e /path/to/clipper_admin` instead.

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Added an additional field `self.external_redis` that we never mutate outside the constructor:

```
        if redis_ip is None:
            self.external_redis = False
        else:
            self.external_redis = True
```

If `self.external_redis` is False when you call `start_clipper`, we will start a redis instance regardless of whether you've already started and stopped clipper.

clipper_admin/docs/conf.py
dcrankshaw: Given that this is just the config to build the docs, I'd actually rather leave that in if we want to add markdown support back, rather than having to go look it up again in the future.

clipper_admin/docs/conf.py
dcrankshaw: See above comment

containers/python/python_closure_container.py
dcrankshaw: Removed

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Yeah good point. Done.

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Yeah that didn't work with the Python docker SDK. I added a check and raise an exception if a Dockerfile already exists at that location.

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Oh yeah. Good catch.

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

dcrankshaw:
@Corey-Zumar addressed your review comments. I still have a few TODOs to address and I need to update the tutorial and quickstart, but other than that this is complete and ready for another round of reviews.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+            host=self.cm.get_admin_addr())
+        req_json = json.dumps({
+            "name": name,
+            "input_type": input_type,
+            "default_output": default_output,
+            "latency_slo_micros": slo_micros
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+        if r.status_code != requests.codes.ok:
```
Corey-Zumar: Else (if the status code equals `requests.codes.ok`), we should log the response text.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+            raise UnconnectedException()
+        version = str(version)
+        url = "http://{host}/admin/get_model".format(
+            host=self.cm.get_admin_addr())
+        req_json = json.dumps({"model_name": name, "model_version": version})
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+
+        if r.status_code == requests.codes.ok:
+            app_info = r.json()
```
Corey-Zumar: Nit: This should be `model_info`. Also, this method does not log any content in the case that the user attempts to get info corresponding to an unregistered model.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+            "model_version": version,
+            "replica_id": replica_id,
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+
+        if r.status_code == requests.codes.ok:
+            app_info = r.json()
+            if len(app_info) == 0:
+                return None
```
Corey-Zumar: Consider indicating to the user that no such replica exists.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+        req_json = json.dumps({
+            "model_name": name,
+            "model_version": version,
+            "replica_id": replica_id,
+        })
+        headers = {'Content-type': 'application/json'}
+        r = requests.post(url, headers=headers, data=req_json)
+
+        if r.status_code == requests.codes.ok:
+            app_info = r.json()
```
Corey-Zumar: Nit: `replica_info`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+            raise UnconnectedException()
+        model_info = self.get_all_models(verbose=True)
+        model_dict = {}
+        for m in model_info:
+            if m["model_name"] in model_names and not m["is_current_version"]:
+                if m["model_name"] in model_dict:
+                    model_dict[m["model_name"]].append(m["model_version"])
+                else:
+                    model_dict[m["model_name"]] = [m["model_version"]]
+        self.cm.stop_models(model_dict)
```
Corey-Zumar, line 1134: It would be nice to log the number and versions of models that were stopped.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    def stop_all_model_containers(self):
+        """Stops all model containers started via Clipper admin commands.
+
+        This method can be used to clean up leftover Clipper model containers even if the
+        Clipper management frontend or Redis has crashed. It can also be called without calling
+        ``connect`` first.
+        """
+        self.cm.stop_all_model_containers()
```
Corey-Zumar, line 1147: Log info about stopped containers or at least a success message.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    def stop_all(self):
+        """Stops all processes that were started via Clipper admin commands.
+
+        This includes the query and management frontend Docker containers and all model containers.
+        If you started Redis independently, this will not affect Redis. It can also be called without calling
+        ``connect`` first.
+        """
+        self.cm.stop_all()
```
Corey-Zumar, line 1157: Log a success message

clipper_admin/clipper_admin/deployers/python.py
```diff
@@ -0,0 +1,134 @@
...
+
+    clipper_conn.register_application(name, input_type, default_output,
+                                      slo_micros)
+    deploy_python_closure(clipper_conn, name, version, input_type, func,
+                          base_image, labels, registry, num_replicas)
+
+    clipper_conn.link_model_to_app(name, name)
+
+
+def deploy_python_closure(
```
Corey-Zumar, line 81: After invoking `deploy_python_closure(conn, "m7", "1", "ints", func) ` with a connection backed by the docker container manager, I encountered an error with the following back trace:

```
Traceback (most recent call last):
  File "/Users/Corey/Documents/RISE/clipper/clipper_admin/clipper_admin/deployers/check_and_write_deps.py", line 127, in <module>
    conda_dep_fname, pip_dep_fname):
  File "/Users/Corey/Documents/RISE/clipper/clipper_admin/clipper_admin/deployers/check_and_write_deps.py", line 91, in check_solvability_write_deps
    missing_packages = missing_packages_error.pkgs
AttributeError: 'ResolvePackageNotFound' object has no attribute 'pkgs'

17-08-30:02:04:13 INFO     [python.py:127] Python closure saved
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Corey/Documents/RISE/clipper/clipper_admin/clipper_admin/deployers/python.py", line 138, in deploy_python_closure
    force=True)
  File "/Users/Corey/Documents/RISE/clipper/clipper_admin/clipper_admin/clipper_admin.py", line 281, in build_and_deploy_model
    container_registry)
  File "/Users/Corey/Documents/RISE/clipper/clipper_admin/clipper_admin/clipper_admin.py", line 346, in build_model
    docker_file_path = os.path.join(model_data_path, "Dockerfile")
  File "/usr/local/bin/anaconda2/envs/clipper/lib/python2.7/posixpath.py", line 70, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'bool' object has no attribute 'endswith'
```

Corey-Zumar (changes requested):
Almost there! A few comments concerning method outputs. I'm of the opinion that all method calls issued from a `ClipperConnection` object should log some sort of output. 

Also, there seems to be an issue with deploying python functions (see comment).

clipper_admin/clipper_admin/deployers/python.py
dcrankshaw: This issue looks like a problem introduced by the latest version of conda. We got a bug report about this in #269.

What version of conda are you on?

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Good point. Fixed.

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Fixed

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Logged the model dict (with a pretty printer so it should be readable)

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Done

clipper_admin/clipper_admin/deployers/python.py
dcrankshaw: Can you try rolling back to a 4.3.* version of conda and running again. This is definitely a bug that we should fix, but isn't a new one introduced by this PR and shouldn't block this getting merged.

Corey-Zumar: I'm on a 4.3* version of conda: 4.3.23

Corey-Zumar: Ah, that seems to be the latest version. Will roll back and test with something earlier

Corey-Zumar (approved):
LGTM. Confirmed that the python closure deployment issue is only present for recent versions of anaconda. Will merge once this is rebased on develop and tests pass.

dcrankshaw:
Sweet. Actually hold off on merging until I get a chance to update the
quickstart and tutorial. Will do that in a few minutes.

dcrankshaw:
jenkins test this please

clipper_admin/clipper_admin/deployers/python.py
```diff
@@ -21,33 +21,53 @@ def create_endpoint(
...
+        the default output is returned. Therefore, it is recommended that
+        the SLO not be set aggressively low unless absolutely necessary.
+        100000 (100ms) is a good starting value, but the optimal latency objective
+        will vary depending on the application.
+    labels : list(str), optional
+        A list of strings annotating the model. These are ignored by Clipper
+        and used purely for user annotations.
+    registry : str, optional
+        The Docker container registry to push the freshly built model to. Note
+        that if you are running Clipper on Kubernetes, this registry must be accesible
```
Corey-Zumar: Nit: s\accesible\accessible\

clipper_admin/docs/index.rst
```diff
@@ -30,16 +30,106 @@ one for running Clipper locally directly on Docker, and one for running Clipper
...
+this module must take a list of inputs as the sole argument, and return a list of strings of
+exactly the same length. The reason the prediction function takes a list of inputs rather than
+a single input is to provide models the possibility of computing multiple predictions in parallel
+to improve model performance. For example, many models that run on a GPU can significantly improve
+throughput by batching predictions to better utilize the many parallel cores of the GPU.
+
+In addition, the function must only use pure Python code. More specifically, all of the state
+captured by the function will be pickled using `Cloudpickle <https://github.com/cloudpipe/cloudpickle>`_,
+so any state captured by the function must be able to be pickled. Most Python libraries that
+use C extensions create objects that cannot be pickled. This includes many common machine-learning
```
Corey-Zumar, line 59: this should be `...use C extensions **to** create objects that cannot be pickled...`

clipper_admin/docs/index.rst
```diff
@@ -30,16 +30,106 @@ one for running Clipper locally directly on Docker, and one for running Clipper
...
+frameworks including PySpark, TensorFlow, PyTorch, and Caffe. You will have to create your own
```
Corey-Zumar: s\including\such as\ (we just used "including" earlier in the sentence)

examples/tutorial/tutorial_part_two.ipynb
```diff
@@ -136,18 +158,21 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To use your newly deployed model to generate predictions, it needs to be linked to your Clipper application. Let Clipper know that your `\"cifar_demo\"` app should use the `\"birds_vs_planes_classifier\"` model to serve predictions.\n",
+    "To use your newly deployed model to generate predictions, it needs to be linked to your Clipper application. This tells Clipper to route requests for `\"cifar-demo\"` app to the `\"birds-vs-planes-classifier\"` model to make predictions.\n",
```
Corey-Zumar: s\\...to route requests for clipper demo app...\\...to route requests for **the** clipper demo app...\

Corey-Zumar (approved):
LGTM. Confirmed that quick start, the example client, and the tutorial work. Just fix a couple typos and I'll merge once tests pass.

clipper_admin/clipper_admin/deployers/python.py
dcrankshaw: Done

clipper_admin/docs/index.rst
dcrankshaw: I think it's correct as is. It's not the most elegant sentence in the world, but i'm leaving it.

clipper_admin/docs/index.rst
dcrankshaw: done

examples/tutorial/tutorial_part_two.ipynb
dcrankshaw: Done

PR #252, opened 2017-07-13
Pass by value where appropriate

nishadsingh1:
Makes benchmark script runnable; doesn't pass data in by reference to threads and then destroy it.


dcrankshaw (approved):
lgtm

PR #264, opened 2017-07-31
Fix docker image versioning

### LINKED ISSUE #244: Pin Docker image version dependencies to specific releases
Fix how Docker files depend on each other (for example, all of the model container Docker files are built with `FROM clipper/py-rpc:latest`, but really the 0.1 containers should be built from the 0.1 RPC base image, etc).

We should also fix which Docker image version the `clipper_admin` library uses, so that the 0.2.x releases use the 0.2 versions of the Docker images, but the version on develop uses the latest version.

dcrankshaw:
Resolves #244

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 #include the license file
 include LICENSE.txt
 include README.md
+include VERSION.txt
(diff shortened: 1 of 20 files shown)


Corey-Zumar (approved):
LGTM. Will merge when tests pass

dcrankshaw:
jenkins test this please

PR #266, opened 2017-08-08
[CLIPPER-224] Replace boost futures with folly futures

### LINKED ISSUE #224: On macOS query always returns the default prediction
This behavior has only been observed on macOS, so I think it's likely that the problem has to do with the way Clang compiles Boost on macs.

Steps to reproduce
1. Start Clipper
2. Register an application with type doubles and default output X and very large latency SLO.
3. Register an external model matching the name specified when registering the application
4. Connect the model to Clipper (I used the Java NoOpModel)
5. Send a prediction request with an array of doubles to Clipper
6. Observe that the response is the default output

From checking the logs, it is clear that both the model container gets queried and the response is returned in time. With further digging, I isolated the problem to `boost::future`s that were initially ready becoming "unready".

When a prediction is requested, the QueryProcessor schedules a task to evaluate the prediction, then waits for the future associated with the requested prediction to complete before the timer expires. Meanwhile, the task executor sends the prediction to a model container. When that prediction is returned from the container, it is placed in the prediction cache. This triggers the cache to [complete promises](https://github.com/ucbrise/clipper/blob/4278da241903fa02b45967a2ced0b3c87b4346a3/src/libclipper/src/task_executor.cpp#L66-L67) for any futures that have been returned, including the one that the QueryProcessor is waiting on. At this point, `boost::future::is_ready() == true`. The QueryProcessor is notified that the future is ready and proceeds to finish processing it and return a response. As part of that post-processing (due to the way that future composition works), QueryProcessor again calls [`future::is_ready`](https://github.com/ucbrise/clipper/blob/5a7d17251046436505ccf925d275f6ebd83c18dc/src/libclipper/src/query_processor.cpp#L116) to determine whether response processing was triggered due to the prediction task completing or the timer expiring. For some reason on macOS, this second call to `future::is_ready` returns false which causes the QueryProcessor to return the default response. I have not encountered this problem when running Clipper through our Docker containers, which use Alpine Linux, compile Clipper with GCC, and use the Alpine package repository's version of [Boost](https://pkgs.alpinelinux.org/package/edge/main/x86/boost-dev).

According to the documentation for [Boost Future](http://www.boost.org/doc/libs/1_64_0/doc/html/thread/synchronization.html#thread.synchronization.futures.reference.unique_future.is_ready), "The result of [future::is_ready] is not stable and that the future could become not ready even if the function returned true or vice-versa." It seems like the only way to fix this problem is to replace our use of Boost::future with an implementation that does not have this bug. Facebook's [Folly](https://github.com/facebook/folly/tree/master/folly/futures) library seems like the leading candidate. It supports future continuations and composition (which Clipper requires) and seems to be actively maintained by Facebook.

Corey-Zumar:
Fixes #224


cmake/FindBoost.cmake
```diff
@@ -1,3 +1,6 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
Corey-Zumar, line 1: This file was out of date and caused issues with Boost 1.64.0.

dcrankshaw:
Did you do any benchmarking before/after this code change? I'm curious if we see any performance difference?

Corey-Zumar:
Yes! There's a 40% improvement in throughput for noop_bench.

CMakeLists.txt
```diff
@@ -20,6 +20,15 @@ set_property(TARGET boost PROPERTY
 set_property(TARGET boost PROPERTY
     INTERFACE_LINK_LIBRARIES ${Boost_LIBRARIES})
 
+# Include Folly as an imported target
+find_package(Folly 2017.07.31.00 REQUIRED)
```
dcrankshaw: Can we pin to a less specific version of Folly? For example, the current version of Folly in homebrew is already a `2017.08.14.00`.

CMakeLists.txt
```diff
@@ -20,6 +20,15 @@ set_property(TARGET boost PROPERTY
...
+add_library(folly INTERFACE IMPORTED)
+set_property(TARGET folly PROPERTY
+    INTERFACE_INCLUDE_DIRECTORIES ${FOLLY_INCLUDE_DIR})
+set(FOLLY_LINK_FLAGS "-lglog")
```
dcrankshaw: What is this flag?

cmake/FindBoost.cmake
dcrankshaw: Sounds good. Did you just find a more recent version of the file?

src/benchmarks/src/end_to_end_bench.cpp
```diff
@@ -4,12 +4,11 @@
 #include <random>
 #include <vector>
 
-#include <boost/thread.hpp>
```
dcrankshaw, line 7: Can you remove all reference to `boost/thread`? I just grepped it, and there are still a bunch in the source code:

```
./src/libclipper/include/clipper/containers.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/datatypes.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/query_processor.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/task_executor.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/threadpool.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/timers.hpp:#include <boost/thread.hpp>
./src/libclipper/include/clipper/util.hpp:#include "boost/thread.hpp"
./src/libclipper/src/containers.cpp:#include <boost/thread.hpp>
./src/libclipper/src/metrics.cpp:#include <boost/thread.hpp>
./src/libclipper/src/persistent_state.cpp:#include <boost/thread.hpp>
./src/libclipper/src/query_processor.cpp:#include <boost/thread.hpp>
./src/libclipper/src/query_processor.cpp:#include <boost/thread/executors/basic_thread_pool.hpp>
./src/libclipper/src/rpc_service.cpp:#include <boost/thread.hpp>
./src/libclipper/src/task_executor.cpp:#include <boost/thread.hpp>
./src/libclipper/src/timers.cpp:#include <boost/thread.hpp>
./src/management/src/management_frontend.hpp:#include <boost/thread.hpp>
./src/management/src/management_frontend_tests.cpp:#include <boost/thread.hpp>
```

src/libclipper/src/query_processor.cpp
```diff
@@ -72,70 +74,74 @@ boost::future<Response> QueryProcessor::predict(Query query) {
...
 
-  std::tie(response_ready_future, all_tasks_completed, timer_future) =
-      future::when_either(std::move(all_tasks_completed),
-                          std::move(timer_future), completed_flag);
+  size_t num_tasks = task_futures.size();
 
-  boost::promise<Response> response_promise;
-  auto response_future = response_promise.get_future();
+  // TODO(czumar): Find a more readable name for std::vector<folly::Unit> and
+  // typedef accordingly
```
dcrankshaw: I don't think this TODO is necessary. It's not a particularly complicated type and I'd rather leave it explicit.

src/libclipper/src/query_processor.cpp
```diff
@@ -72,70 +74,74 @@ boost::future<Response> QueryProcessor::predict(Query query) {
...
-        try {
-          boost::rethrow_exception((*r).get_exception_ptr());
-        } catch (std::exception& e) {
+  std::shared_ptr<std::mutex> outputs_mutex = std::make_shared<std::mutex>();
+  std::vector<Output> outputs;
+  outputs.reserve(task_futures.size());
+  std::shared_ptr<std::vector<Output>> outputs_ptr =
+      std::make_shared<std::vector<Output>>(std::move(outputs));
+
+  std::vector<folly::Future<folly::Unit>> wrapped_task_futures;
```
dcrankshaw, line 97: Are you wrapping the futures for error-handling purposes? Just want to make sure I understand what's going on.

dcrankshaw (changes requested):
Just a couple minor comments. This looks great. It looks like once you got the cmake and docker stuff working, folly was basically a drop-in replacement for boost.

CMakeLists.txt
Corey-Zumar: This flag specifies that [glog](https://github.com/google/glog) should also be linked when linking Folly. Without this flag, I ran into a build error on OSX indicating that glog wasn't found.

src/libclipper/src/query_processor.cpp
Corey-Zumar: Agreed, removed the TODO

src/libclipper/src/query_processor.cpp
Corey-Zumar: I'm wrapping for error handling as well as aggregating future outputs in the vector owned by the shared pointer `outputs_ptr`. This aggregation is necessary because, in the case that the timer future completes before all of the task futures have completed, we need to access to the values of the completed tasks. `folly::collectAny` doesn't support this behavior directly, hence the wrapping.

CMakeLists.txt
Corey-Zumar: It turns out that this version pinning was completely ineffective because `FindFolly.cmake` doesn't set any version flags, nor does it appear that folly specifies an installation version in its include directory. For now, I've removed the version entirely.

src/benchmarks/src/end_to_end_bench.cpp
Corey-Zumar: Done!

Corey-Zumar (commented):

dcrankshaw (approved):
LGTM. Once we get Folly installed on the Jenkins cluster and the tests pass, I'll merge.

dcrankshaw:
@Corey-Zumar You need to update the ManagementFrontendDockerfile and any other docker files that build Clipper to install Folly. Optionally, you can create a single base docker image that installs all dependencies needed to build `libclipper` and then use that as a base image to build the rest of the docker images.

Corey-Zumar:
@dcrankshaw Updated the dockerfiles by creating a base image, as you suggested, called `clipper/lib_base` (built with `ClipperLibBaseDockerfile`).

ClipperLibBaseDockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
+
+RUN apk add --no-cache git bash make boost-dev cmake libev-dev hiredis-dev zeromq-dev \
+    libtool automake autoconf-archive@testing lz4-dev xz-dev snappy-dev zlib-dev \
+    glog-dev@testing scons openssl-dev libevent-dev
+
+# Set up Folly
+
+## Install DoubleConversion
+RUN git clone https://github.com/google/double-conversion \
```
dcrankshaw, line 12: Let's pin this to a specific release

ClipperLibBaseDockerfile
```diff
@@ -0,0 +1,36 @@
...
+    && make \
+    && make install \
+    && ln -s src double-conversion
+
+ENV DOUBLE_CONVERSION_HOME=/double-conversion/double-conversion/
+
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$DOUBLE_CONVERSION_HOME
+
+## Install GFLags
+RUN git clone https://github.com/gflags/gflags \
```
dcrankshaw, line 25: Pin this to a specific release

ClipperLibBaseDockerfile
```diff
@@ -0,0 +1,36 @@
...
+    && mkdir gflags/build \
+    && cd gflags/build \
+    && cmake -DBUILD_SHARED_LIBS=ON .. \
+    && make install
+
+## Install Folly
+RUN git clone https://github.com/facebook/folly \
```
dcrankshaw, line 34: Pin this to a specific release

dcrankshaw (changes requested):
Added a couple comments, but also the docker files aren't building folly for some reason.

Note that I updated FindFolly.cmake so make sure to pull and not force push.

ClipperLibBaseDockerfile
Corey-Zumar: Done

ClipperLibBaseDockerfile
Corey-Zumar: Done

ClipperLibBaseDockerfile
Corey-Zumar: Done

Corey-Zumar:
@dcrankshaw Addressed your comments. Version pinning fixed the docker build issues - thanks for the catch!

dcrankshaw:
jenkins test this please

Corey-Zumar:
jenkins test this please

Corey-Zumar:
jenkins test this please

ManagementFrontendDockerfile
```diff
@@ -1,4 +1,5 @@
-FROM frolvlad/alpine-gxx
+ARG CODE_VERSION
```
dcrankshaw: Give this arg a default version (eg https://github.com/ucbrise/clipper/blob/develop/PythonContainerDockerfile#L1), otherwise the automated builds on Docker Hub will fail.

QueryFrontendDockerfile
```diff
@@ -1,9 +1,11 @@
-FROM frolvlad/alpine-gxx
+ARG CODE_VERSION
```
dcrankshaw: Give this a default argument (see above comment)

dcrankshaw (changes requested):
LGTM. Just need to give the `ARG CODE_VERSION` in the dockerfiles a default value.

Corey-Zumar:
@dcrankshaw Added default versions to management frontend and query frontend docker files.

Corey-Zumar:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw (approved):

PR #268, opened 2017-08-14
[CLIPPER-212] Add CLOCK cache eviction policy

### LINKED ISSUE #212: add a cache eviction policy
Currently the prediction cache in Clipper can grow without bound. This is basically a memory leak an we should implement cache eviction. This involves determining how to safely delete cache entries.

Corey-Zumar:
Fixes #212

The cache eviction policy is a slightly-modified version of the clock algorithm. Notably, we don't evict any entries from the cache that have incomplete promises. Additionally, multiple evictions may occur for a single insertion due to the heterogeneity of entry sizes.


src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -24,6 +24,7 @@
 namespace clipper {
 
 const std::string LOGGING_TAG_TASK_EXECUTOR = "TASKEXECUTOR";
+constexpr size_t CACHE_SIZE_ELEMENTS = 2048;
```
Corey-Zumar: Because cache entries are of variable size and the user is unaware of the size of entry metadata, I think this should be a configurable value named `CACHE_SIZE_BYTES` instead. However, making this change complicates cache eviction substantially and will reduce CLOCK performance. Let's discuss @dcrankshaw.

Corey-Zumar: Actually, making this change shouldn't be too bad. Adding the "in progress" label instead.

Corey-Zumar: Done!

Corey-Zumar:
@dcrankshaw This is now ready for review.

Corey-Zumar:
jenkins test this please

src/frontends/src/query_frontend_main.cpp
```diff
@@ -11,18 +11,21 @@ int main(int argc, char* argv[]) {
   // clang-format off
   options.add_options()
     ("redis_ip", "Redis address",
-        cxxopts::value<std::string>()->default_value("localhost"))
+        cxxopts::value<std::string>()->default_value(clipper::DEFAULT_REDIS_ADDRESS))
     ("redis_port", "Redis port",
-        cxxopts::value<int>()->default_value("6379"))
+        cxxopts::value<int>()->default_value(std::to_string(clipper::DEFAULT_REDIS_PORT)))
     ("threadpool_size", "Number of threads for the task execution threadpool",
```
dcrankshaw: You can actually delete this command line arg and corresponding config field entirely. It's no longer used. Instead we create a thread in the task execution threadpool for each connected container.

src/frontends/src/query_frontend_main.cpp
```diff
@@ -11,18 +11,21 @@ int main(int argc, char* argv[]) {
...
-        cxxopts::value<int>()->default_value("4"));
+        cxxopts::value<int>()->default_value(std::to_string(clipper::DEFAULT_TASK_EXECUTION_THREADPOOL_SIZE)))
+    ("prediction_cache_size", "Size of the prediction cache in bytes, excluding cache metadata",
+       cxxopts::value<long>()->default_value(std::to_string(clipper::DEFAULT_PREDICTION_CACHE_SIZE_BYTES)));
```
dcrankshaw: There should a way to set this value from clipper_admin. Add an optional arg to set this in the old version of clipper_admin and I'll make sure it gets ported to v2.

src/libclipper/include/clipper/task_executor.hpp
dcrankshaw: Now that you changed the cache size to be measured in bytes, should this variable be deleted?

src/libclipper/src/task_executor.cpp
```diff
@@ -59,21 +62,68 @@ void PredictionCache::put(const VersionedModelId &model,
...
-    cache_.insert(std::make_pair(key, std::move(new_entry)));
+    insert_entry(key, new_entry);
+  }
+}
+
+void PredictionCache::insert_entry(const long key, CacheEntry &value) {
+  size_t entry_size_bytes = value.completed_ ? value.value_.y_hat_.size() : 0;
+  if (entry_size_bytes <= max_size_bytes_) {
+    evict_entries(size_bytes_ + entry_size_bytes - max_size_bytes_);
+    page_buffer_.insert(page_buffer_.begin() + page_buffer_index_, key);
```
dcrankshaw, line 92: What is the page buffer?

src/libclipper/src/task_executor.cpp
```diff
@@ -59,21 +62,68 @@ void PredictionCache::put(const VersionedModelId &model,
...
+                        entry_size_bytes, max_size_bytes_);
+  }
+}
+
+void PredictionCache::evict_entries(long space_needed_bytes) {
+  if (space_needed_bytes <= 0) {
+    return;
+  }
+  while (space_needed_bytes > 0) {
+    long page_key = page_buffer_[page_buffer_index_];
```
dcrankshaw: What is the role of the page_buffer and page_buffer_index here?

dcrankshaw (changes requested):
Overall looks good. I added a comment asking for clarification about how eviction is implemented.

Do you think this will impose a large overhead on workloads with a low cache hit rate? Should we think about adding a cache eviction policy that very cheaply evicts any entries that have already been fetched? We would need to ensure that an entry is only ever fetched from the cache once-per-query.

src/frontends/src/query_frontend_main.cpp
Corey-Zumar: Done!

src/frontends/src/query_frontend_main.cpp
Corey-Zumar: Done: added a `cache_size` argument to the constructor of `clipper_manager.Clipper`

src/libclipper/src/task_executor.cpp
Corey-Zumar: The `page_buffer_` is a vector of input hashes and acts as the clock in the cache eviction algorithm. In conjunction with the buffer (clock) itself, `page_buffer_index_` acts as the clock hand. To examine a page buffer entry with key `K`, we query the map `entries_` with `K` as the input hash key. We then determine whether or not the entry has all of its futures completed (via `entry.completed_`) and whether or not it was recently used (via the `entry.used_` boolean flag). If the entry is completed and was not recently used, we remove the key `K` from the page buffer, and the entry itself is removed from the `entries_` map.

src/libclipper/src/task_executor.cpp
Corey-Zumar: See the above comment. We use `page_buffer_` and `page_buffer_index_` to evict as many pages as necessary to obtain the required free space.

Corey-Zumar:
@dcrankshaw Addressed your review comments. Regarding applications with low cache hit rates, the user can avoid eviction overhead by setting the cache size to zero, effectively disabling caching.

dcrankshaw (approved):
LGTM

Corey-Zumar:
jenkins test this please

PR #274, opened 2017-08-30
Make unit tests run in a Docker container

dcrankshaw:
This adds a Docker container that can run all of our unit tests. In order to make this work, I also had to switch the base image for the query and management frontends from alpine to ubuntu. Other than increasing image size, this should not be a noticeable change.

### DIFF
--- a/ClipperLibBaseDockerfile
+++ b/ClipperLibBaseDockerfile
@@ -1,8 +1,5 @@
 FROM ubuntu:16.04
 
-# TODO: add this back
-# ENV DEBIAN_FRONTEND noninteractive
-
 RUN apt-get update && apt-get install -y g++ automake autoconf autoconf-archive libtool libboost-all-dev \
     libevent-dev libdouble-conversion-dev libgoogle-glog-dev libgflags-dev liblz4-dev \
     liblzma-dev libsnappy-dev make zlib1g-dev binutils-dev libjemalloc-dev libssl-dev \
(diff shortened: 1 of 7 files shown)


dcrankshaw:
You can run the unit tests with the following command:

```
docker run -it --rm --network=host -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp clipper/unittests:develop
```

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

ClipperLibBaseDockerfile
```diff
@@ -1,40 +1,17 @@
-FROM frolvlad/alpine-gxx
+FROM ubuntu:16.04
```
Corey-Zumar, line 1: From an ease of development standpoint, this change of base image is a plus. Any concerns about the difference in image size and overhead?

Corey-Zumar (approved):
LGTM.

ClipperLibBaseDockerfile
dcrankshaw: Not at this point. We can always build a minimal size docker image on alpine in the future. But that doesn't seem super important right now.

Corey-Zumar: Imo there shouldn't really be any since this is a base file used for frontend containers that we won't be replicating as frequently as model containers. Merging...

PR #279, opened 2017-09-01
Remove default label placeholder when no label is provided

### LINKED ISSUE #208: deploy_model with an empty label list causes an error
If you supply `deploy_model` with an argument `labels` that's an empty list, `add_container` gets an index out of bounds exception when iterating through the result of the redis query given by:
`redis-cli -h {host} -p 6379 -n {db} hgetall {key}`.

dcrankshaw:
I think we had already fixed #208 so this removes the now unnecessary placeholder label.


Corey-Zumar (approved):
LGTM

PR #280, opened 2017-09-01
Build new docker images in build_model with in-memory Dockerfile

dcrankshaw:
Removes the need to write a Dockerfile to disk in the user's file system by creating an in-memory Dockefile with StringIO then adding it directly to a temporary tar file to be used as the Docker build context.

The end result is that we don't write (and potentially overwrite) Dockerfile's in the user's system.

### DIFF
--- a/clipper_admin/clipper_admin/clipper_admin.py
+++ b/clipper_admin/clipper_admin/clipper_admin.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 import logging
 import docker
+import tempfile
 import requests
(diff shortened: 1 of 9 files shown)


dcrankshaw:
jenkins test this please

Corey-Zumar (changes requested):
Looks good! Part two of the tutorial needs to be updated to remove the `force` argument from `build_and_deploy_model`. Once that's done, I'll merge.

dcrankshaw:
Good catch. Fixed.

Corey-Zumar (approved):
LGTM

PR #283, opened 2017-09-02
Update docker build script to optionally publish images to DockerHub

### LINKED ISSUE #275: Move DockerHub automated builds to be built by Jenkins
Currently, all of our public Docker images are built using Docker Hub automated builds. However, the automated builds are fairly restrictive and hard to work with. We should switch to building the images with Jenkins as a post-commit hook on `develop` and each release branch and just directly pushing to Docker hub.

cc @shaneknapp

dcrankshaw:
This script cleans up the `bin/build_docker_images.sh` script and adds the ability to publish the images to DockerHub. The interface is the same, so it shouldn't break our existing Jenkins setup.

This script creates two tags for each image:
1. The git SHA.
2. The version (read from VERSION.txt)

With the publish option, both tags will be pushed to Dockerhub.

Closes #275.

## TODOs after the PR is merged:
+ [ ] Deactivate the automated builds on DockerHub and run this script with the `--publish` option as a post-commit job in Jenkins.
+ [ ] Remove the separate ./bin/build_docke_images` in Jenkins from the Clipper and Clipper-PRB jobs. This PR adds that call to the `./bin/run_ci.sh` script, so we don't need to call it separately in Jenkins.


bin/build_docker_images.sh
```diff
@@ -10,39 +18,102 @@ unset CDPATH
...
+        docker push $namespace/$image:$sha_tag
+        echo "Publishing $namespace/$image:$version_tag"
+        docker push $namespace/$image:$version_tag
+    fi
+}
+
+
+# Build the Clipper Docker images.
+build_images () {
+    # True and false indicate whether or not to public the image. We public public images.
```
Corey-Zumar: Nit: Can we use an alternative to the verb usage of `public`? "make the image public" is easier to understand.

dockerfiles/ClipperTestsDockerfile
```diff
@@ -1,4 +1,4 @@
-ARG CODE_VERSION=develop
```
Corey-Zumar, line 1: I'm guessing that we're removing these default arguments because the images build script, in conjunction with `VERSION.txt` are intended to be the point of control for image versioning?

dockerfiles/SparkScalaContainerDockerfile
```diff
@@ -1,3 +1,5 @@
+# This ARG isn't used but prevents warnings in the build script
```
Corey-Zumar, line 1: Should we also add this note to `RPCDockerfile` and `ClipperLibBaseDockerfile` ?

bin/build_docker_images.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+# This script builds all the public Docker images that the Clipper project distributes
+# and pushes them to Docker hub.
+# Each image is pushed with the tag <image_name>:version
+# In addition, if the version is "develop", each image will also be
```
Corey-Zumar: I don't think this is accurate. We're always tagging images with a git hash, regardless of version. What's the reasoning behind only tagging of the version is "develop"?

Corey-Zumar (changes requested):

bin/build_docker_images.sh
dcrankshaw: Oops. That's the result of a botched find/replace. Should read "True and false indicate whether or not to *publish* the image. We *publish* public images."

bin/build_docker_images.sh
dcrankshaw: Yeah I wrote the comment before I wrote the script. Will update the wording to be consistent.

dockerfiles/ClipperTestsDockerfile
dcrankshaw: Exactly. And removing the default makes sure that the build fails if we forget a build arg instead of silently building with "develop"

dockerfiles/SparkScalaContainerDockerfile
dcrankshaw: Done.

Corey-Zumar (approved):
LGTM

PR #294, opened 2017-09-13
Batch size estimation fix

Corey-Zumar:
This fixes a bug in our batch size calculations. Previously, when we received an RPC response for a batch of **N** prediction requests with a total latency of **L**, we were performing **N** invocations of `processing_container->update_batch_size(1, L)`. This was telling the batch size estimator that there are **N** new measurements which each indicate that it takes **L** time to process a batch consisting of one message. Instead, we are now correctly telling the estimator that there is 1 new measurement that indicates that it takes **L** time to process a batch consisting of **N** messages. This is done via a single call to `processing_container->update_batch_size(N, L)`.

### DIFF
--- a/src/libclipper/include/clipper/task_executor.hpp
+++ b/src/libclipper/include/clipper/task_executor.hpp
@@ -477,6 +477,10 @@ class TaskExecutor {
           std::chrono::duration_cast<std::chrono::microseconds>(task_latency)
               .count();
 
+      // Because an RPCResponse is guaranteed to contain data received from
+      // a single model container, the processing container for the first
+      // InflightMessage in the batch is the same processing container
+      // for all InflightMessage objects in the batch
       std::shared_ptr<ModelContainer> processing_container =
           active_containers_->get_model_replica(cur_model, cur_replica_id);
 
--- a/src/libclipper/include/clipper/task_executor.hpp
+++ b/src/libclipper/include/clipper/task_executor.hpp
@@ -498,7 +498,8 @@ class TaskExecutor {
       for (int batch_num = 0; batch_num < batch_size; ++batch_num) {
         InflightMessage completed_msg = keys[batch_num];
         cache_->put(completed_msg.model_, completed_msg.query_id_,
-                    Output{parsed_response.outputs_[batch_num], {completed_msg.model_}});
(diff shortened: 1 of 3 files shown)


src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -469,7 +469,24 @@ class TaskExecutor {
...
+
+      // Because an RPCResponse is guaranteed to contain data received from
+      // a single model container, the processing container for the first
+      // InflightMessage in the batch is the same processing container
+      // for all InflightMessage objects in the batch
+      std::shared_ptr<ModelContainer> processing_container =
+          active_containers_->get_model_replica(cur_model, cur_replica_id);
+
+      processing_container->update_throughput(batch_size, task_latency_micros);
+      processing_container->latency_hist_.insert(task_latency_micros);
```
dcrankshaw: As it turns out, this is a duplicate call. We already insert the message latency into the containers latency histogram in [`ModelContainer::update_throughput()`](https://github.com/ucbrise/clipper/blob/956f48746d9598fd27c2524a2c01642ea7a024e5/src/libclipper/src/containers.cpp#L46).

Delete this line.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -479,42 +496,17 @@ class TaskExecutor {
         (*cur_model_metric).throughput_->mark(batch_size);
         (*cur_model_metric).num_predictions_->increment(batch_size);
         (*cur_model_metric).batch_size_->insert(batch_size);
+        (*cur_model_metric)
+            .latency_->insert(static_cast<int64_t>(task_latency_micros));
```
dcrankshaw: Move this call inside the for loop (line 502). The model latency histogram should have an entry for every prediction request.

dcrankshaw (changes requested):
Nice catch. Just a couple quick fixes.

src/libclipper/include/clipper/task_executor.hpp

src/libclipper/include/clipper/task_executor.hpp

Corey-Zumar (commented):
@dcrankshaw addressed your comments

dcrankshaw (approved):
lgtm

PR #298, opened 2017-10-03
Update docker publish script to publish minor version tags

dcrankshaw:
Updates the Docker build script so that if we're publishing containers on a release branch (anything that matches "release-*", we'll publish Docker images with the exact version in VERSION.txt and the MAJOR.MINOR version. For example, for release 0.2.3, Docker images with the tag 0.2.3 and the tag 0.2 will be published.

### DIFF
--- a/bin/build_docker_images.sh
+++ b/bin/build_docker_images.sh
@@ -8,6 +8,12 @@
 # + The current version as read from VERSION.txt: <image_name>:version
 # For the images that we publish, both tags will be pushed.
 
+# In addition, if we are on a release branch (one that matches the regex "release-*")
+# and the version in VERSION.txt is not a release candidate and matches the form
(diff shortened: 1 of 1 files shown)


bin/build_docker_images.sh
```diff
@@ -27,6 +27,186 @@ cd $CLIPPER_ROOT
...
+    local LT=$?
+
+    if [ $EQ -ne 0 ] && [ $LT -ne 0 ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+# if [ "___semver.sh" == "___`basename $0`" ]; then
```
Corey-Zumar: Is this code useful to preserve?

Corey-Zumar (approved):
LGTM

bin/build_docker_images.sh
dcrankshaw: It is not. Removed.

PR #313, opened 2017-11-08
Support generic Tensorflow Models

### LINKED ISSUE #240: Out of the box support for deploying TensorFlow models
Support deploying TensorFlow models directly from the `clipper_admin` tool, similarly to the way we support deploying PySpark models directly by calling `Clipper.deploy_pyspark_model`.

santi81:
Support deploying TensorFlow models directly from the clipper_admin tool, similarly to the way we support deploying PySpark models directly by calling Clipper.deploy_pyspark_model.

Closes #240

### DIFF
--- a/clipper_admin/clipper_admin/docker/docker_container_manager.py
+++ b/clipper_admin/clipper_admin/docker/docker_container_manager.py
@@ -10,6 +10,7 @@ from ..container_manager import (
     CLIPPER_MGMT_FRONTEND_CONTAINER_LABEL, CLIPPER_INTERNAL_RPC_PORT,
     CLIPPER_INTERNAL_QUERY_PORT, CLIPPER_INTERNAL_MANAGEMENT_PORT)
 from ..exceptions import ClipperException
+from requests.exceptions import ConnectionError
 
 logger = logging.getLogger(__name__)
 
@@ -91,6 +92,9 @@ class DockerContainerManager(ContainerManager):
         except docker.errors.APIError as e:
             logger.debug(
                 "{nw} network already exists".format(nw=self.docker_network))
+        except ConnectionError:
+            msg = "Unable to connect to Docker. Please Check if Docker is running."
+            raise ClipperException(msg)
 
         if not self.external_redis:
             logger.info("Starting managed Redis instance in Docker")
--- a/clipper_admin/clipper_admin/exceptions.py
+++ b/clipper_admin/clipper_admin/exceptions.py
@@ -1,7 +1,9 @@
 class ClipperException(Exception):
     """A generic exception indicating that Clipper encountered a problem."""
 
-    pass
+    def __init__(self, msg, *args):
+        self.msg = msg
+        super(Exception, self).__init__(msg, *args)
 
 
 class UnconnectedException(ClipperException):


dcrankshaw:
jenkins test this please

santi81:
jenkins test this please

dcrankshaw:
jenkins ok to test

@santi81 Only whitelisted users can tell Jenkins to test. The "ok to test" directive means that Jenkins will automatically test any subsequent changes to this PR. I'll review this later tonight.

dcrankshaw:
jenkins ok to test

santi81:
@dcrankshaw There seems to be an issue with the build server.The builds fail immediately with the error: java.io.IOException: Failed to mkdirs: /home/jenkins/workspace/Clipper-PRB.

Could you please have a look?

dcrankshaw:
@santi81 Yeah we're running into that with the Jenkins master build as well. Looks like it's a Jenkins issue.

containers/python/tf_container.py
```diff
@@ -0,0 +1,94 @@
...
+        sys.path.append(os.path.abspath(modules_folder_path))
+        predict_fname = "func.pkl"
+        predict_path = "{dir}/{predict_fname}".format(
+            dir=path, predict_fname=predict_fname)
+        self.predict_func = load_predict_func(predict_path)
+        self.sess = tf.Session('', tf.Graph())
+        metagraph_path = os.path.join(path, "tfmodel/model.ckpt.meta")
+        checkpoint_path = os.path.join(path, "tfmodel/model.ckpt")
+        with self.sess.graph.as_default():
+            saver = tf.train.import_meta_graph(metagraph_path)
```
dcrankshaw: What do you think about clearing the devices when you import the meta graph? This will ensure that the model runs if GPU isn't available (or a different number of GPUs are available). But I think it will prevent the model from using GPUs if they are available.

```
saver = tf.train.import_meta_graph(metagraph_path, clear_devices=True)
```

@santi81 Do you have thoughts on the best way to support GPUs if they are available, but not crash if they aren't? @Corey-Zumar has looked at this before too. Thoughts?

dockerfiles/ClipperTestsDockerfile
```diff
@@ -17,12 +17,12 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh \
...
 RUN conda install -y libgcc pyzmq
  
-RUN pip install requests subprocess32 scikit-learn numpy pyyaml docker kubernetes pyspark
+RUN pip install requests subprocess32 scikit-learn numpy pyyaml docker kubernetes pyspark tensorflow
 
 # Install maven
 ARG MAVEN_VERSION=3.5.0
 ARG SHA=beb91419245395bd69a4a6edad5ca3ec1a8b64e41457672dc687c173a495f034
-ARG BASE_URL=https://apache.osuosl.org/maven/maven-3/${MAVEN_VERSION}/binaries
+ARG BASE_URL=https://archive.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries
```
dcrankshaw: 👍

containers/python/tf_container.py
dcrankshaw: On further thought, let's have this PR only support deploying tensorflow models to CPUs. We can follow it up with a separate PR that will support deploying to GPUs. The difficulty here is that in order for the container to support GPUs, it needs to have `tensorflow-gpu` installed. But this depends on CUDA, which means that if the user attempts to start the container without CUDA available (i.e. without using nvidia-docker to run the container), importing TensorFlow will fail and the container will crash.

@santi81 can you add a comment to the documentation in `deployers/tensorflow.py` that says that this does not support GPU containers yet? And definitely include the `clear_devices=True` flag as mentioned in the above comment.

integration-tests/deploy_tensorflow_models.py
```diff
@@ -0,0 +1,200 @@
...
+    transformedlabels = [objective(ele, pos_label) for ele in labels]
+    return (records, transformedlabels)
+
+
+def reset_vars():
+    sess.run(tf.global_variables_initializer())
+
+
+def reset_tf():
+    global sess
```
dcrankshaw: Can you make `sess` a local variable that gets passed around instead of a global?

dcrankshaw (changes requested):
This looks great! Thanks for taking this on. Just a few minor comments.

Also, make sure that you format the code to make sure it passes the format checker. From the root Clipper directory, run `./bin/format_code.sh`.

dcrankshaw:
Also, this PR closes #240

containers/python/tf_container.py
Corey-Zumar: For future reference, replacing line 27 with `with tf.device("/gpu:0"):` and specifying `allow_soft_placement=True` as a configuration option for the tensorflow session will correctly place GPU-compatible tensor operations on the first visible GPU. If no GPUs are visible, soft placement will ensure that they are placed on a CPU instead.

dcrankshaw: Oh great. Let's do that then.

dcrankshaw:
jenkins test this please

integration-tests/deploy_tensorflow_models.py
santi81: Done

santi81:
@dcrankshaw Can you please review the changes

containers/python/tf_container.py
```diff
@@ -0,0 +1,100 @@
...
+            '',
+            tf.Graph(),
+            config=tf.ConfigProto(
+                allow_soft_placement=True, log_device_placement=True))
+        metagraph_path = os.path.join(path, "tfmodel/model.ckpt.meta")
+        checkpoint_path = os.path.join(path, "tfmodel/model.ckpt")
+        with tf.device("/gpu:0"):
+            with self.sess.graph.as_default():
+                saver = tf.train.import_meta_graph(
+                    metagraph_path, clear_devices=False)
```
dcrankshaw: This should be `clear_devices=True`

dcrankshaw (changes requested):
Looks great! One minor fix then this is good to go.

dcrankshaw:
Did you mean to close this?

santi81:
that was a mistake ...sorry ..

dcrankshaw:
No problem 😃

santi81:
Pushed from a wrong branch and it closed the PR. I created a new one.I am sorry for the mess

dcrankshaw:
Got it. We'll switch over to the new PR.

PR #316, opened 2017-11-13
Support InternalIP address for K8S nodes with no IP address of type ExternalDNS

santi81:
Currently when using on premise K8S Cluster, the connect method throws an exception if none of the nodes in the cluster has an IP address of type ExternalDNS.It should be possible via an optional parameter to the KubernetesContainerManager to allow the usage of an address of type  InternalIP.


dcrankshaw:
jenkins ok to test

dcrankshaw (approved):
This is great! It's hard to anticipate how everyone runs their Kubernetes cluster, so this type of PR is especially useful for us.

PR #318, opened 2017-11-16
Fixed python deployer image path on Windows

es1024:
Fixes an issue caused by Windows using backslash path delimiters, which are used in os.path.join, and Docker expecting POSIX paths instead, resulting in copy operations failing when deploying a python closure from Windows.


dcrankshaw:
jenkins ok to test

dcrankshaw (approved):
lgtm. Will merge once the tests pass.

PR #321, opened 2017-11-20
Batch predict

tatevikstep:
Added support for input of type: `{ "input_batch" := [[double] | [int] | [byte] | [float] | string] }`

Source code changes:
 - `query_frontend.hpp`
 - `json_util.cpp/hpp`
 - `query_frontend_tests.cpp`

Added testing for:
 - Batch inputs of type: `[[double] | [int] | [float] | string]`
 - Expected error message on malformed batch inputs (non-matching types)

Questions/Still to-do:
 - No testing for: 
    - Input batch of type `[[byte]]`. There's no testing for type `[byte]` either, I think because `EXPECT_EQ()` doesn't take binary inputs
    - Return batch predictions only after all queries are processed
 - Some error messages/comments aren't updated to show the acceptable input formats

### DIFF
--- a/src/benchmarks/src/bench_utils.hpp
+++ b/src/benchmarks/src/bench_utils.hpp
@@ -56,8 +56,8 @@ int get_int(const std::string &key,
             std::unordered_map<std::string, std::string> &config);
 
 /**
-* Returns the value corresponding to `key` in `config` as a long
-*/
+ * Returns the value corresponding to `key` in `config` as a long
+ */
 long get_long(const std::string &key,
               std::unordered_map<std::string, std::string> &config);
 
--- a/src/benchmarks/src/bench_utils.hpp
+++ b/src/benchmarks/src/bench_utils.hpp
@@ -56,8 +56,8 @@ int get_int(const std::string &key,
             std::unordered_map<std::string, std::string> &config);
 
 /**
- * Returns the value corresponding to `key` in `config` as a long
- */
+* Returns the value corresponding to `key` in `config` as a long
+*/
 long get_long(const std::string &key,
               std::unordered_map<std::string, std::string> &config);
 
--- a/src/container/include/container/container_rpc.hpp
+++ b/src/container/include/container/container_rpc.hpp
@@ -155,9 +155,9 @@ class RPC {
   std::shared_ptr<boost::circular_buffer<RPCLogItem>> event_log_;
 
   /**
- * @return `true` if the received heartbeat is a request for container metadata.
- * `false` otherwise.
- */
+   * @return `true` if the received heartbeat is a request for container
+   * metadata. `false` otherwise.
+   */
   bool handle_heartbeat(zmq::socket_t& socket) const;
 
   void send_heartbeat(zmq::socket_t& socket) const;
--- a/src/container/include/container/container_rpc.hpp
+++ b/src/container/include/container/container_rpc.hpp
@@ -155,9 +155,9 @@ class RPC {
   std::shared_ptr<boost::circular_buffer<RPCLogItem>> event_log_;
 
   /**
-   * @return `true` if the received heartbeat is a request for container
-   * metadata. `false` otherwise.
-   */
+ * @return `true` if the received heartbeat is a request for container metadata.
+ * `false` otherwise.
+ */
   bool handle_heartbeat(zmq::socket_t& socket) const;
 
   void send_heartbeat(zmq::socket_t& socket) const;
--- a/src/frontends/src/query_frontend.hpp
+++ b/src/frontends/src/query_frontend.hpp
@@ -326,7 +326,7 @@ class RequestHandler {
         predictions
             .then([response,
                    app_metrics](std::vector<folly::Try<Response>> tries) {
-              std::string final_content;
+              std::stringstream final_content;
               for (auto t : tries) {
                 try {
                   Response r = t.value();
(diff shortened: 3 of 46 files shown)


dcrankshaw:
jenkins ok to test

src/benchmarks/src/bench_utils.hpp
```diff
@@ -56,8 +56,8 @@ int get_int(const std::string &key,
             std::unordered_map<std::string, std::string> &config);
 
 /**
-* Returns the value corresponding to `key` in `config` as a long
-*/
+ * Returns the value corresponding to `key` in `config` as a long
+ */
 long get_long(const std::string &key,
```
dcrankshaw: Revert these changes (to keep the commit history clean).

src/container/include/container/container_rpc.hpp
```diff
@@ -155,9 +155,9 @@ class RPC {
   std::shared_ptr<boost::circular_buffer<RPCLogItem>> event_log_;
 
   /**
- * @return `true` if the received heartbeat is a request for container metadata.
- * `false` otherwise.
- */
+   * @return `true` if the received heartbeat is a request for container
+   * metadata. `false` otherwise.
+   */
```
dcrankshaw: Revert

src/frontends/src/query_frontend.hpp
```diff
@@ -318,25 +318,35 @@ class RequestHandler {
...
+                  }
+                  app_metrics.latency_->insert(r.duration_micros_);
+                  app_metrics.num_predictions_->increment(1);
+                  app_metrics.throughput_->mark(1);
+
+                  std::string content = get_prediction_response_content(r);
+                  final_content += content + "\n";
+                } catch (const std::exception& e) {
+                  // case: returned a response before all predictions in the
+                  // batch were ready
```
dcrankshaw: Log this error

src/frontends/src/query_frontend.hpp
```diff
@@ -318,25 +318,35 @@ class RequestHandler {
...
-                app_metrics.default_pred_ratio_->increment(0, 1);
+        folly::Future<std::vector<folly::Try<Response>>> predictions =
+            decode_and_handle_predict(request->content.string(), name,
+                                      versioned_models, policy,
+                                      latency_slo_micros, input_type);
+
+        predictions
+            .then([response,
+                   app_metrics](std::vector<folly::Try<Response>> tries) {
+              std::string final_content;
```
dcrankshaw: Use a `std::stringstream` to accumulate the full response.

src/frontends/src/query_frontend.hpp
```diff
@@ -318,25 +318,35 @@ class RequestHandler {
...
+                    app_metrics.default_pred_ratio_->increment(1, 1);
+                  } else {
+                    app_metrics.default_pred_ratio_->increment(0, 1);
+                  }
+                  app_metrics.latency_->insert(r.duration_micros_);
+                  app_metrics.num_predictions_->increment(1);
+                  app_metrics.throughput_->mark(1);
+
+                  std::string content = get_prediction_response_content(r);
+                  final_content += content + "\n";
```
dcrankshaw: This will turn into a stringstream appends (e.g. `final_content << content << "\n";`)

src/frontends/src/query_frontend.hpp
```diff
@@ -479,22 +489,33 @@ class RequestHandler {
...
-        Query{name, uid, input, latency_slo_micros, policy, models});
-    return prediction;
+    std::vector<folly::Future<Response>> predictions;
+
+    if (d.HasMember("input")) {
+      std::shared_ptr<Input> input = clipper::json::parse_input(input_type, d);
+      auto prediction = query_processor_.predict(
+          Query{name, uid, input, latency_slo_micros, policy, models});
+      predictions.push_back(std::move(prediction));
+    } else {  // d.HasMember("input_batch") instead
```
dcrankshaw: Does it make sense to explicitly check for `d.hasMember("input_batch")` here, rather than leaving it as an else statement?

src/libclipper/include/clipper/redis.hpp
```diff
@@ -41,7 +41,7 @@ bool contains_prohibited_chars_for_group(std::string value);
 /**
  * Issues a command to Redis and checks return code.
  * \return Returns true if the command was successful.
```
dcrankshaw: revert all these whitespace changes

src/libclipper/include/clipper/rpc_service.hpp
```diff
@@ -73,13 +73,13 @@ class RPCService {
...
-  * The messages will be sent as a single, multi-part ZeroMQ message so
-  * it is very efficient.
-  */
+   * Send message takes ownership of the msg data because the caller cannot
+   * know when the message will actually be sent.
+   *
+   * \param `msg`: A vector of individual messages to send to this container.
+   * The messages will be sent as a single, multi-part ZeroMQ message so
+   * it is very efficient.
+   */
```
dcrankshaw: revert

dcrankshaw (changes requested):
This looks good. There are a couple things that you should add:
+ [ ] Add an integration test that uses the batch prediction interface for end-to-end queries. Just add another test like [this one](https://github.com/ucbrise/clipper/blob/e7ac4856afd534686f5da61b509f65acde480206/integration-tests/clipper_admin_tests.py#L376-L395) that sends batch predictions and checks to make sure the right number of predictions are returned.
+ [ ] Add an example of querying Clipper using the batch prediction interface to the [examples](https://github.com/ucbrise/clipper/tree/develop/examples/basic_query) directory and add a subsection to the Clipper documentation [here](https://github.com/ucbrise/clipper-website/blob/master/content/tutorials/basic_concepts.md#querying-clipper) that explains the batch prediction interface.

dcrankshaw (approved):
LGTM

PR #328, opened 2017-11-29
Support deploying Saved Tensorflow Models

santi81:
Current deployment of tensorflow models only supports deployment by passing the tensorflow session as a parameter.It should be possible to deploy any saved tensorflow model by passing the model directory path, like in the tfcifar example.This PR addresses the above issue.

### DIFF
--- a/clipper_admin/clipper_admin/deployers/tensorflow.py
+++ b/clipper_admin/clipper_admin/deployers/tensorflow.py
@@ -5,10 +5,12 @@ import logging
 import re
 import os
 import json
+import glob
 
 from ..version import __version__
 from ..clipper_admin import ClipperException
 from .deployer_utils import save_python_function
+from tensorflow import Session
 
 logger = logging.getLogger(__name__)
 
@@ -17,7 +19,7 @@ def create_endpoint(clipper_conn,
                     name,
                     input_type,
                     func,
-                    tf_sess,
+                    tf_sess_or_saved_model_path,
                     default_output="None",
                     version=1,
                     slo_micros=3000000,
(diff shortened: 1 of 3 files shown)


dcrankshaw:
jenkins test this please

santi81:
@dcrankshaw supports serving Frozen tensor-flow models  as well now. Can you pls review

dcrankshaw (approved):
LGTM

dcrankshaw:
jenkins ok to test

clipper_admin/clipper_admin/deployers/tensorflow.py
```diff
@@ -156,16 +158,87 @@ def predict(sess, inputs):
...
+        # Single encapsulated file(.pb extension) without un-necessary meta-data, gradients and un-necessary training variables
+
+        if os.path.isdir(tf_sess_or_saved_model_path):
+            # Directory - Check for Frozen Graph or a Saved Tensorflow Model
+            is_frozen_graph = glob.glob(
+                os.path.join(tf_sess_or_saved_model_path, "*.pb"))
+            if (len(is_frozen_graph) > 0):
+                os.makedirs(os.path.join(serialization_dir, "tfmodel"))
+                try:
+                    shutil.copyfile(
```
dcrankshaw: @kakru's suggestion

>I'd suggest using copytree instead of copyfile to keep the behavior consistent between model types (cf. line 207). Possible problem might arise if there is more than one file in the saved_model_path (check is for >0) and only the first .pb file is copied

santi81:
@dcrankshaw fixed the issue based on  comment from @kakru

santi81:
@dcrankshaw i guess the tests failed due to some other sporadic error.Can you please re-trigger the build.

PR #344, opened 2017-12-09
R Support: Handle missing libraries

Corey-Zumar:
Adds error handling when deserializing and loading a model's library dependencies. If a library cannot be loaded, the container no longer crashes immediately. Instead, a warning message is emitted.

This fixes the issue where loading the `Rclipper` library into an R environment and calling `Rclipper::build_model(...)` causes the built container to crash on deployment due to the fact that `Rclipper` is not available for installation on CRAN.


dcrankshaw:
jenkins test this please

dcrankshaw (approved):
lgtm

PR #348, opened 2017-12-20
Metrics Monitoring: A minimal viable product

simon-mo:
This PR sets up the basic structure of monitoring and provide basic functionalities:
- It exports query frontend metrics through a coupled node exporter. 
- It exports model container metrics (count, 3 types of latencies in seconds) through a node exporter in the same container but different python process.
- It initiates a mechanism to spin up a prometheus monitor and update it when necessary. 

RPC image needs to be rebuilt. A new frontend-exporter image is also created. 

Full implementation available in commit messages. 

[Simon Mo]


dcrankshaw:
jenkins okay to test

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -123,24 +125,38 @@ def start_clipper(self, query_frontend_image, mgmt_frontend_image,
...
                 '%s/tcp' % CLIPPER_INTERNAL_QUERY_PORT:
                 self.clipper_query_port,
                 '%s/tcp' % CLIPPER_INTERNAL_RPC_PORT: self.clipper_rpc_port
             },
             labels=query_labels,
             **self.extra_container_kwargs)
+
+        # Metric Section
+        query_frontend_metric_name = "query_frontend_exporter-{}".format(
+            random.randint(0, 100000))
```
dcrankshaw: Let's use the same random integer as the `query_frontend_name`

containers/python/front_end_exporter.py
```diff
@@ -0,0 +1,69 @@
+import requests
```
dcrankshaw: The containers directory is for stuff related to model containers. Can you create a separate monitoring directory (`CLIPPER_ROOT/monitoring`) and put this file in it?

containers/python/rpc.py
```diff
@@ -488,4 +503,37 @@ def start(self, model, host, port, model_name, model_version, input_type):
...
+
+
+class MetricCollector:
+    def __init__(self, pipe_child_conn):
+        self.pipe_conn = pipe_child_conn
+
+    def collect(self):
+        curr = None
+        while self.pipe_conn.poll():
+            curr = self.pipe_conn.recv()
```
dcrankshaw: If there are multiple items in the pipe, won't we overwrite all except the last item? It seems like `curr` should be a list you append to?

containers/python/rpc.py
```diff
@@ -488,4 +503,37 @@ def start(self, model, host, port, model_name, model_version, input_type):
...
+        if curr:
+            for name, val in curr.items():
+                try:
+                    yield GaugeMetricFamily(name, 'help', value=val)
```
dcrankshaw: What is the "help" string for?

containers/python/front_end_exporter.py
dcrankshaw: Can you also put a short README in the monitoring directory that provides instructions on how to access the Prometheus server once it is up?

dcrankshaw (changes requested):
This looks like a great MVP for monitoring. I just had a few small comments.

clipper_admin/clipper_admin/docker/docker_container_manager.py
simon-mo: Fixed.

containers/python/front_end_exporter.py
simon-mo: Got it! This is a much better idea.

containers/python/rpc.py
simon-mo: This is a required argument for this constructor. But not available for end-user. 
I changed it to:
```python
yield GaugeMetricFamily(name=name, 
                                          documentation=name, # Required Argument
                                           value=val)
```
for clarity.

containers/python/rpc.py
simon-mo: The purpose is to just get the the last item. Prometheus 'scrape' for the latest metric every 5 seconds and the metric is updated only for each `GET` call. 

Because the prediction time is variable, in the rpc loop, we push the most recent metric dictionary into the pipe after each prediction call. In the metric collector, we just need to get the most recent one. 

I changed the name to `lastest_metric_dict` for better clarity.

simon-mo:
@dcrankshaw It looks like Jenkins isn't starting. Can you start the check manually?

dcrankshaw:
jenkins ok to test

simon-mo:
~~Don't merge. I commented out the metrics section to let Jenkins to build new docker images.~~ Resolved

simon-mo:
@dcrankshaw Help please. 

It seems `clipper_admin_tests.py` causes the failure. It fails on the same error but with different unit tests that has nothing to do with the code I changed:
- test_inspect_instance_returns_json_dict: Bind for 0.0.0.0:37967 failed: port is already allocated. [829 log](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/829/console).
- test_python_closure_deploys_successfully: Error starting userland proxy: listen tcp 0.0.0.0:38198: bind: address already in use. [828 log](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/828/console).
- test_link_not_registered_model_to_app_fails : Error starting userland proxy: listen tcp 0.0.0.0:39526: bind: address already in use. [822 log (for batch predict)](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/822/console).

dcrankshaw:
Yeah there's an issue with out Jenkins tests. It causes the master branch builder to be super flaky too. I haven't had time to chase it down. Can you file a issue with the above comment?

simon-mo:
Sure! Done.

containers/python/rpc.py
```diff
@@ -515,13 +515,16 @@ def __init__(self, pipe_child_conn):
         self.pipe_conn = pipe_child_conn
 
     def collect(self):
-        curr = None
+        lastest_metric_dict = None
```
dcrankshaw: s/lastest/latest

monitoring/README.md
```diff
@@ -0,0 +1,13 @@
+This module is related Clipper's metric monitoring function. For full design, see [Design Doc.](https://docs.google.com/document/d/10whRxCc97gOJl4j2lY6R-v7cI_ZoAMVcNGPG_9oj6iY/edit?usp=sharing)
```
dcrankshaw: "is related to"

containers/python/rpc.py
```diff
@@ -306,6 +312,15 @@ def run(self):
 
                         response.send(socket, self.event_history)
 
+                        pred_metric['model_pred_count'] += 1
```
dcrankshaw, line 315: You're accumulating the counts here, but just sending samples of the latencies. This is why I was confused about how you were dequeuing from the pipe. In order to handle the latencies correctly you should be creating histograms. For this MVP PR, let's just accumulate the count. So delete the `*_time` metrics.

containers/python/rpc.py
```diff
@@ -488,4 +503,40 @@ def start(self, model, host, port, model_name, model_version, input_type):
...
+        lastest_metric_dict = None
+        while self.pipe_conn.poll():
+            lastest_metric_dict = self.pipe_conn.recv()
+        if lastest_metric_dict:
+            for name, val in lastest_metric_dict.items():
+                try:
+                    yield GaugeMetricFamily(
+                        name=name,
+                        documentation=name,  # Required Argument
+                        value=val)
```
dcrankshaw, line 522: s/lastest/latest

dcrankshaw (changes requested):
This is cool! I got it working and even just playing around with the bare bones Prometheus is pretty cool. I'm excited to start building out our monitoring infrastructure.

Just a couple minor comments and then this is good to go.

containers/python/rpc.py
simon-mo: Fixed

containers/python/rpc.py
simon-mo: Got it. I was thinking about this recently. I deleted `*_time` metrics. 

If we can finalize (at least for now) all the metrics needed to be collected, I will initialize them in the prometheus collector process just once [1] and run update through each dictionary in the pipe. So that every metric record is updated in the metric process (especially matters for histogram and summary data type. 

[1]: the current process initialize a new `gauge` every time the metric is updated.

containers/python/rpc.py
simon-mo: fixed

monitoring/README.md
simon-mo: fixed

containers/python/rpc.py
dcrankshaw: Yeah that's definitely the right way to do it. Let's merge this PR, then you can start on one that tracks the timing metrics with histograms.

dcrankshaw (approved):
LGTM. Fix the merge conflicts then I'll get this merged.

dcrankshaw:
jenkins test this please

PR #349, opened 2017-12-21
Fix an wrong parameter on rpc.py

withsmilo:


dcrankshaw:
Hi @withsmilo. Thanks for filing this, but currently the feedback codepath is not currently supported in Clipper. That code likely has additional bugs beyond a wrong parameter, and needs to be overhauled and tested. When we decide to add support for feedback, we'll go ahead and do that refactoring. I'm closing for now.

PR #350, opened 2018-01-02
StateDB Cache

### LINKED ISSUE #329: Add StateDB cache
Add a cache to the StateDB to prevent unnecessary Redis operations on the critical path. Cache entries should be updated appropriately in response to database “put” operations. Concurrency control is particularly important to ensure that the cache remains valid.

See https://github.com/dcrankshaw/clipper/commit/1252e07079c38f65524b4bafbf31c3f0b730d78d for an incomplete solution

Corey-Zumar:
Fixes #329


dcrankshaw:
jenkins test this please

src/libclipper/src/persistent_state.cpp
```diff
@@ -42,21 +48,58 @@ StateDB::StateDB() {
...
+  auto entry_search = cache_.find(key);
+  if (entry_search != cache_.end()) {
+    auto& cache_entry = entry_search->second;
+    std::lock_guard<std::mutex> entry_lock(cache_entry->entry_mtx_);
+    return cache_entry->entry_value_;
+  } else {
+    std::string redis_key = generate_redis_key(key);
+    const std::vector<std::string> cmd_vec{"GET", redis_key};
+    auto result =
+        redis::send_cmd_with_reply<std::string>(redis_connection_, cmd_vec);
```
dcrankshaw, line 72: If you fetch the value from Redis, you should cache it here as well.

src/libclipper/src/persistent_state.cpp
```diff
@@ -42,21 +48,58 @@ StateDB::StateDB() {
...
+    return result;
+  }
 }
 
 bool StateDB::put(StateKey key, std::string value) {
+  auto entry_search = cache_.find(key);
+  bool new_entry = (entry_search == cache_.end());
+  if (!new_entry) {
+    std::lock_guard<std::mutex> lock(entry_search->second->entry_mtx_);
```
dcrankshaw: What's the point of this lock? It will be released once it goes out of scope, so it won't be held after the if-statement.

src/libclipper/src/persistent_state.cpp
```diff
@@ -42,21 +48,58 @@ StateDB::StateDB() {
...
   const std::vector<std::string> cmd_vec{"SET", redis_key, value};
-  return redis::send_cmd_no_reply<std::string>(redis_connection_, cmd_vec);
+  bool success =
+      redis::send_cmd_no_reply<std::string>(redis_connection_, cmd_vec);
+  if (success) {
+    if (new_entry) {
+      // This method is being invoked as part of the 'add application'
+      // procedure in the query frontend. Because any subsequent feedback
+      // updates to the StateDB depend on the completion of 'add application',
+      // it's safe to add a new cache entry immediately
```
dcrankshaw: Hmm will this always be true? This seems like a complicated behavior to rely on for safety. What if another code path ends up putting things in the StateDB? Will that break this code?

dcrankshaw: Ahh I understand your comment now. Can you clarify in the comment that *if we are creating a new entry, it means that the method is being invoked as part of adding an application*?

But my previous question still remains, this seems like tricky behavior to rely on for safety.

dcrankshaw (changes requested):
This looks pretty good. It doesn't look like you ever evict things from the cache. Is that intentional?

Corey-Zumar:
@dcrankshaw A couple questions regarding performance vs consistency tradeoffs:

1. In order to ensure strong consistency, it's necessary to update the cache and database as a single atomic operation. Therefore, writers need to hold a lock on both the cache and redox handle until the write is complete. This seems fine if most StateDB operations are reads and we don't frequently evict elements (this is currently the case). However, if the distribution of reads and writes becomes more balanced and we intend to evict elements, we may not see much of an improvement in performance by adding the cache. Can we make any assumptions about StateDB access patterns going forward?

2. Another option is to use a weaker consistency model. If we take a lock on the redox handle during a write, we can ensure consistent ordering between writers. However, readers will continue to read old values from the cache until the thread that completed the Redox write operation is able to write the new cache value. This model will perform significantly faster in the case of simultaneous reads and writes. Does this sound like a better approach?

src/libclipper/src/persistent_state.cpp
Corey-Zumar: Done

Corey-Zumar:
@dcrankshaw I redesigned the `put` and `get` procedures to meet the following requirements:

1. Write operations are serialized
2. Read operation results reflect all prior writes (they may also reflect subsequent writes)

This design no longer makes any assumptions about the origin of the first `put` call.

Corey-Zumar:
jenkins test this please

dcrankshaw (approved):
LGTM. Will merge once tests pass.

Issue #339, opened 2017-11-29
Improved metrics and monitoring infrastructure


dcrankshaw:
Clipper already tracks some basic internal performance metrics, but these are tracked on an ad-hoc basis and not aggregated. As we expand support for running larger Clipper clusters with multiple query frontend replicas, we need better metrics tracking and monitoring. Furthermore, we will be expanding to provide statistical performance monitoring early next year and will want to integrate these into the existing metrics tracking infrastructure. Because this requires fleshing out some real architectural questions and expanding our set of dependencies, this feature has a separate design document associated with it.

[Design Doc (WIP)](https://docs.google.com/a/berkeley.edu/document/d/10whRxCc97gOJl4j2lY6R-v7cI_ZoAMVcNGPG_9oj6iY/edit?usp=sharing)

simon-mo:
#348 gives a MVP. I'll start working on a second PR that includes:
- [x] Allow port specification for Prometheus container port forwarding (to avoid clash, at least).
- [x] Add Monitor Example/Tutorial in example folder. Include a default visualization with Grafana section in it. [Grafana Docker](http://docs.grafana.org/installation/docker/)
- [x] Set up "configuration" like metric specification from a file `metric_config.yaml` or `metric_config.py` or `metric_config.json` looks very roughly like the following:
```yaml
# Four Metric Type {Counter | Gauge | Histogram | Summary}, 
# see: https://prometheus.io/docs/concepts/metric_types/

Model Container:
  pred_count:
    - type: Counter
    - description: The number of prediction made in this model container
  
  parse_time_ms:
    - type: Histogram
    - decription: The time in ms takes to parse RPC call from query frontend.

Query Frontend:
  query_count:
    - type: Counter
    - description: The number of prediction make in current query frontend


# Redis:
#   Redis metrics is exproted by [Redis Exporter](https://github.com/oliver006/redis_exporter)
#   It export the default redis [info](https://redis.io/commands/info)
#   BUT: it can monitors metrics for a specific key too.
Redis:
  - some_key_1
  - some_key_2

```
I proposed this to have a record of all the metrics we are collecting documented in one place (located in `$CLIPPER_ROOT/monitoring`), so that in the future it will be easy to add more metrics without changing much of the code or even without the knowledge how the metric monitoring system is setup. 

- [x] Metrics for model container:
```markdown
- Latencies
     - End-to-end
     - User’s predict function
- Counts:
     - Number of predictions
- Rates
     - Predictions per second
```
[...]

Issue #352, opened 2018-01-05
Integration tests flaky with "Error starting userland proxy... bind: address already in use"


simon-mo:
It seems `clipper_admin_tests.py` causes a lot of unexpected failure. It fails on the same error but with different unit tests that has nothing to do with the code I changed:
- test_inspect_instance_returns_json_dict: Bind for 0.0.0.0:37967 failed: port is already allocated. [829 log](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/829/console).
- test_python_closure_deploys_successfully: Error starting userland proxy: listen tcp 0.0.0.0:38198: bind: address already in use. [828 log](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/828/console).
- test_link_not_registered_model_to_app_fails : Error starting userland proxy: listen tcp 0.0.0.0:39526: bind: address already in use. [822 log (for batch predict)](https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/822/console).

lgendrot:
Total shot in the dark here, because I only recently stumbled on Clipper, but I run into port issues like this very frequently in Docker, usually after starting and stopping containers in quick succession. I can usually trace it back to containers I explicitly stopped failing to actually be stopped, or appearing to be stopped (via `docker ps`) but not giving up their ports for one reason or another.

A little digging and I found a lot of examples of similar issues:

https://github.com/moby/moby/issues/25981
https://github.com/moby/moby/issues/34097
https://github.com/moby/moby/issues/7856 

They either claim that whatever the issue was, it was fixed in more recent versions of docker, or the issues are as of yet unresolved. In general the problem seems to be with docker itself, though.

From what I can see the unit tests all run in a container built from `ClipperTestsDockerfile`, and the logs provided show that step of the build process uses the cache, and that line of the Dockerfile hasn't been updated in a while, so maybe building that image with the `--no-cache` flag so it will definitely get the latest version of docker (and everything else) will mitigate this?

Again, shot in the dark.

PR #357, opened 2018-01-18
Metrics and Monitoring V0.0.2: Model Container, Latencies, and Example

simon-mo:
As mentioned in #339. 

This PR adds the following features to monitoring MVP:
1. Create a **configuration for metrics** we are actively monitoring in `monitoring/metrics_config.yaml`. With this configuration file, we load up model container with corresponding metrics just for one time. Thus the histogram and summary observation will much more accurate. 
2. Monitoring **5 metrics in model container** now. In particular, `prediction count`, `end-to-end latency`, `{receive | parse | handle} latency`. The first in `Counter` format and the rest in `Histogram` format. 
3. Create an **example for monitoring** in `example/monitoring`, user can launch two scripts `query.py` and `init_grafana` to view Clipper Dashboard like this:
screen shot 2018-01-18 at 1 10 47 am

### DIFF
--- a/containers/python/rpc.py
+++ b/containers/python/rpc.py
@@ -7,10 +7,11 @@ import time
 from datetime import datetime
 import socket
 import sys
+import yaml
 from collections import deque
 from multiprocessing import Pipe, Process
 from prometheus_client import start_http_server
-from prometheus_client.core import GaugeMetricFamily, REGISTRY
+from prometheus_client.core import Counter, Gauge, Histogram, Summary
 
 INPUT_TYPE_BYTES = 0
 INPUT_TYPE_INTS = 1
@@ -193,8 +194,6 @@ class Server(threading.Thread):
         sys.stdout.flush()
         sys.stderr.flush()
 
-        pred_metric = dict(model_pred_count=0)
-
         while True:
             socket = self.context.socket(zmq.DEALER)
             poller.register(socket, zmq.POLLIN)
@@ -312,13 +311,21 @@ class Server(threading.Thread):
 
                         response.send(socket, self.event_history)
 
-                        pred_metric['model_pred_count'] += 1
+                        recv_time = (t2 - t1).microseconds
+                        parse_time = (t3 - t2).microseconds
+                        handle_time = (t4 - t3).microseconds
 
-                        metric_conn.send(pred_metric)
+                        model_container_metric = {}
+                        model_container_metric['pred_total'] = 1
+                        model_container_metric['recv_time_us'] = recv_time
+                        model_container_metric['parse_time_us'] = parse_time
+                        model_container_metric['handle_time_us'] = handle_time
+                        model_container_metric['end_to_end_latency_us'] = (recv_time + parse_time + handle_time)
+                        metric_conn.send(model_container_metric)
 
                         print("recv: %f us, parse: %f us, handle: %f us" %
-                              ((t2 - t1).microseconds, (t3 - t2).microseconds,
-                               (t4 - t3).microseconds))
+                              (recv_time, parse_time, handle_time))
+
                         sys.stdout.flush()
                         sys.stderr.flush()
 
@@ -499,29 +506,76 @@ class RPCService:
         self.server.model_input_type = model_input_type
         self.server.model = model
 
-        parent_conn, child_conn = Pipe(duplex=True)
+        child_conn, parent_conn = Pipe(duplex=False)
         metrics_proc = Process(target=run_metric, args=(child_conn, ))
         metrics_proc.start()
         self.server.run(parent_conn)
 
 
 class MetricCollector:
+    """
+    Note this is no longer a Prometheus Collector.
+    Instead, this is simply a class to encapsulate metric recording.
+    """
+
     def __init__(self, pipe_child_conn):
         self.pipe_conn = pipe_child_conn
+        self.metrics = {}
+        self.name_to_type = {}
+
+        self._load_config()
+
+    def _load_config(self):
+        with open('metrics_config.yaml', 'r') as f:
+            config = yaml.load(f)
+        config = config['Model Container']
+
+        prefix = 'clipper_{}_'.format(config.pop('prefix'))
+
+        for name, spec in config.items():
+            metric_type = spec.get('type')
+            metric_description = spec.get('description')
+
+            if not metric_type and not metric_description:
+                raise Exception(
+                    "{}: Metric Type and Metric Description are Required in Config File.".
+                        format(name))
+
+            if metric_type == 'Counter':
+                self.metrics[name] = Counter(prefix + name, metric_description)
+            elif metric_type == 'Gauge':
+                self.metrics[name] = Gauge(prefix + name, metric_description)
+            elif metric_type == 'Histogram':
+                if 'bucket' in spec.keys():
+                    buckets = np.linspace(*spec['bucket']).tolist()+[float("inf")]
+                    self.metrics[name] = Histogram(prefix + name,
+                                               metric_description,
+                                               buckets=buckets)
+                else:
+                    self.metrics[name] = Histogram(prefix + name,
+                                                   metric_description)
+            elif metric_type == 'Summary':
+                self.metrics[name] = Summary(prefix + name, metric_description)
+            else:
+                raise Exception(
+                    "Unknown Metric Type: {}. See config file.".format(
+                        metric_type))
+
+            self.name_to_type[name] = metric_type
 
     def collect(self):
-        latest_metric_dict = None
-        while self.pipe_conn.poll():
-            latest_metric_dict = self.pipe_conn.recv()
-        if latest_metric_dict:
-            for name, val in latest_metric_dict.items():
-                try:
-                    yield GaugeMetricFamily(
-                        name=name,
-                        documentation=name,  # Required Argument
-                        value=val)
-                except ValueError:
-                    pass
+        while True:
+            latest_metric_dict = self.pipe_conn.recv()  # This call is blocking.
+            for name, value in latest_metric_dict.items():
+                metric = self.metrics[name]
+                if self.name_to_type[name] == 'Counter':
+                    metric.inc(value)
+                elif self.name_to_type[name] == 'Gauge':
+                    metric.set(value)
+                elif self.name_to_type[name] == 'Histogram' or self.name_to_type[name] == 'Summary':
+                    metric.observe(value)
+                else:
+                    raise Exception("Unknown Metric Type for {}. See config file.".format(name))
 
 
 def run_metric(child_conn):
@@ -529,9 +583,8 @@ def run_metric(child_conn):
     This function takes a child_conn at the end of the pipe and
     receive object to update prometheus metric.
 
-    It is recommended to be ran in a separate process. 
+    It is recommended to be ran in a separate process.
     """
-    REGISTRY.register(MetricCollector(child_conn))
+    collector = MetricCollector(child_conn)
     start_http_server(1390)
-    while True:
-        time.sleep(1)
+    collector.collect()
--- a/containers/python/rpc.py
+++ b/containers/python/rpc.py
@@ -320,7 +320,8 @@ class Server(threading.Thread):
                         model_container_metric['recv_time_us'] = recv_time
                         model_container_metric['parse_time_us'] = parse_time
                         model_container_metric['handle_time_us'] = handle_time
-                        model_container_metric['end_to_end_latency_us'] = (recv_time + parse_time + handle_time)
+                        model_container_metric['end_to_end_latency_us'] = (
+                            recv_time + parse_time + handle_time)
                         metric_conn.send(model_container_metric)
 
                         print("recv: %f us, parse: %f us, handle: %f us" %
@@ -539,7 +540,7 @@ class MetricCollector:
             if not metric_type and not metric_description:
                 raise Exception(
                     "{}: Metric Type and Metric Description are Required in Config File.".
-                        format(name))
+                    format(name))
 
             if metric_type == 'Counter':
                 self.metrics[name] = Counter(prefix + name, metric_description)
@@ -547,10 +548,10 @@ class MetricCollector:
                 self.metrics[name] = Gauge(prefix + name, metric_description)
             elif metric_type == 'Histogram':
                 if 'bucket' in spec.keys():
-                    buckets = np.linspace(*spec['bucket']).tolist()+[float("inf")]
-                    self.metrics[name] = Histogram(prefix + name,
-                                               metric_description,
-                                               buckets=buckets)
+                    buckets = np.linspace(
+                        *spec['bucket']).tolist() + [float("inf")]
+                    self.metrics[name] = Histogram(
+                        prefix + name, metric_description, buckets=buckets)
                 else:
                     self.metrics[name] = Histogram(prefix + name,
                                                    metric_description)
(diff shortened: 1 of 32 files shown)


simon-mo:
Issue #352 failing the build again:
```
APIError: 500 Server Error: Internal Server Error 
("driver failed programming external connectivity on endpoint query_frontend-89208 
(5a4d3ec0ecef36ece07bce7b5ff390e794c9ea483861a6da618bffa1252607cb): 
Error starting userland proxy: listen tcp 0.0.0.0:38782: bind: address already in use")
```

containers/python/rpc.py
```diff
@@ -499,39 +507,88 @@ def start(self, model, host, port, model_name, model_version, input_type):
...
     """
     This function takes a child_conn at the end of the pipe and
     receive object to update prometheus metric.
 
-    It is recommended to be ran in a separate process. 
+    It is recommended to be ran in a separate process.
     """
-    REGISTRY.register(MetricCollector(child_conn))
+    collector = MetricCollector(child_conn)
     start_http_server(1390)
```
dcrankshaw, line 604: Does this port need to be open in the Docker container?

examples/monitoring/query.py
```diff
@@ -0,0 +1,67 @@
...
+def feature_sum(xs):
+    return [str(sum(x)) for x in xs]
+
+
+# Stop Clipper on Ctrl-C
+def signal_handler(signal, frame):
+    print("Stopping Clipper...")
+    clipper_conn = ClipperConnection(DockerContainerManager())
+    clipper_conn.stop_all()
+    sys.exit(0)
```
dcrankshaw, line 40: This is great!

integration-tests/clipper_metric.py
```diff
@@ -28,33 +29,77 @@ def feature_sum(xs):
...
+
+def get_matched_query(metric_name):
+    query = gen_match_query(metric_name)
+    logger.info("Querying: {}".format(query))
+    res = requests.get(query).json()
+    logger.info(res)
+    return res
+
+
+def check_three_node_healthy(res, node_num):
```
dcrankshaw: What is this testing for?

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
+# This Configuration files details the all metrics Clipper is collecting.
```
dcrankshaw: nit: "This configuration file details all the metrics Clipper collects."

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+Model Container:
+  prefix: mc
+
+  pred_total:
+    type: Counter
+    description: The number of prediction made in this model container
+
+  end_to_end_latency_us:
+    type: Histogram
+    description: The time in ms takes from receive RPC call to send off prediction.
```
dcrankshaw: Is this in milliseconds (ms) or microseconds (us)? If it's in microseconds, change the description. If it's milliseconds, change the name to `end_to_end_latency_ms`.

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+    bucket: [0, 500, 5] # This will be interpreted as np.linspace(0, 500, 5) + inf
+
+  recv_time_us:
+    type: Histogram
+    description: The time in ms takes to receive RPC call.
```
dcrankshaw: Same comment about ms vs us

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+    bucket: [0, 100, 5]
+
+  parse_time_us:
+    type: Histogram
+    description: The time in ms takes to parse RPC call.
```
dcrankshaw: ms vs us

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  end_to_end_latency_us:
+    type: Histogram
+    description: The time in ms takes from receive RPC call to send off prediction.
+    bucket: [0, 500, 5] # This will be interpreted as np.linspace(0, 500, 5) + inf
+
+  recv_time_us:
+    type: Histogram
+    description: The time in ms takes to receive RPC call.
+    bucket: [0, 100, 5]
```
dcrankshaw: It's unlikely that recv will ever take more than a few milliseconds. I would set the buckets to be `[0.2, 10, 0.2]`

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  parse_time_us:
+    type: Histogram
+    description: The time in ms takes to parse RPC call.
+    bucket: [0, 100, 5]
```
dcrankshaw: Parsing is also very fast. Change to `[0.2, 10, 0.2]`

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  handle_time_us:
+    type: Histogram
+    description: The time in ms takes to make prediction.
```
dcrankshaw: ms vs us

monitoring/metrics_config.yaml
```diff
@@ -0,0 +1,49 @@
...
+    bucket: [0, 100, 5]
```
dcrankshaw: Handling could take longer than 100 ms. Update to `[0, 500, 5]`

examples/monitoring/init_grafana.py
```diff
@@ -0,0 +1,48 @@
...
+        grafana.stop()
+    except Exception as e:
+        pass
+    sys.exit(0)
+
+
+if __name__ == '__main__':
+    signal.signal(signal.SIGINT, signal_handler)
+
+    print("(1/3) Initiating Grafana")
```
dcrankshaw: s/Initiating/Initializing

examples/monitoring/init_grafana.py
```diff
@@ -0,0 +1,48 @@
...
+    client = docker.from_env()
+    container = client.containers.run(
+        "grafana/grafana", ports={'3000/tcp': 3000}, detach=True)
+    print("(2/3) Grafana Initiated ")
```
dcrankshaw: s/Initiated/Initialized

dcrankshaw (changes requested):
Very cool. I like the grafana example.

containers/python/rpc.py
simon-mo: Nope. This port does not need to be exposed to the user. Prometheus can detect and access this port since the Prometheus container and model container are technically in the same "pod". This might need to change to k8s though.

examples/monitoring/init_grafana.py
simon-mo: Fixed

examples/monitoring/init_grafana.py
simon-mo: Fixed

integration-tests/clipper_metric.py
simon-mo: It was a helper method to test how many container metric is prometheus collecting. This is just bad naming. I should change it to `parse_res_and_assert_node` and write a docstring for it.

monitoring/metrics_config.yaml
simon-mo: Fixed

monitoring/metrics_config.yaml
simon-mo: Changed it to microseconds (us)

monitoring/metrics_config.yaml
simon-mo: Changed it to microseconds (us)

monitoring/metrics_config.yaml
simon-mo: Changed it to microseconds (us)

monitoring/metrics_config.yaml
simon-mo: @dcrankshaw 

That's too many buckets though. Each bucket is a prometheus time series. We should have around 5 to 10 "bucket" for each histogram metric. 

For the simple query example running on my local machine, recv time takes around `40 us`, parse time takes around `100 us`, handle time takes around `90 us`. So the end to end is about `230 us`. 

For the range specification, it works like `[start, end, num_of_interval]`.

How about the following bucket (in us)? I make the general histogram to have a long tail to observe abnormal event.

Type | Time
-|-
recv| `[20, 120, 5]`
parse | `[50, 250, 5]`
handle | `[50, 250, 5]`
end-to-end | `[100, 600, 5]`

In the end, we need to let the user configure it and pass it dynamically somehow. Or configure it according to the user's SLO objective, but that is dynamically configured still. Right now, the metric and bucket is set when building the RPC container, I should think of a better way to dynamically pass it in somehow.

monitoring/metrics_config.yaml
simon-mo: changed to us

monitoring/metrics_config.yaml
dcrankshaw: Okay that makes sense. Those buckets are fine for recv and parse. But handle will be much longer for any real application. I agree that it makes sense to dynamically set the handle buckets based on the SLO, but for now can we do logarithmically spaced buckets from 1ms to 500 ms. Something like (in ms):
[5, 10, 20, 35, 50, 75, 100, 150, 200, 250, 300, 400, 500]

That's a little bit more than 10 buckets, but for any given application most of those buckets will be empty.

simon-mo: in `us`, right? 

Since the model container log is printing:
```
Got start of message 145
recv: 62.000000 us, parse: 100.000000 us, handle: 81.000000 us
...
```

dcrankshaw: No in ms. Are you testing with the noop model container? Any real models will take substantially longer. E.g. a deep learning model may take up to a few hundred ms.

Actually now that I think about it let's change the recv and parse histograms to use (in us) `[100, 1000, 100]` (so go up to 1 ms in 100 us increments). We don't need fine-grained latency tracking on the order of microseconds or tens of microseconds. Remember that some applications may be sending very larger inputs that take much longer than a few microseconds to recv and parse.

simon-mo: Ok what about the following:

```yaml
Model Container:
  prefix: mc

  pred_total:
    type: Counter
    description: The number of prediction made in this model container

  end_to_end_latency_us:
    type: Histogram
    description: The time in us takes from receive RPC call to send off prediction.
    bucket: 
      log: [100, 10000, 10]
          #[  100,   170,   280,   460,   770,  1290,  2150,  3590,  5990, 10000]
  recv_time_us:
    type: Histogram
    description: The time in us takes to receive RPC call.
    bucket: 
      linear: [100, 1000, 10]
                #[ 100,  200,  300,  400,  500,  600,  700,  800,  900, 1000]
  parse_time_us:
    type: Histogram
    description: The time in us takes to parse RPC call.
    bucket: 
      linear: [100, 1000, 10]
                #[ 100,  200,  300,  400,  500,  600,  700,  800,  900, 1000]
  handle_time_us:
    type: Histogram
    description: The time in us takes to make prediction.
    bucket: 
      log: [100, 10000, 10]
           #[  100,   170,   280,   460,   770,  1290,  2150,  3590,  5990, 10000]
```

dcrankshaw: 10 ms for handle time and end-to-end time is not enough. Handle time is the time for the model to make a prediction. It needs to go up to 500 ms. Can you just explicitly set the bucket boundaries to what I specified before ([5, 10, 20, 35, 50, 75, 100, 150, 200, 250, 300, 400, 500])?

simon-mo: Sure! The implementation in RPC will be simpler too! 

```yaml
Model Container:
  prefix: mc

  pred_total:
    type: Counter
    description: The number of prediction made in this model container

  end_to_end_latency_ms:
    type: Histogram
    description: The time in ms takes from receive RPC call to send off prediction.
    bucket: [5, 10, 20, 35, 50, 75, 100, 150, 200, 250, 300, 400, 500]

  recv_time_us:
    type: Histogram
    description: The time in us takes to receive RPC call.
    bucket: [100, 200, 300, 400, 500, 600, 700, 800, 900, 1000]

  parse_time_us:
    type: Histogram
    description: The time in us takes to parse RPC call.
    bucket: [100, 200, 300, 400, 500, 600, 700, 800, 900, 1000]

  handle_time_ms:
    type: Histogram
    description: The time in ms takes to make prediction.
    bucket: [5, 10, 20, 35, 50, 75, 100, 150, 200, 250, 300, 400, 500]

```

dcrankshaw: Great. Yeah this looks good. It's probably better to use the same time units for all the histograms though (this will avoid future confusion). Can you make all of the microseconds into milliseconds and just use floats for recv and parse time?

simon-mo: Sure! Done.

simon-mo:
@dcrankshaw ready for another round of review now.

containers/python/rpc.py
```diff
@@ -312,13 +311,25 @@ def run(self, metric_conn):
...
+                        model_container_metric = {}
+                        model_container_metric['pred_total'] = 1
+                        model_container_metric[
+                            'recv_time_ms'] = recv_time / 1000
+                        model_container_metric[
+                            'parse_time_ms'] = parse_time / 1000
+                        model_container_metric[
+                            'handle_time_ms'] = handle_time / 1000
+                        model_container_metric['end_to_end_latency_ms'] = (
+                            recv_time + parse_time + handle_time) / 1000
```
dcrankshaw: Can you make these divisions by `1000.0`?

examples/monitoring/query.py
```diff
@@ -0,0 +1,67 @@
...
+
+if __name__ == '__main__':
+    signal.signal(signal.SIGINT, signal_handler)
+    clipper_conn = ClipperConnection(DockerContainerManager())
+    clipper_conn.start_clipper()
+    print("Starting Clipper")
+    python_deployer.create_endpoint(
+        clipper_conn, "simple-example", "doubles", feature_sum, num_replicas=2)
+    time.sleep(2)
+    print("Staring Prediction")
```
dcrankshaw: nit: "Starting predictions"

integration-tests/clipper_metric.py
```diff
@@ -28,33 +29,77 @@ def feature_sum(xs):
...
+    return res
+
+
+def parse_res_and_assert_node(res, node_num):
+    assert res['status'] == 'success'
+    assert len(res['data']) == node_num
+
+
+def log_docker_ps():
+    logger.info(subprocess.check_output(['docker', 'ps']).decode('ascii'))
```
dcrankshaw: Use the [Python Docker SDK](https://docker-py.readthedocs.io/en/stable/) not subprocess

dcrankshaw (changes requested):
Just a couple nitpicks then this is good to go

dcrankshaw:
@simon-mo This basically ready. Just had a couple small comments.

containers/python/rpc.py
simon-mo: Fixed.

examples/monitoring/query.py
simon-mo: Fixed

integration-tests/clipper_metric.py
simon-mo: Fixed.

dcrankshaw (approved):
LGTM

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
Last try of today... The new update does not seem relevant to the "conda install tensorflow" issue.

simon-mo:
Similar issue here for reference: https://github.com/tensorflow/tensorflow/issues/8096#issuecomment-285672395

PR #368, opened 2018-01-21
Fix port already bound error in integration tests

### LINKED ISSUE #352: Integration tests flaky with "Error starting userland proxy... bind: address already in use"

dcrankshaw:
I'm still trying to figure out why the unit tests are flaky. I'm hoping this fixes #352. This PR attempts two fixes:
+ Expanding the port range that we randomly generate ports from to reduce the collision probability.
+ To test if a port is bound, we call `socket.bind(PORT_NUM)` in Python and if the bind successfully completes, we know that nothing else is already bound on that port. However, we rely on the socket (that we just bound to) to be closed by the Python socket object going out of scope and being garbage collected. It's possible that that GC doesn't always happen fast enough. This PR explicitly closes the socket before returning the port number.

### DIFF
--- a/bench/bench_init.py
+++ b/bench/bench_init.py
@@ -35,8 +35,8 @@ class BenchSetup():
 
     def get_cifar_data(self):
         train_x, train_y = cifar_utils.filter_data(*cifar_utils.load_cifar(
-            self.cifar_dir_path, cifar_filename="cifar_train.data",
-            norm=False))
+            self.cifar_dir_path, cifar_filename="cifar_train.data", norm=False)
+                                                   )
         test_x, test_y = cifar_utils.filter_data(*cifar_utils.load_cifar(
             self.cifar_dir_path, cifar_filename="cifar_test.data", norm=False))
 
--- a/bench/bench_init.py
+++ b/bench/bench_init.py
@@ -35,8 +35,8 @@ class BenchSetup():
 
     def get_cifar_data(self):
         train_x, train_y = cifar_utils.filter_data(*cifar_utils.load_cifar(
-            self.cifar_dir_path, cifar_filename="cifar_train.data", norm=False)
-                                                   )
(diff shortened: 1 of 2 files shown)


dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

simon-mo:
@dcrankshaw Looks like Jenkins is stuck on forever loop:
https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/869/console

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw:
My final workaround was to simply catch the error that the Docker API already throws, clean everything up, and try starting Clipper again on new ports. This does not address the root cause of the error (which is still unknown but likely a Docker issue), but it does prevent the issue from affecting our tests.

Corey-Zumar (approved):
LGTM

PR #378, opened 2018-01-28
Container heartbeat

Unous1996:
Print error log when losing contact with a container.


Unous1996:
Sorry, it was a little bit messy.

src/libclipper/src/rpc_service.cpp
```diff
@@ -141,16 +145,34 @@ void RPCService::manage_service(const string address) {
       // TODO: Balance message sending and receiving fairly
       // Note: We only receive one message per event loop iteration
       log_info(LOGGING_TAG_RPC, "Found message to receive");
-
       receive_message(socket, connections, connections_containers_map,
                       zmq_connection_id, redis_connection);
+      auto current_time = std::chrono::system_clock::now();
```
Corey-Zumar: This logic should be moved outside the `if` conditional. We want to check for dead containers regardless of whether or not a message has been received.

src/libclipper/src/rpc_service.cpp
```diff
@@ -141,16 +145,34 @@ void RPCService::manage_service(const string address) {
...
+      std::chrono::duration<double> time_pass_since_last_check = current_time - last_check_time_;
```
Corey-Zumar: Use `std::chrono::duration<std::chrono::milliseconds>` instead. I'm not sure what `double` means here.

src/libclipper/src/rpc_service.cpp
```diff
@@ -28,16 +29,19 @@ namespace clipper {
 namespace rpc {
 
 constexpr int INITIAL_REPLICA_ID_SIZE = 100;
+constexpr int CONTACTING_GAP_TOLORANCE = 30;
```
Corey-Zumar: Rename this to `CONTAINER_ACTIVITY_TIMEOUT`

src/libclipper/src/rpc_service.cpp
```diff
@@ -28,16 +29,19 @@ namespace clipper {
...
+constexpr double CONTAINER_EXISTENCE_CHECK_FREQUENCY = 0.2;
```
Corey-Zumar: What does `.2` mean in terms of time? Let's make this a long called `CONTAINER_ACTIVITY_CHECK_FREQUENCY_MILLIS` and define it to be `10000` (10 seconds, in milliseconds).

src/libclipper/src/rpc_service.cpp
```diff
@@ -141,16 +145,34 @@ void RPCService::manage_service(const string address) {
...
+         last_check_time_ = current_time;
+      }
     }
     // Note: We send all queued messages per event loop iteration
     send_messages(socket, connections);
   }
   shutdown_service(socket);
 }
 
+void RPCService::check_containers_existence(std::map<const vector<uint8_t>, std::chrono::system_clock::time_point> receiving_history_){
```
Corey-Zumar: Nit: rename method to `check_container_activity`

Corey-Zumar: This method doesn't need to take the receiving history as a parameter. `receiving_history_` is a member variable of the `RPCService` class and can be referenced directly in the method body.

src/libclipper/src/rpc_service.cpp
```diff
@@ -141,16 +145,34 @@ void RPCService::manage_service(const string address) {
...
+    std::map<const vector<uint8_t>, std::chrono::system_clock::time_point>::iterator it;
+    std::chrono::system_clock::time_point current_time;
+    std::chrono::duration<double> time_elapsed_since_last_contact;
```
Corey-Zumar: Use `std::chrono::duration<std::chrono::milliseconds>` instead

Corey-Zumar (changes requested):
@Unous1996 This is a great start! I saw that you accidentally tracked some files in the build directory, so I reset the branch to the commit with hash `606ad512172145d0aa93a9d559c663278a300485`. Please address these comments when you have time.

Unous1996:
OK. I will fix them as soon as possible.

src/libclipper/src/rpc_service.cpp
Unous1996: Well, these constants I defined was in seconds, and the CONTAINER_EXISTENCE_CHECK_FREQUENCY is the time gap between the two continuous container existence check, so wouldn't 10 seconds be too long.

src/libclipper/src/rpc_service.cpp
```diff
@@ -141,33 +141,32 @@ void RPCService::manage_service(const string address) {
       poll_timeout = 1;
     }
     zmq_poll(items, 1, poll_timeout);
+
+    auto current_time = std::chrono::system_clock::now();
```
Corey-Zumar: Nit: Place this logic directly after the call to `send_messages`.

Corey-Zumar (changes requested):
@Unous1996  Nice! This looks good. Remove the `cmake-build-debug` subdirectory from your PR by running the following:

`git rm -r cmake-build-debug`

Then, commit these changes to your branch and push.

Unous1996:
It seems that I met some problem when I ran `git m -r cmake-build-debug`, I could only remove the local caches.

src/libclipper/include/clipper/rpc_service.hpp
```diff
@@ -111,9 +117,11 @@ class RPCService {
   std::atomic_bool active_;
   // The next available message id
   int message_id_ = 0;
+  std::chrono::system_clock::time_point last_check_time_;
   std::unordered_map<VersionedModelId, int> replica_ids_;
   std::shared_ptr<metrics::Histogram> msg_queueing_hist_;
-
+  //std::vector<std::pair<const vector<uint8_t>, std::chrono::system_clock::time_point>> receiving_history_;
```
Corey-Zumar: Remove commented code

src/libclipper/src/rpc_service.cpp
```diff
@@ -269,8 +289,10 @@ void RPCService::receive_message(
 
         TaskExecutionThreadPool::create_queue(model, cur_replica_id);
         zmq_connection_id += 1;
+        document_receive_time(connection_id);
```
Corey-Zumar: Instead of calling this function in each "case" statement, let's make a single call on line 244 (after the creation of `connection_id`)

src/libclipper/src/rpc_service.cpp
```diff
@@ -28,16 +29,19 @@ namespace clipper {
 namespace rpc {
 
 constexpr int INITIAL_REPLICA_ID_SIZE = 100;
+constexpr long CONTAINER_ACTIVITY_TIMEOUT = 30000;
```
Corey-Zumar: Nit: Make this `CONTAINER_ACTIVITY_TIMEOUT_MILLIS`. Specifying units of time in these variable names makes it easier to understand their meaning.

src/libclipper/src/rpc_service.cpp
```diff
@@ -28,16 +29,19 @@ namespace clipper {
...
+constexpr long CONTAINER_EXISTENCE_CHECK_FREQUENCY = 10000;
```
Corey-Zumar: Nit: Make this `CONTAINER_EXISTENCE_CHECK_FREQUENCY_MILLIS`

src/libclipper/src/rpc_service.cpp
```diff
@@ -137,20 +141,36 @@ void RPCService::manage_service(const string address) {
...
   shutdown_service(socket);
 }
 
+void RPCService::check_container_activity(){
+    std::map<const vector<uint8_t>, std::chrono::system_clock::time_point>::iterator it;
+    std::chrono::system_clock::time_point current_time;
+    for(it = receiving_history_.begin(); it!=receiving_history_.end(); it++){
+        current_time = std::chrono::system_clock::now();
+        if(std::chrono::duration_cast<std::chrono::milliseconds>(current_time - it->second).count() > CONTAINER_ACTIVITY_TIMEOUT){
+            log_info(LOGGING_TAG_RPC, "lost contact with a container");
```
Corey-Zumar: If we lose contact with the container, we should remove it from `receiving_history_`. Otherwise, this log message continues to fire every 10 seconds.

Corey-Zumar (changes requested):
Almost there! Just a few minor comments.

Unous1996:
I got the following error while trying to remove the cmake-build-debug/ folder.

`fatal: could not lookup name for submodule 'cmake-build-debug/googletest-src'`

Corey-Zumar:
@Unous1996 try `git rm -r --cached cmake-build-debug`

Unous1996:
Yeah, I just tried this command ... and got the above error log. So I used git rm -r --cached cake-build-debug/ instead.

Corey-Zumar:
Did that work? Can you commit the result?

Unous1996:
It only removed some files in the cmake-build-debug/ folder but I can't remove the folder completely...I ran into the same issue while removing that folder last time.

src/libclipper/src/rpc_service.cpp
```diff
@@ -137,20 +143,46 @@ void RPCService::manage_service(const string address) {
...
       // Note: We only receive one message per event loop iteration
       log_info(LOGGING_TAG_RPC, "Found message to receive");
-
       receive_message(socket, connections, connections_containers_map,
                       zmq_connection_id, redis_connection);
     }
     // Note: We send all queued messages per event loop iteration
     send_messages(socket, connections);
+    auto current_time = std::chrono::system_clock::now();
+    if(std::chrono::duration_cast<std::chrono::milliseconds>(current_time - last_fire_log_time_).count() > LOST_CONTACT_FIRE_LOG_FREQUQNCY_MILLS){
```
Corey-Zumar: I don't think this is necessary. We're already logging the loss of container contact on line 175. After logging on line 175, we now correctly remove the container from the receiving history and should not log again. Remove `fire_lost_contact_log()` and its associated timing variable `LOST_CONTACT_FIRE_LOG_FREQUQNCY_MILLS`.

If I'm missing something here, please let me know.

Corey-Zumar (changes requested):
Almost there, just remove the `lost_contact_fire_log()` functionality and we should be good to go.

Corey-Zumar:
@Unous1996 I'm going to format the code and get this merged. Nice job!

Unous1996:
OK.

Corey-Zumar:
jenkins test this please

Unous1996:
?

Corey-Zumar:
@Unous1996 That message is used to direct Jenkins (our test executor) to test the PR's changes. No further action is required.

Corey-Zumar:
jenkins test this please

amogkam:
@Corey-Zumar I added the updates to remove the container from `active_containers` if it is killed. This works as expected and the container is removed.

However, when I try sending in a new request once the container is killed, I still get a response that it exceeded the latency SLO. Clipper adds a task to the model queue, but gets stuck after that. Can you confirm the expected behavior for when you send a request after the model is disconnected?

amogkam:
Once we lose connection to a container, `get_model_replica` which is called in `on_container_ready` should return nullptr, meaning that the following if statement is true and the program should raise an error that the container doesn't exist anymore. But, I added log statements to debug and it seems like `on_container_ready` isn't even being called when I send a request to a dead container.

amogkam:
@dcrankshaw Maybe you can help we out with this too?

Corey-Zumar:
@amogkam Can you clarify what you mean by "the request gets stuck?" It's perfectly expected behavior for a query to time out in the case that there are no connected containers corresponding to the model associated with the queried application. The idea is that, when the container reconnects after the activity timeout window (30 seconds), queries should be successfully routed to the new container without any timeouts. Does this make sense?

amogkam:
@Corey-Zumar The task gets added to the model queue, but the request is not sent to the model. But shouldn't an error be thrown if a request is sent to a disconnected container? 

Inside `on_container_ready`, a call is made to `get_model_replica`  which should return a null pointer since the replica is no longer in `active_containers`. This means that `on_container_ready` should error out, but this isn't happening, so I think there's something else before `on_container_ready` that Clipper is getting stuck on or is waiting on.

Corey-Zumar:
@amogkam The `on_container_ready` function is executed on a separate thread managed by the task execution thread pool. Therefore, an error during the function call will not cause clipper to crash and will not be logged. It is therefore very possible that the function crashes due to an attempt to dereference a null pointer. Try adding a try/catch block with a log statement to verify this.

amogkam:
@Corey-Zumar  Ah ok, I'll add a try catch to see if the error was raised. 

If I bring the container back online, it successfully routes requests to the new container. However, the original code without these changes also behaves in the same way, so I'm confused as to what behavior should change with these modifications.

Corey-Zumar:
@amogkam the idea is that the first request after the container comes back online does not route properly. All subsequent requests should succeed. Please verify that this is the case.

amogkam:
@Corey-Zumar I see what you mean. Correct me if I'm wrong, but it seems like the requests are being routed sequentially. The way I was testing before, I would connect a container, send a request, then disconnect the container, start it back up again, and send another request, this would work. But, if I disconnect the container and reconnect it multiple times before sending in another request, all the replicas are still considered active, and the request gets routed to an inactive replica.

In the updated code, once a container replica is removed from `active_containers`,requests no longer get routed to it. However, the TaskExecutor doesn't actually call the inactive container callback unless a request is sent. So I have to send a request first, then the replica is removed, and then all subsequent requests no longer get routed to the inactive replica. This seems like a problem with the order the threads are being executed in, and I'm not sure what's causing this issue.

src/libclipper/src/rpc_service.cpp
```diff
@@ -140,16 +145,61 @@ void RPCService::manage_service(const string address) {
...
+        std::pair<VersionedModelId, int> container_info =
+            container_info_entry->second;
+
+        VersionedModelId vm = container_info.first;
+        int replica_id = container_info.second;
+      TaskExecutionThreadPool::submit_job(
+            vm, replica_id, inactive_container_callback_, vm, replica_id);
+      
+      log_info(LOGGING_TAG_RPC, "lost contact with a container");
+      receiving_history_.erase(it);
```
amogkam: @Corey-Zumar after we exceed  CONTAINER_ACTIVITY_TIMEOUT_MILLIS, we enter this if statement. The log statement on line 197 is logged correctly. However, `submit_job` for the `inactive_container_callback` is not being executed unless a request is sent. But, these 2 statements are right after each other, so I'm not really sure why the `inactive_container_callback_` is only executed when a request is sent. Every time the log statement is printed out, we want the callback to be executed as well, not only when a request is sent.

dcrankshaw: The issue here is that the `TaskExecutionThreadPool` has a single thread for each container, and the callbacks get executed sequentially on that threadpool. The `inactive_container_callback_` is blocked by a call to `on_container_ready` which is already in the threadpool queue. `on_container_ready` blocks until there is a request to dequeue here (`get_batch()` is a blocking call). https://github.com/ucbrise/clipper/blob/42abb257976f8384f96febab270c8fd0baafb906/src/libclipper/include/clipper/task_executor.hpp#L449-L450

I think the easiest fix is to make `get_batch()` a non-blocking call by changing the call from `wait` to `wait_for` on line 144, and waiting for 100ms: https://github.com/ucbrise/clipper/blob/42abb257976f8384f96febab270c8fd0baafb906/src/libclipper/include/clipper/task_executor.hpp#L144

Instead of blocking until there's a new batch available, we will just return a batch of size 0. Then, in the else statement on `on_container_ready`, instead of logging an error, submit `on_container_ready` to the threadpool again:
https://github.com/ucbrise/clipper/blob/42abb257976f8384f96febab270c8fd0baafb906/src/libclipper/include/clipper/task_executor.hpp#L479-L485
If no other callbacks get submitted to the threadpool, this will be functionally the same as making `get_batch` blocking. But it will allow the `inactive_container_callback` to be executed without having to wait for another request to come in before the container is removed.

dcrankshaw: @Corey-Zumar what do you think of my suggested modification?

Corey-Zumar: @dcrankshaw This sounds similar to our previous approach (circa last April) of resubmitting the `on_container_ready` callback to the threadpool if a query was not instantaneously available; we did not wait for 100ms. That approach seemed to be CPU-intensive, but I would imagine that the 100 millisecond wait would alleviate some of the overhead. 

Another option would be to create a garbage collection service backed by a single thread that handles this cleanup without relying upon threadpool execution. This service could later be extended to perform other GC-related tasks as they arise. Let me know what your thoughts are.

dcrankshaw: Yeah I was also thinking about a GC thread. A GC thread is probably the right way to do this, as you rightly point out that we are likely to want to do more cleanup in the future. We need to be careful about how we clean up though, because in `on_container_ready` we acquire a shared_ptr to the `ModelContainer` object. If that container becomes inactive, the next time a request comes in, it will be dequeued from the model queue to be sent to the inactive container.

To do proper GC, we need the model queue to not dequeue a batch to an inactive container. Maybe we could pass the container pointer to `get_batch()` explicitly instead of capturing it with closure capture, and then the model queue will check that the container it is returning the batch to is still active before returning. If it finds that the container is inactive, it can return an empty batch.

I have another question too. Is the container's threadpool removed when the container is found to be inactive? If not, it should be.

Corey-Zumar:
@amogkam Is this ready for review?

dcrankshaw:
jenkins ok to test

Corey-Zumar:
@amogkam I refactored the threadpool header to implement a `ModelQueueThreadPool` and a `FixedSizeThreadPool` that both inherit from the same `ThreadPool` object. This removes the need to create artificial/internal queue id's for the garbage collection use case. `GarbageCollectionThreadPool` is now a fixed-size threadpool.

dcrankshaw:
Looks like there's a compile error causing the tests to fail

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -347,22 +365,12 @@ class TaskExecutor {
       // add each task to the queue corresponding to its associated model
       boost::shared_lock<boost::shared_mutex> lock(model_queues_mutex_);
       auto model_queue_entry = model_queues_.find(t.model_);
+
       if (model_queue_entry != model_queues_.end()) {
-        output_futures.push_back(cache_->fetch(t.model_, t.input_));
```
Corey-Zumar, line 349: @amogkam Why was this logic rewritten?

amogkam: @Corey-Zumar Even if a container is removed, we want to return a prediction if it has been cached.

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -286,26 +247,104 @@ class ThreadPool {
...
+          std::forward_as_tuple(&ModelQueueThreadPool::worker, this, queue_id));
+      log_info_formatted(LOGGING_TAG_THREADPOOL,
+                         "Work queue created for model {}, replica {}",
+                         vm.serialize(), std::to_string(replica_id));
+      return true;
+    }
+  }
+
+ private:
+  static size_t get_queue_id(const VersionedModelId& vm, const int replica_id) {
```
amogkam: @Corey-Zumar Was there any reason you made `get_queue_id` private? It's failing some one of the tests that is calling this method.

dcrankshaw: If there's a test that's calling this method, go ahead and make it public.

src/libclipper/include/clipper/task_executor.hpp
```diff
@@ -371,6 +378,29 @@ class TaskExecutor {
             cur_model_metric.cache_hit_ratio_->increment(1, 1);
           }
         }
+
+        else {
+          if (active_containers_->get_replicas_for_model(t.model_).size() ==
```
Corey-Zumar: Nit: Instead of `else { if (condition)`, rewrite to use `else if (condition) { ...`

Corey-Zumar (changes requested):

src/libclipper/src/rpc_service.cpp
```diff
@@ -102,14 +103,14 @@ vector<RPCResponse> RPCService::try_get_responses(const int max_num_responses) {
 void RPCService::manage_service(const string address) {
   // Map from container id to unique routing id for zeromq
   // Note that zeromq socket id is a byte vector
-  log_info_formatted(LOGGING_TAG_RPC, "RPC thread started at address: ",
-                     address);
+  log_info_formatted(LOGGING_TAG_RPC,
+                     "RPC thread started at address: ", address);
```
withsmilo: Add brackets.
```
log_info_formatted(LOGGING_TAG_RPC,
                                 "RPC thread started at address: {}", address);
```

amogkam:
jenkins test this please

simon-mo:
@amogkam I have the same `grpc: the connection is unavailable` occurs twice. Now it's a different caffe issue.

amogkam:
jenkins test this please

amogkam:
@simon-mo I just ran it again and it seems to be working now, so the error only happens sometimes.

Corey-Zumar (approved):
LGTM

PR #387, opened 2018-01-31
Fix flakiness in fixed batch size integration test

### LINKED ISSUE #386: Fixed batch size integration test is flaky
`test_fixed_batch_size_model_processes_specified_query_batch_size_when_saturated` from the Clipper manager integration tests is flaky. After discussing with @Corey-Zumar, sticking a `time.sleep` in the model container to allow Clipper enough time to enqueue all the requests should address the issue.

Corey-Zumar:
Fixes #386

### DIFF
--- a/dockerfiles/PyClosureContainerDockerfile
+++ b/dockerfiles/PyClosureContainerDockerfile
@@ -3,12 +3,11 @@ FROM clipper/py-rpc:${CODE_VERSION}
 
 COPY clipper_admin/clipper_admin/python_container_conda_deps.txt /lib/
 RUN conda config --set ssl_verify no \
+  && conda install -y -c anaconda cloudpickle=0.5.2 \
   && conda install -y --file /lib/python_container_conda_deps.txt
-RUN conda install -c anaconda cloudpickle=0.5.2
 
 COPY containers/python/python_closure_container.py containers/python/python_closure_container_entry.sh /container/
 
-
-CMD ["/container/python_closure_container_entry.sh"]
+ENTRYPOINT ["/container/python_closure_container_entry.sh"]
 
 # vim: set filetype=dockerfile:
--- a/integration-tests/clipper_admin_tests.py
+++ b/integration-tests/clipper_admin_tests.py
@@ -375,7 +375,7 @@ class ClipperManagerTestCaseLong(unittest.TestCase):
 
         self.clipper_conn.register_application(
             self.app_name_4, self.input_type, self.default_output,
-            self.latency_slo_micros)
+            slo_micros=30000000)
 
     @classmethod
     def tearDownClass(self):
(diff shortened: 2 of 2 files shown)


dcrankshaw (approved):
lgtm

dcrankshaw:
Hmm @Corey-Zumar looks like this maybe doesn't fix the flakiness https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/936/console

Corey-Zumar:
@dcrankshaw Based on testing with AWS, it appears that the container crashes when queried. No stack trace is produced. The issue appears to be associated with the serialization and deserialization of functions that call `time.sleep` or `Timer.Start()`.

For now, in place of `time.sleep()`, it seems reasonable to perform a higher-latency computation before returning the batch sizes as outputs. Do you agree?

dcrankshaw:
Hmm weird. You could just hardcode a new testing docker container that has a sleep in it (similar to the noop container). That seems like a potentially cleaner and more controlled solution.

Corey-Zumar:
Good call. Will do that.

Corey-Zumar:
@dcrankshaw The container-based approach doesn't fare any better. I'm still encountering the same traceless crash. Will need to debug.

dockerfiles/PyClosureContainerDockerfile
```diff
@@ -3,12 +3,11 @@ FROM clipper/py-rpc:${CODE_VERSION}
...
 RUN conda config --set ssl_verify no \
+  && conda install -y -c anaconda cloudpickle=0.5.2 \
   && conda install -y --file /lib/python_container_conda_deps.txt
-RUN conda install -c anaconda cloudpickle=0.5.2
 
 COPY containers/python/python_closure_container.py containers/python/python_closure_container_entry.sh /container/
 
-
-CMD ["/container/python_closure_container_entry.sh"]
+ENTRYPOINT ["/container/python_closure_container_entry.sh"]
```
Corey-Zumar, line 11: Changing this from `CMD` to `ENTRYPOINT` fixes the crash associated with the python call to `time.sleep`. I'll have to explore why this is the case. `CMD` executes the python script within `/bin/sh -c` directive, while `ENTRYPOINT` executes the python script in some other fashion; I'd imagine that this accounts for the difference.

dockerfiles/PyClosureContainerDockerfile
```diff
@@ -3,12 +3,11 @@ FROM clipper/py-rpc:${CODE_VERSION}
 
 COPY clipper_admin/clipper_admin/python_container_conda_deps.txt /lib/
 RUN conda config --set ssl_verify no \
+  && conda install -y -c anaconda cloudpickle=0.5.2 \
```
Corey-Zumar, line 6: Installing this first fixes an installation error.

PR #389, opened 2018-02-01
updated clipper_admin.py for config files

nduggirala:


clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -91,11 +93,26 @@ def start_clipper(
...
         cache_size : int, optional
             The size of Clipper's prediction cache in bytes. Default cache size is 32 MiB.
-
+        config_file: str(optional)
+            File for the configuration of the application or model
         Raises
         ------
         :py:exc:`clipper.ClipperException`
         """
+        if(config_file is not None):
```
Corey-Zumar: This logic should occur after clipper has been started successfully. Move it to a separate method called `parse_configuration` and invoke the method after the try/catch block below.

Corey-Zumar: Nit: `None` evaluates to `False`. This can be `if config_file`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -91,11 +93,26 @@ def start_clipper(
...
+           with open(config_file, "r") as stream:
+                config = yaml.load(stream)
+                for key in config:
+                    name = config[key]["name"]
+                    version = config[key]["version"]
+                    input_type = config[key]["input_type"]
+                    obj_type = config[key]["obj_type"]
+                    config_image = config[key]["image"]
```
Corey-Zumar: Applications will not need a configuration image, so we should only check for this key if we're parsing a **model** configuration.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -91,11 +93,26 @@ def start_clipper(
...
+                    slo = config[key]["slo"]
```
Corey-Zumar: Models will not need an associated SLO, so we should only check for this key if we're parsing an **application** configuration

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -91,11 +93,26 @@ def start_clipper(
...
+                    if(obj_type == "application"):
+                        self.register_application(name, input_type, "DEFAULT", slo)
```
Corey-Zumar: On second thought, let's force the user to specify a "default_value" parameter in the **application** configuration (this is not needed for models) because the "default_value" python parameter does not have a default value.

Corey-Zumar (changes requested):
This is a good start. Please address these comments. Can you also commit an example configuration file for reference purposes? We'll remove the example file before merging the PR.

dcrankshaw:
jenkins ok to test

dcrankshaw:
@Corey-Zumar Let's include an example configuration file as a reference. It's always easier to copy and modify an existing template than writing your own from scratch, especially for new users.

Corey-Zumar:
@dcrankshaw Makes sense! Will make sure that gets included in the final merge

nduggirala:
I addressed the comments and added a sample configuration file.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -131,6 +117,24 @@ def start_clipper(
             logger.warning("Error starting Clipper: {}".format(e.msg))
             raise e
 
+        def parse_configuration(file):
```
Corey-Zumar: Nit: let's rename `file` to `config_path`. Also, add method documentation.

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,23 @@
+Model1:
```
Corey-Zumar: In addition to specifying applications and models for deployment, we want to be able to specify links between models and applications. Define a yaml dictionary (or equivalent) called "Links" whose key-value entries are defined as follows:

* `key = <APP_NAME>`
* `value = <MODEL_NAME>`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -131,6 +117,24 @@ def start_clipper(
...
+                with open(file, "r") as stream:
+                    config = yaml.load(stream)
+                    for name in config:
+                        obj_type = config[name]["obj_type"]
+                        input_type = config[name]["input_type"]
+                        if(obj_type == "application"):
+                            slo = config[name]["slo"]
+                            default_value = config[name]["default_value"]
+                            self.register_application(name, input_type, default_value, slo)
+                        elif(obj_type == "model"):
```
Corey-Zumar: Let's add support for specifying parameter values for the `deploy_model` function that have default values (`labels`, `num_replicas`). Any additional key-word arguments should be passed via the `**kwargs` parameter.

Corey-Zumar (changes requested):
Nice progress! This just needs a bit of additional functionality

simon-mo:
hmm... @dcrankshaw where do you think the `metric_config.yaml` fit into this? 

Should we do kubernetes style (concat yaml files so user config the app in one go) or should be have the separate files, and let users specific the config file location inside `metric_config.yaml`? 

Basically, are they parallel or parent-child relationship? 

https://github.com/ucbrise/clipper/blob/develop/monitoring/metrics_config.yaml

(Apologize if this is the wrong thread to talk about this, didn't find the corresponding issue)

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  slo: 10000
+  default_value: "not received"
+
+Model2:
+  obj_type: "model"
+  input_type: "floats"
+  version: "1.2.7"
+  image: "train-model:2"
+
+Application2:
```
simon-mo: Great work! It's a better practice to keep the key of dictionary to be static/immutable/predictable: checkout https://github.com/ucbrise/clipper/blob/develop/clipper_admin/clipper_admin/kubernetes/query-frontend-deployment.yaml

clipper_admin/clipper_admin/sample_config.yaml
dcrankshaw: Let's be more explicit with the links. Instead of making key value pairs let's structure the links as a list of objects, where each object has a `model` key and an `app` key. For example:

```
links:
  - model: <MODEL_NAME>
     app: <APP_NAME
  - model: <OTHER_MODEL_NAME>
     app: <OTHER_APP_NAME>
```

dcrankshaw:
@nduggirala This failed the format checker. Run `./bin/format_code.sh` to format

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -91,7 +93,8 @@ def start_clipper(
             compability and preserve the expected behavior of the system.
         cache_size : int, optional
             The size of Clipper's prediction cache in bytes. Default cache size is 32 MiB.
-
+        config_file: str(optional)
+            File for the configuration of the application or model
```
dcrankshaw: Can you make this description more explicit? It's for configuring the state of the Clipper cluster (models, applications, and links, not just a single app or model).

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,42 @@ def start_clipper(
             logger.warning("Error starting Clipper: {}".format(e.msg))
             raise e
 
+        def parse_configuration(config_path):
+            """ if a configuration file exists,
```
dcrankshaw: Nit: Complete sentence

dcrankshaw:
What's the status on this?

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,46 @@
+Model1:
+  obj_type: "model"
+  input_type: "integers"
+  version: "1.2.7"
+  image: "test-model:1"
+
+Application1:
+  obj_type: "application"
```
Corey-Zumar: As suggested by @simon-mo , let's define separate arrays called "Applications" and "Models". We can then remove the `obj_type` field from elements of these arrays because every entry under "Applications" will be an application object. i.e. Do the same thing for models and applications that you've done for links.

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,46 @@
...
+  slo: 10000
+  default_value: "not received"
+
+Model2:
+  obj_type: "model"
+  input_type: "floats"
+  version: "1.2.7"
+  image: "train-model:2"
+  labels: ["second model configured", "working successfully"]
+  num_replicas: 2
```
Corey-Zumar: Models should support the `batch_size` field as well. See https://github.com/ucbrise/clipper/blob/137cc852a5c9ed5cca11c6c2f89f1e708a561198/clipper_admin/clipper_admin/clipper_admin.py#L403

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,42 @@ def start_clipper(
...
+            """
+            if(config_path):
+                with open(config_path, "r") as stream:
+                    config = yaml.load(stream)
+                    for name in config:
+                        if(name == "links"):
+                            for link in config[name]:
+                                self.link_model_to_app(config[name][link]["app_name"], config[name][link]["model_name"])
+                        else:
+                            obj_type = config[name]["obj_type"]
```
Corey-Zumar: We should check that required keys are present and throw `ClipperException` errors if they are absent (https://github.com/ucbrise/clipper/blob/137cc852a5c9ed5cca11c6c2f89f1e708a561198/clipper_admin/clipper_admin/exceptions.py#L1).

Using python's inspect module (https://docs.python.org/3/library/inspect.html), you can determine the set of parameters associated with the `deploy_model`, `register_application`, and `link_model_to_app` methods, as well as which parameters have default values (see https://stackoverflow.com/questions/218616/getting-method-parameter-names-in-python).

Corey-Zumar (changes requested):
Good progress! My comments concern restructuring the yaml based on our discussion a couple of weeks ago, as well as error handling

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,69 @@ def start_clipper(
...
+
+            Parameters
+            ----------
+            config_path : str
+                Path to the configuration file.
+            """
+            if (config_path):
+                with open(config_path, "r") as stream:
+                    config = yaml.load(stream)
+                    if ("links" in config):
```
Corey-Zumar: We should process "links" last (after models and applications). The current behavior throws "application not found" exceptions because the applications specified by the links haven't been registered yet.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,69 @@ def start_clipper(
...
+                        for link in config["links"]:
+                            try:
+                                self.link_model_to_app(
+                                    config["links"][link]["app_name"],
+                                    config["links"][link]["model_name"])
+                            except ClipperException as e:
```
Corey-Zumar: This will only catch `ClipperException` objects thrown by `link_model_to_app`, but it won't catch `KeyError` exceptions in the case that fields are missing. We should also catch these exceptions and indicate to the user that required keys are missing by throwing a `ClipperException` with a message to the effect of "Required keys missing from Clipper link specification"

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,69 @@ def start_clipper(
...
+                                input_type = config["applications"][
+                                    application]["input_type"]
+                                slo = config["applications"][application][
+                                    "slo"]
+                                default_value = config["applications"][
+                                    application]["default_value"]
+                                self.register_application(
+                                    application, input_type, default_value,
+                                    slo)
+                            except ClipperException as e:
```
Corey-Zumar: Catch `KeyErrors` here too and throw an appropriate `ClipperException`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -114,6 +117,69 @@ def start_clipper(
...
+                                    num_replicas = config["models"][model][
+                                        "num_replicas"]
+                                batch_size = -1
+                                if ("batch_size" in config["models"][model]):
+                                    batch_size = config["models"][model][
+                                        "batch_size"]
+                                self.deploy_model(model, version, input_type,
+                                                  config_image, labels,
+                                                  num_replicas, batch_size)
+                            except ClipperException as e:
```
Corey-Zumar: Catch `KeyError`s and throw an appropriate `ClipperException`

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
+models:
+  Model1:
```
Corey-Zumar: This is an invalid model name:

```
clipper_admin.exceptions.ClipperException: Invalid value: Model3: a model name must be a valid DNS-1123  subdomain. It must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?\Z'
```

use `model1` (lowercase `m`) instead.

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    input_type: "integers"
+    version: "1.2.7"
+    image: "test-model:1"
+  Model2:
```
Corey-Zumar: `Model2` >> `model2`

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    input_type: "floats"
+    version: "1.2.7"
+    image: "train-model:2"
+    labels: ["second model configured", "working successfully"]
+    num_replicas: 2
+  Model3:
```
Corey-Zumar: `Model3` >> `model3`

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
+models:
+  Model1:
+    input_type: "integers"
+    version: "1.2.7"
```
Corey-Zumar: This is an invalid model version

```
clipper_admin.exceptions.ClipperException: Invalid value: 4.8.9: a model version must be a valid DNS-1123  subdomain. It must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?\Z'
```

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    image: "test-model:1"
+  Model2:
+    input_type: "floats"
+    version: "1.2.7"
```
Corey-Zumar: Invalid model version

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    image: "train-model:2"
+    labels: ["second model configured", "working successfully"]
+    num_replicas: 2
+  Model3:
+    input_type: "doubles"
+    version: "4.8.9"
```
Corey-Zumar: Invalid model version

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    default_value: "undetected"
+  Application3:
+    obj_type: "application"
+    input_type: "integers"
+    slo: 2000
+    default_value: "message not received"
+
+links:
+  link1:
+    app_name: "Application1"
```
Corey-Zumar: Let's make sure we're linking models with applications that have the same input type

clipper_admin/clipper_admin/clipper_admin.py
Corey-Zumar: Nit: `str(optional) >> str, optional`

LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
Corey-Zumar: Remove this file

clipper_admin/clipper_admin/clipper_admin.py
Corey-Zumar: Expand on this parameter documentation: 

1. Indicate that `config_file` specifies the path to a configuration file.
 
2. Explain that, if the path is present, Clipper will deploy models, register applications, and create links as specified by the file. 

3. Finally, provide the name of the sample configuration file that you wrote so that the user can reference it.

Corey-Zumar (changes requested):
Nice progress - a few comments regarding correctness and documentation improvements. 

**Please respond to each comment as you make its associated change**

Corey-Zumar:
@nduggirala What is the status of this?

nduggirala:
@Corey-Zumar I finished everything except the parameter documentation. I should be done by tomorrow

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: In Progress

LICENSE
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: In Progress

nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -94,7 +95,10 @@ def start_clipper(
             compability and preserve the expected behavior of the system.
         cache_size : int, optional
             The size of Clipper's prediction cache in bytes. Default cache size is 32 MiB.
-
+        config_file : str, optional
+            Path to a configuration file. If the path is present, Clipper will deploy models,
+            register applications, and create links as specified by the file. The file 
+            'sample_config.yaml' is an example of a configuration file.
```
dcrankshaw: Where can users find `sample_config.yaml`?

clipper_admin/clipper_admin/clipper_admin.py
dcrankshaw: Reminder to make this a complete sentence.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -117,6 +121,77 @@ def start_clipper(
             logger.warning("Error starting Clipper: {}".format(e.msg))
             raise e
 
+        def parse_configuration(config_path):
```
dcrankshaw: Change the method name to `_parse_configuration` to indicate this is a private method. Also, there is no need for this to be a nested method. Just make it a normal method for the object.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -117,6 +121,77 @@ def start_clipper(
...
+            """ if a configuration file exists,
+            parse it and accordingly deploy models and applications
+            Parameters
+            ----------
+            config_path : str
+                Path to the configuration file.
+            """
+            if (config_path):
```
dcrankshaw: Instead of checking for the existence of `config_path` here, check for it in the method that calls `parse_configuration` and only call `parse_configuration()` if `config_path` is not None.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -117,6 +121,77 @@ def start_clipper(
...
+                                    application]["input_type"]
+                                slo = config["applications"][application][
+                                    "slo"]
+                                default_value = config["applications"][
+                                    application]["default_value"]
+                                self.register_application(
+                                    application, input_type, default_value,
+                                    slo)
+                            except ClipperException as e:
+                                print("required parameters not found")
```
dcrankshaw: Use `logger.error` instead of `print`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -117,6 +121,77 @@ def start_clipper(
...
+                                raise e
+                            except KeyError, f:
```
dcrankshaw: This should be `except KeyError as f`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -117,6 +121,77 @@ def start_clipper(
...
+                                    config["links"][link]["app_name"],
+                                    config["links"][link]["model_name"])
+                            except ClipperException as e:
+                                print("required parameters not found")
+                                raise e
+                            except KeyError, f:
+                                print("required parameters not found")
+                                raise ClipperException(KeyError)
+
+        parse_configuration(config_file)
```
dcrankshaw: Check for the existence of config_file before calling parse_configuration here:

```py
if config_file is not None:
    parse_configuration(config_file)
```

clipper_admin/clipper_admin/sample_config.yaml
```diff
@@ -0,0 +1,42 @@
+models:
```
dcrankshaw: This file shouldn't be in clipper_admin. You should create an example in the [`examples/`](https://github.com/ucbrise/clipper/tree/develop/examples) directory that includes both the sample config file and an example Python file using it.

dcrankshaw (changes requested):
This is looking pretty good. Can you add a test to `integration_tests/clipper_admin_tests.py` to test this feature.

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

clipper_admin/clipper_admin/clipper_admin.py
nduggirala: Done

dcrankshaw:
@nduggirala did you forget to push your latest changes? You marked a bunch of these comments as done but haven't pushed anything.

nduggirala:
@dcrankshaw I still need to write the integration test.

clipper_admin/clipper_admin/sample_config.yaml
nduggirala: Done

nduggirala:
@dcrankshaw I wrote the integration test and added an example

shaneknapp:
test this please

rkooo567:
@simon-mo @withsmilo I think this feature can be useful in the future? Are we planning to handle this again?

simon-mo:
I agree. However this requires a bit refactoring and careful design discussion with respect to how the config.yaml should be designed.

withsmilo:
@rkooo567 , I will check this PR tonight.

simon-mo:
Closed because outdated. But config file is a good idea and we will submit a new PR based on this one.

PR #394, opened 2018-02-01
Model container testing function to clipper admin (Vanilla python)

### LINKED ISSUE #360: Adding model container testing function to clipper admin
Currently, when users use one of the model deployers to deploy a predict function to Clipper, they have no way to debug their function before deploying the model. We should add a `test_predict_function` method to each model deployer that allows users to test that their function has the correct signature and can be properly saved and loaded.

The `test_predict_function` should have the following arguments:
+ a dict request object in the same format that the query frontend expects JSON _or_ a list of dicts if the user wishes to test their batching logic behaves correctly.
+ The predict function
+ The model (for non-vanilla Python model deployers such as PyTorch or TensorFlow)
+ The model input type

The test function will first save the predict function using the appropriate model saver and then reload it to test serialization. It will then transform the JSON input into a list of the appropriate type (e.g. a list of numpy float vectors) and call the reloaded predict function, returning the results. We will need a `test_*` function for each model deployer.

Note that this feature requires an integration test as part of our CI tests to ensure that the test function and the actual query frontend maintain matching input formats.

rohsuresh:
Closes #360 
Added a function test_predict_function in clipper_admin/clipper_admin/clipper_admin.py that verifies input type and function serialization/reloading for vanilla python models. Will resolve PyTorch and TensorFlow after this has been approved.


dcrankshaw:
jenkins ok to test

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,53 @@ def stop_all(self):
...
+    def test_predict_function(self, query, func, input_type):
+        """Tests that the user's function has the correct signature and can be properly saved and loaded.
+
+        The function should take a dict request object like the query frontend expects JSON, 
+        the predict function, and the input type for the model.
+
+        Parameters
+        ----------
+        query: JSON or list of dicts
+            Input that the user sends to the query frontend
```
dcrankshaw: "Inputs to test the prediction function on."

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,53 @@ def stop_all(self):
...
+        func: function
+            Predict function that the user's model is using
```
dcrankshaw: "Predict function to test."

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,53 @@ def stop_all(self):
...
+                if type(x) != int:
+                    return "Invalid input type"
+        if input_type == "doubles":
+            for x in flattened_data:
+                if type(x) != float:
+                    return "Invalid input type"
+        if input_type == "strings":
+            for x in flattened_data:
+                if type(x) != str:
+                    return "Invalid input type"
```
dcrankshaw: Can you add checks for the other two accepted input types: (floats and bytes)

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,53 @@ def stop_all(self):
...
+        s = six.StringIO()
+        c = CloudPickler(s, 2)
+        c.dump(func)
+        serialized_func = s.getvalue()
+        reloaded_func = pickle.loads(serialized_func)
+
+        try:
+            assert reloaded_func
+        except AssertionError as e:
+            return 'Function does not properly serialize and reload'
```
dcrankshaw: Let's log this error as well `logger.error("Function ...")`

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,53 @@ def stop_all(self):
...
+
+        numpy_data = list(np.array(x) for x in query_data)
```
dcrankshaw: The only tricky thing here is to ensure that the numpy array as the right dtype. The mapping is defined [here](https://github.com/ucbrise/clipper/blob/97ef10bb1bf03178d8c7d3d261fe175b5e3bbe72/containers/python/rpc.py#L71-L81).

dcrankshaw (changes requested):
Can you delete all the extra Python files that got committed?

This implementation looks good! The only thing missing now is an integration test. Add a test to [`clipper_admin_tests.py`] that deploys a predict function (using deploy python closure) that sums the elements in an input and check to make sure that your newly added test function returns the same result as the deployed model.

dcrankshaw:
jenkins add to whitelist

dcrankshaw:
Jenkins is failing because of the format checker. Run `./bin/format_code.sh` to format.

clipper_admin/clipper_admin/deployers/python.py
```diff
@@ -4,6 +4,8 @@
 import os
 import posixpath
 import shutil
+import pickle
+import numpy as np
```
dcrankshaw: Revert the changes to this file

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,68 @@ def stop_all(self):
...
+
+        Parameters
+        ----------
+        query: JSON or list of dicts
+            Inputs to test the prediction function on.
+        func: function
+            Predict function to test.
+        input_type: str
+            The input_type to be associated with the registered app and deployed model.
+            One of "integers", "floats", "doubles", "bytes", or "strings".
```
dcrankshaw, line 1210: Can you add an example in the method comment?

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,68 @@ def stop_all(self):
...
+        """
+        query_data = list(x for x in list(query.values()))
+        
+        if type(query_data[0][0]) == list:
+            query_data = query_data[0]
```
dcrankshaw: What are you checking for here?

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1187,3 +1190,68 @@ def stop_all(self):
...
+        ----------
+        query: JSON or list of dicts
+            Inputs to test the prediction function on.
+        func: function
+            Predict function to test.
+        input_type: str
+            The input_type to be associated with the registered app and deployed model.
+            One of "integers", "floats", "doubles", "bytes", or "strings".
+        """
+        query_data = list(x for x in list(query.values()))
```
dcrankshaw: What's going on here?

dcrankshaw: You should be checking the JSON key as well, to ensure that their input is properly formatted

integration-tests/clipper_admin_tests.py
```diff
@@ -343,6 +345,20 @@ def predict_func(inputs):
...
+        def predict_func(xs):
+            return [sum(x) for x in xs]
+
+        deploy_python_closure(self.clipper_conn, name="sum-model", version=1, input_type="doubles", func=predict_func)
+        self.clipper_conn.link_model_to_app(app_name="hello-world", model_name="sum-model")
+        
+        headers = {"Content-type": "application/json"}
+        test_input = list(np.random.random(10))
+        pred = requests.post("http://localhost:1337/hello-world/predict", headers=headers, data=json.dumps({"input": test_input})).json()
+        test_predict_result = test_predict_function(self.clipper_conn, query={"input": test_input}, func=predict_func, input_type="doubles")
```
dcrankshaw: You should call the function like this `self.clipper_conn.test_predict_function(query=...)`, not by passing in `self.clipper_conn` as an argument.


Can you add a second test prediction that uses the `batch_predict` interface? e.g.

```py
batch_input = [list(np.random.random(10)) for _ in range(4)]
batch_pred = requests.post("http://localhost:1337/hello-world/predict", headers=headers, data=json.dumps({"input_batch": batch_input})).json()
test_batch_predict_result = self.clipper_conn.test_predict_function(query={"input_batch": test_input}, func=predict_func, input_type="doubles")
```

dcrankshaw (changes requested):
Looks good. Just a few minor changes.

clipper_admin/clipper_admin/clipper_admin.py
rohsuresh: I'm checking the nesting of the query data, whether it is a list or a list of lists.

clipper_admin/clipper_admin/clipper_admin.py
rohsuresh: Assuming that the query is in a JSON/dict structure, I'm getting the values of each key. I'm then checking to make sure the values are of the right input_type before then converting it into the respective numpy array of right dtype. What formatting of the key input needs to be checked?

dcrankshaw: If the user wants to provide a single input, the key should be `"input"`. If they want to provide a list of inputs, the key should be `"input_batch"`. See [the website](http://clipper.ai/tutorials/basic_concepts/) for details.

rohsuresh: Got it. Thanks for the clarification.

rohsuresh:
Any ideas on why Jenkins is still failing after running the format script?

integration-tests/clipper_admin_tests.py
```diff
@@ -343,6 +344,49 @@ def predict_func(inputs):
...
+        def predict_func(xs):
+            return [sum(x) for x in xs]
+
+        deploy_python_closure(
+            self.clipper_conn,
+            name="sum-model",
+            version=1,
+            input_type="doubles",
+            func=predict_func)
+        self.clipper_conn.link_model_to_app(
```
dcrankshaw, line 363: You need to register the "hello-world" app before you can link to it

integration-tests/clipper_admin_tests.py
```diff
@@ -343,6 +344,55 @@ def predict_func(inputs):
...
+            app_name="hello-world", model_name="sum-model")
+
+        headers = {"Content-type": "application/json"}
+        test_input = list(np.random.random(10))
+        pred = requests.post(
```
dcrankshaw: Set the address the way we set it here: https://github.com/ucbrise/clipper/blob/f9a19d142ce3a0c70aab3386fdd9b055e05b9599/integration-tests/clipper_admin_tests.py#L438-L440

dcrankshaw:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw (approved):
lgtm

PR #395, opened 2018-02-01
Added functionality for single input models

### LINKED ISSUE #285: Model containers should have a choice of running in batch or single-input mode
Currently, we expect all deployed models to be a function from a list of inputs to a list of outputs. This allows models the ability to do batching, but it makes the function signature of models more complicated. For simple models that want to operate on a single-input at a time, users have to remember to wrap the input in for-loop.

Instead, we should support deploying two types of models: batch and single-input. Batch models are still given the full available batch at once. Single-input models would be a function from a single input to a single output, and the RPC server embedded in the model container would call the model in a for-loop transparently to the user.

This should make the model deployment process simpler in many common cases.

amogkam:
Model containers now have a choice of running in batch mode or single input mode. If single input mode is chosen, the RPC server wraps the input in a for loop to allow for predictions on a single input at a time.

Fixes #285

### DIFF
--- a/clipper_admin/clipper_admin/clipper_admin.py
+++ b/clipper_admin/clipper_admin/clipper_admin.py
@@ -237,7 +237,8 @@ class ClipperConnection(object):
                                labels=None,
                                container_registry=None,
                                num_replicas=1,
-                               batch_size=-1):
+                               batch_size=-1,
+                               batch_mode=True):
         """Build a new model container Docker image with the provided data and deploy it as
         a model to Clipper.
 
@@ -282,10 +283,14 @@ class ClipperConnection(object):
             :py:meth:`clipper.ClipperConnection.set_num_replicas`.
         batch_size : int, optional
             The user-defined query batch size for the model. Replicas of the model will attempt
-            to process at most `batch_size` queries simultaneously. They may process smaller 
+            to process at most `batch_size` queries simultaneously. They may process smaller
             batches if `batch_size` queries are not immediately available.
             If the default value of -1 is used, Clipper will adaptively calculate the batch size for individual
             replicas of this model.
+        batch_mode : bool, optional
+            Specifies whether the model is a batch model or a single input model. If it is a single input model
+            (batch_mode is False), the RPC server embedded in the model calls the model in a for loop to allow the model to operate
+            on a single input at a time. By default, batch_mode is set to True.
         Raises
         ------
         :py:exc:`clipper.UnconnectedException`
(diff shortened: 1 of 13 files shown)


Corey-Zumar:
@dcrankshaw If we choose to support the single-input mode for all models, we'll have to make analogous RPC changes that affect the Java, C++, and C++ for R containers. However, if we limit single-input mode to deployer-friendly models, no changes to the RPC layer or the core system are required. In the case of R/Python-deployable functions, we can wrap the user-supplied function in a "for" loop and achieve the desired result. 

In my view, if users aren't using model deployers and are opting to write their own containers, we can reasonably require them to write their `predict` functions for the batch mode specification.

What are your thoughts?

dcrankshaw (changes requested):
This looks great. Can you write a test in `clipper_admin_tests.py` to test this? The test should deploy a model that takes a single input (as opposed to a list of inputs) and deploy it with `batch_mode=False` and then test to make sure we get correct predictions back.

dcrankshaw:
@Corey-Zumar I don't think that the overhead of modifying the C++ and Java RPC containers is very high. This is basically a 2 line code change in `rpc.py` that checks for the `CLIPPER_BATCH_MODE` environment variable and if it is present and set to false calls the user's predict function in a for loop.  I'm open to being persuaded, but this just seems like a pretty minor change.

Corey-Zumar:
@dcrankshaw It's a ~10 line change for each RPC system that will necessitate a separate integration test for all 4 RPC implementations. I would also argue that this implementation diminishes the abstraction between the RPC layer and the model's predict function. Where the model could previously treat `predict` as a black box from which to get outputs, it is now tasked with the responsibility of determining how `predict` operates.

Additionally, implementing this change at the container level means that the user is now responsible for knowing whether each model they are deploying is a "batch mode" or "single input" model, introducing additional overhead. In contrast, if the user deploys a "single input" prediction function using the wrapping method, a batch mode container will be built that can be used directly with `build_model` and `build_and_deploy_model`.

Ultimately, it seems much more straightforward to make a change consisting of ~20 total lines that targets the majority of use cases and does not require the user to keep track of additional metadata.

dcrankshaw:
Alright those are fair points. I'm now wondering if this feature even makes sense in the first place. Given how integral batching is to the system, perhaps it would be better to force developers to write their models to process a batch at a time. The ability to test models before they are deployed that will be added with #394 should help developers catch batch processing related errors before the model is deployed.

Corey-Zumar:
@dcrankshaw Agreed. If we can use pre-deployment function checking to guide the user towards a correct, batching-enabled solution, then this feature doesn't seem particularly useful.

dcrankshaw:
Alright I'm going to put this on hold. Sorry for wasting your time @amogkam, your implementation was great! This was totally my fault. We'll get you working on something cool.

amogkam:
@dcrankshaw No problem- this was a great way for me to get familiar with the codebase. I will talk to you on Friday for a new task.

PR #402, opened 2018-02-05
Metric Monitoring: Add to/Delete from the metric config of Prometheus

withsmilo:
* AS-IS
Add only a new model container name to the metric config
 using update_metric_config() in docker_metric_utils.py.

* CHANGED
We might need to delete some model containers from the metric config
if the number of model replicas is decreased. So, I did like below,
  * Rename update_metric_config() to add_to_metric_config()
  * Add delete_from_metric_config()


simon-mo:
Jenkins ok to test

simon-mo:
Great catch! This patch LGTM. Once @dcrankshaw approve, we are good to merge.

dcrankshaw (approved):
lgtm

PR #390, opened 2018-02-01
Added MXNet Deployment Capability

chester-leung:
Added capability to save/load and deploy MXNet models.
Added mxnet.py in clipper_admin/clipper_admin/deployers/ , mxnet_container.py in containers/python/ , and an integration test for MXNet. Also added necessary files to build a docker image from MXNet model.

Fixes: #364

PR #404, opened 2018-02-05
Added MXNet Deployment Capability (replaces #390)

dcrankshaw:
Sorry @chester-leung ran into some git issues when fixing the format checker and had to close #390. This is a replacement.


bin/build_docker_images.sh
```diff
@@ -254,11 +254,13 @@ build_images () {
     create_image py-rpc RPCDockerfile $public
     create_image sum-container SumDockerfile  $private
     create_image noop-container NoopDockerfile $public
+    create_image mxnet-container MXNetContainerDockerfile $public
     create_image python-closure-container PyClosureContainerDockerfile $public
     create_image pyspark-container PySparkContainerDockerfile $public
     create_image tf_cifar_container TensorFlowCifarDockerfile $public
     create_image tf-container TensorFlowDockerfile $public
     create_image pytorch-container PyTorchContainerDockerfile $public
+    create_image mxnet-container MXNetContainerDockerfile $public
```
dcrankshaw, line 263: Remove one of these lines

dockerfiles/ClipperTestsDockerfile
```diff
@@ -66,6 +66,7 @@ RUN cd /clipper/src/libs/spdlog \
     && git apply ../patches/make_spdlog_compile_linux.patch
 
 
-ENTRYPOINT ["/clipper/bin/ci_checks.sh"]
+# ENTRYPOINT ["/clipper/bin/ci_checks.sh"]
+ENTRYPOINT ["bash"]
```
dcrankshaw, line 70: Sorry this was my fault, but delete line 70 and uncomment line 69

dockerfiles/MXNetContainerDockerfile
```diff
@@ -0,0 +1,20 @@
...
+      && apt-get update --fix-missing \
+      && apt-get install -yqq -t jessie-backports openjdk-8-jdk \
+      && conda install -y --file /lib/python_container_conda_deps.txt \
+      && pip install mxnet==1.0.0 \
+      && apt-get install -y graphviz \
+      && pip install graphviz \
+      && conda install -c anaconda cloudpickle=0.5.2
+
+
+COPY containers/python/mxnet_container.py containers/python/mxnet_container_entry.sh containers/python/train.data /container/
```
dcrankshaw, line 16: Why are you copying `train.data` into the container?

dcrankshaw (commented):
@chester-leung added a review

dockerfiles/MXNetContainerDockerfile
chester-leung: When I load the serialized model I need to bind it with some data. I just used train.data. Because we load the model in the container I had to copy train.data into the container to access it - I was having some problems accessing it in its original location in integration_tests/data. If this is causing problems I can bind the loaded model with some dummy data and see if that works.

PR #405, opened 2018-02-06
Added MXNet Deployment Capability (replaces #390)

chester-leung:


bin/build_docker_images.sh
```diff
@@ -254,11 +254,13 @@ build_images () {
     create_image py-rpc RPCDockerfile $public
     create_image sum-container SumDockerfile  $private
     create_image noop-container NoopDockerfile $public
+    create_image mxnet-container MXNetContainerDockerfile $public
```
dcrankshaw: Delete this line

.gitignore
```diff
@@ -19,3 +19,4 @@ dependency-reduced-pom.xml
 *.iml
 
 docker-compose.yml
+venv/
```
Corey-Zumar: Revert this change

bench/bench_init.py
```diff
@@ -35,8 +35,8 @@ def run(self):
 
     def get_cifar_data(self):
         train_x, train_y = cifar_utils.filter_data(*cifar_utils.load_cifar(
-            self.cifar_dir_path, cifar_filename="cifar_train.data",
-            norm=False))
+            self.cifar_dir_path, cifar_filename="cifar_train.data", norm=False)
```
Corey-Zumar, line 38: Revert

containers/python/mxnet_container.py
```diff
@@ -0,0 +1,122 @@
...
+
+# load mxnet model from serialized dir
+def load_mxnet_model(model_path):
+    data_iter = mx.io.CSVIter(
+        data_csv="train.data", data_shape=(785, ), batch_size=1)
+
+    sym, arg_params, aux_params = mx.model.load_checkpoint(
+        prefix=model_path, epoch=0)
+    mxnet_model = mx.mod.Module(symbol=sym)
+    mxnet_model.fit(
```
Corey-Zumar: The user's model should be fit (trained) prior to deployment. We should not be calling any additional training functions within the container. Rather, the model should be loaded from its saved checkpoint files (symbol and parameter). Let me know if you have any questions.

containers/python/mxnet_container.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def load_predict_func(file_path):
+    with open(file_path, 'r') as serialized_func_file:
+        return cloudpickle.load(serialized_func_file)
+
+
+# load mxnet model from serialized dir
+def load_mxnet_model(model_path):
+    data_iter = mx.io.CSVIter(
```
Corey-Zumar: The container should not be loading any training data or doing any training. See next comment.

dockerfiles/MXNetContainerDockerfile
```diff
@@ -0,0 +1,20 @@
...
+      && apt-get update --fix-missing \
+      && apt-get install -yqq -t jessie-backports openjdk-8-jdk \
+      && conda install -y --file /lib/python_container_conda_deps.txt \
+      && pip install mxnet==1.0.0 \
+      && apt-get install -y graphviz \
+      && pip install graphviz \
+      && conda install -c anaconda cloudpickle=0.5.2
+
+
+COPY containers/python/mxnet_container.py containers/python/mxnet_container_entry.sh containers/python/train.data /container/
```
Corey-Zumar: Remove the `containers/python/train.data` file from the container

dcrankshaw:
jenkins add to whitelist

Corey-Zumar (changes requested):
This is a good start. However, the container should not be performing any model-specific training or loading any additional data beyond the serialized model structure and parameters. The user is responsible for training the model before deploying it; we should only serialize the provided model in the deployer and deserialize it within the container.

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,207 @@
...
+MXNET_MODEL_RELATIVE_PATH = "mxnet_model"
+
+
+def create_endpoint(
+        clipper_conn,
+        name,
+        input_type,
+        func,
+        mxnet_model,
+        data_shapes,
```
Corey-Zumar: Would the variable name `mxnet_input_shape` be more appropriate here? Please add documentation for this variable under the `Parameters` section below (and explain why it's necessary).

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def deploy_mxnet_model(
+        clipper_conn,
+        name,
+        version,
+        input_type,
+        func,
+        mxnet_model,
+        data_shapes,
```
Corey-Zumar: Rename variable and add documentation

dockerfiles/ClipperTestsDockerfile
```diff
@@ -11,7 +11,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install --fix-missi
       && apt-get install -y docker-ce
 
 RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh \
-    && wget --quiet 'https://repo.continuum.io/archive/Anaconda2-4.2.0-Linux-x86_64.sh' -O /tmp/anaconda.sh \
+    && wget --quiet 'https://repo.continuum.io/archive/Anaconda2-5.0.1-Linux-x86_64.sh' -O /tmp/anaconda.sh \
```
Corey-Zumar, line 14: Is this change necessary?

dockerfiles/ClipperTestsDockerfile
```diff
@@ -66,6 +66,7 @@ RUN cd /clipper/src/libs/spdlog \
     && git apply ../patches/make_spdlog_compile_linux.patch
 
 
-ENTRYPOINT ["/clipper/bin/ci_checks.sh"]
+# ENTRYPOINT ["/clipper/bin/ci_checks.sh"]
```
Corey-Zumar: Revert

Corey-Zumar (changes requested):
Nice job! Just cosmetic / documentation-related comments

clipper_admin/clipper_admin/deployers/mxnet.py
dcrankshaw: I think the variable name `mxnet_data_shapes` would be best, to match the naming conventions of MXNet. In the comment about why this is necessary, add a link to the relevant portion of the mxnet documentation. Here's the relevant markup for adding an external link in Python documentation: https://devguide.python.org/documenting/#external-links

dockerfiles/ClipperTestsDockerfile
dcrankshaw: This was my change.

chester-leung: @dcrankshaw edited this file when trying to get Jenkins to build.

dockerfiles/ClipperTestsDockerfile
chester-leung: Done.

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: Done.

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: Done.

bench/bench_init.py
dcrankshaw: This is needed to fix formatting issues we've been having

dockerfiles/ClipperTestsDockerfile
```diff
@@ -67,5 +67,6 @@ RUN cd /clipper/src/libs/spdlog \
 
 
 ENTRYPOINT ["/clipper/bin/ci_checks.sh"]
+ENTRYPOINT ["bash"]
```
dcrankshaw: Delete this line

dcrankshaw (changes requested):
Couple more cleanup items

.gitignore
chester-leung: Done.

dockerfiles/MXNetContainerDockerfile
chester-leung: Done.

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: @dcrankshaw If I wanted to add the web address of the relevant MXNet documentation, should I separate the address and the rest of the description for `mxnet_data_shapes` with an empty line?

dockerfiles/ClipperTestsDockerfile
chester-leung: Done.

chester-leung:
The latest update (~Wed 11:45 pm) contains changes Dan pushed to the same branch I was working on. The only changes I made were in .gitignore, dockerfiles/ClipperTestsDockerfile, and containers/python/mxnet_container.py.

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,212 @@
...
+        The name to be assigned to both the registered application and deployed model.
+    input_type : str
+        The input_type to be associated with the registered app and deployed model.
+        One of "integers", "floats", "doubles", "bytes", or "strings".
+    func : function
+        The prediction function. Any state associated with the function will be
+        captured via closure capture and pickled with Cloudpickle.
+    mxnet_model : mxnet model object
+        The MXNet model to save.
+    mxnet_data_shapes : list(int)
```
Corey-Zumar, line 48: Make this documentation consistent with the `mxnet_data_shapes` parameter documentation for `deploy_mxnet_model`

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,212 @@
...
+        List of integers representing the dimensions of data used for model prediction.
+        Required because loading serialized MXNet models involves binding, which requires
+        the shape of the data used to train the model.
+        https://mxnet.incubator.apache.org/api/python/module.html#mxnet.module.BaseModule.bind
+
+        NOTE: Clipper may provide the model with variable size input batches. Because MXNet can't
+              handle variable size input batches, we recommend setting batch size for input data
+              to 1, or dynamically reshaping the model with every prediction based on the current
```
Corey-Zumar: Can we do this dynamic reshaping automatically for the user?

Corey-Zumar (changes requested):
Just a minor documentation change and a question about reshaping. Should be good to go after this!

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: @dcrankshaw and I had a conversation about this. Because MXNet cannot handle variable size batches, but Clipper varies batch size for optimization purposes, it would just be easier to have the user input 1 as the batch size for all MXNet models. However, I don't think we thought about automatically reshaping according to what Clipper has been doing -- that's a good thought. 
How difficult would it be to get the input batch size before each prediction? 
If it's not that difficult, maybe we can reshape the existing model according to the next batch size (and next input dimensions).

Corey-Zumar: We can obtain the batch size directly from the `inputs` parameter to the `predict` method as `batch_size = len(inputs)`. We should then be able to use this to reshape before evaluating the model, right?

chester-leung: I think we would also have to get the dimensions of the data itself, e.g. (1,3,224,224). Will we will be able to read these dimensions from the `inputs` parameter?

Corey-Zumar: Ah, I see. Each input is a numpy array of datatype `float`. `np.array(inputs).shape` should do it

chester-leung: In that case, I think we should be able to do it. There's a reshape() function that can reshape the model. @dcrankshaw am I missing something here? Or will this work?

Corey-Zumar: I'm confident that we'll be able to do this. Currently, the user specifies a data shape at deployment as a tuple consisting of:

1. Batch size (`deployed_shape[0]`)
2. Item dimensions (`deployed_shape[1:]`)

We can therefore compute the batch size from the inputs array using the method described above: `batch_size = len(inputs)` and set the model data shape such that 
`new_shape[0] = batch_size` 
`new_shape[1:] = deployed_shape[1:]`

Can you try this?

dcrankshaw: The big question in my mind was about how slow re-shaping the model is. It might be an expensive operation to do for every batch. For the sake of simplicity, let's get do the automatic reshaping for the user for now. We can always optimize this later (or people can write their own model containers).

dcrankshaw:
jenkins retest this please

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: I think it'd be best to reshape the model in the `predict` function, based off the input data passed into the function, so that the user can adjust the dimensions every prediction (which presumably uses different input data each time). Does the user write his own `predict` function? It seems that in the integration test we wrote our own `predict` function and called it after loading a serialized model.

dcrankshaw: Hmm @chester-leung you raise a good point. I wonder if the right thing to do is let the user do the model re-shaping if they want from within the predict function. Automatically re-shaping for them could be confusing. That lets them choose whether to accept the overhead of reshaping in order to make batch predictions, or whether to make single predictions in a loop.

@Corey-Zumar what do you think? The idea of modifying the model internally without the users knowledge seems like a potential source of confusion.

Corey-Zumar: @dcrankshaw Relying on user specification of reshaping necessitates the extra "batch_size" flag specification to avoid incorrect behavior. Automatically handling it provides the benefits of adaptive batching and reduces the likelihood of user error. @chester-leung can you time the reshaping procedure for multiple iterations (~200) with the integration test model? If the overhead is low, I am strongly in favor of reshaping automatically.

chester-leung: Do we write a `predict` function for the user? Or does the user have to write it himself? If he writes it himself similar to how we wrote it in the integration test, i.e. with data as the input, I think there's no way for us to automatically reshape -- we need the input data to know its dimensions. Since the input data is only passed into the `predict` function, and the user writes that function himself, we cannot automatically reshape.

If I'm mistaken, @Corey-Zumar do we have 200 sample data lying around? I could just reshape the model 200 times to the same initial dimensions (basically call `reshape` but not actually do any reshaping), but I'm thinking that there could be more overhead if we drastically change the model's dimensions each time we call `predict`.

Corey-Zumar: If the user writes the `predict` function themselves, we will obtain the deserialized prediction function in the container as `self.predict_fn`. We also obtain the deserialized model as `self.model`. When we receive an input, we can then mutate the model before calling `self.predict(model, input)`.

I'd like to measure the reshaping overhead. Here's the procedure:

1. Fix an initial data shape (you can use the same one that's specified in the integration test)
2. 
```
for batch_size in range(1, 64): 
    for trial in range(10):
         // Time the latency associated with reshaping the data to support a batch size of "batch_size"

You can do this locally without creating and deploying a container. Let me know if this is still unclear.

Comment (dcrankshaw) [reply clipper_admin/clipper_admin/deployers/mxnet.py]: @Corey-Zumar I'm saying don't do anything about the varying batch sizes for the user. Just provide them the model as they serialized it and provide them batches of inputs. This is what we do for every other model anyway. There are lots of other cases where the users have to account for variable-sized batches in their predict function code anyway. The only reason we extract out the mxnet model to be serialized separately is because it cannot be pickled. We are just trying to provide the user the ability to deploy predict functions that contain mxnet models.

I'm strongly against transparently modifying the model behind the users back. This makes their predict function harder to test and debug, and can make it harder to handle edge cases. E.g. what if they want to handle variable sized inputs as well as variable sized batches. I think we should maintain the contract we hold with other model deployers and simply stick to automatically serializing and deserializing their models.

Comment (Corey-Zumar) [reply clipper_admin/clipper_admin/deployers/mxnet.py]: @dcrankshaw Awesome, that makes sense.

**Review by Corey-Zumar — APPROVED** [review]
LGTM. Nice job @chester-leung !

File: clipper_admin/clipper_admin/deployers/mxnet.py [hunk 14]
```diff
@@ -0,0 +1,220 @@
...
+        captured via closure capture and pickled with Cloudpickle.
+    mxnet_model : mxnet model object
+        The MXNet model to save.
+    mxnet_data_shapes : list(int)
+        List of integers representing the dimensions of data used for model prediction.
+        Required because loading serialized MXNet models involves binding, which requires
+        the shape of the data used to train the model.
+        https://mxnet.incubator.apache.org/api/python/module.html#mxnet.module.BaseModule.bind
+
+        NOTE: Clipper may provide the model with variable size input batches. Because MXNet can't
```
dcrankshaw: This is great. Can you make it a separate Notes section (below the Parameters section) so that it gets formatted properly with pydoc. Check out this example: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
```
dcrankshaw: Same comment as above about creating a separate Notes section

clipper_admin/clipper_admin/deployers/mxnet.py
```diff
@@ -0,0 +1,220 @@
...
+        mxnet_model.fit(data_iter, num_epoch=1)
+
+        deploy_mxnet_model(
+            clipper_conn,
+            name="example",
+            version = 1,
+            input_type="doubles",
+            func=predict,
+            mxnet_model=model,
+            mxnet_data_shapes=data_shape)
```
dcrankshaw, line 197: Where is the `data_shape` variable defined in this example?

dump.rdb
```diff
@@ -0,0 +1 @@
+REDIS0006�ܳC�Z��V
```
dcrankshaw: Delete this file

dcrankshaw (changes requested):
@chester-leung almost there. Just a couple documentation cleanup comments.

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: Oops, forgot to define. Just added in.

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: Done

clipper_admin/clipper_admin/deployers/mxnet.py
chester-leung: Done

dcrankshaw (approved):
LGTM!

Issue #138, opened 2017-04-26
Make Clipper compatible with Python 3


lucasmoura:
While running the tutorials, I realized that Clipper is not supporting Python 3. It will be necessary to update the python files to add support to Python 3 while also maintaining support for Python 2.

It will also be necessary to check the dependencies of the project. For example, [fabric](https://github.com/fabric/fabric) is only available for Python 2.

lucasmoura:
I will be working on that issue and I will report any problems that I can find

dcrankshaw:
This is still a relevant issue. As we refactor the Clipper admin to make it more modular and add support for Kubernetes, we will remove the dependency on fabric which is not Python 3 compatible.

simon-mo:
Hi @jwirjo 

The latest development is just to add version specification in package dependency for pip packaging in pull request #311. You just need to add these files to your branch: https://github.com/ucbrise/clipper/pull/311/files

Here’s few things I researched that might be helpful:
* A great resource:
    * Instagram’s 2to3 lessons: https://youtu.be/66XoCk79kjM?t=1487
    * Use ensure_byte(), ensure_string(), etc, abstraction. 
* Don’t forget to update integration tests for python2 and make separate (maybe?) case for python3. https://github.com/ucbrise/clipper/tree/develop/integration-tests
* The byte string conversion commonly appears in:
    * RPC.py
    * Python Model Deployer:
        * In clipper_admin/clipper_admin/deployers/python.py, add sys.version specification.

jwirjo:
1.) Make clipper_admin Python 3 compatible.
2.) Make rpc.py Python 3 compatible. 
3.) Make Python 3 version of Python Closure Dockerfile & RPC Dockerfile. (make sure all dependencies are python3 compatible)
4.) Modify integration tests: clipper_admin_tests.py

simon-mo:
Just reminded me:
- yapf python formatter and linter expect different syntax behavior for different python version. (Something I struggled a lot in PR #356) If we are doing this, we should all format the code using same python version (since bin/format_code.sh) use the python that’s in your PATH.

dcrankshaw:
@jwirjo you may find this guide on creating packages for PyPy helpful when converting the Python package to be both 2 and 3 compatible:

https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/

dcrankshaw:
[...]

PR #409, opened 2018-02-09
Py 3 RPC implementation + Clipper Admin compatibility with Python 3

jwirjo:
- RPC implementation now compatible with python 3
- Clipper Admin also compatible with python 3
Issue #138


dcrankshaw:
jenkins add to whitelist

dcrankshaw:
@jwirjo What's the difference between this PR and #416?

dcrankshaw:
Dup of #416

PR #410, opened 2018-02-12
Explicitly install redis-server when building test container

Corey-Zumar:


dockerfiles/ClipperTestsDockerfile
```diff
@@ -10,6 +10,9 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install --fix-missi
       && apt-get update \
       && apt-get install -y docker-ce
 
+# install redis-server
+RUN apt-get install -y redis-server
```
dcrankshaw: Instead of making a new layer why don't you just add it to the list of deps installed in lines 5-7

Corey-Zumar: I was confused by the `install docker` header, but it appears that we're doing more than installing docker in that section. Updated the comment and moved the installation into that layer.

dcrankshaw (approved):
lgtm

PR #417, opened 2018-02-24
Container Manager Wait Until Ready

chester-leung:
The changes force the `DockerContainerManager` and the `KubernetesContainerManager` to sleep until the container/deployment is ready.


.gitignore
```diff
@@ -18,4 +18,6 @@ _build/
 dependency-reduced-pom.xml
 *.iml
 
+k8s_starter.py
```
dcrankshaw: Remove this

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -237,8 +241,17 @@ def set_num_replicas(self, name, version, input_type, image, num_replicas):
                     name=name,
                     version=version,
                     missing=(num_missing)))
+
+            model_container_names = []
             for _ in range(num_missing):
-                self._add_replica(name, version, input_type, image)
+                name = self._add_replica(name, version, input_type, image)
+                model_container_names += [name]
```
dcrankshaw: Change this to `model_container_names.append(name)`. The way you're doing it right now constructs a new list in each iteration.

dcrankshaw (changes requested):
Not sure why that test is failing, but other than that this looks good. Just some minor comments.

.gitignore
chester-leung: Done

dcrankshaw: It doesn't look like this was removed?

.gitignore
```diff
@@ -18,4 +18,6 @@ _build/
...
+integration-tests/replica_test.py
```
dcrankshaw: Remove this.

dcrankshaw (changes requested):
This is good to go once you remove the changes to .gitignore

.gitignore
chester-leung: Oops. Must have forgotten to save. Sorry about that.

dcrankshaw (approved):
lgtm

PR #418, opened 2018-02-25
results of profiling JSON parsing latency

ryanhoque:
Modified src/frontends/README.md with results of measuring latency in `get_float_array()` and `get_float_arrays()` in `json_util.cpp` from Clipper REST API calls


dcrankshaw (changes requested):
This looks good. Can you change the units of the input sizes to bytes instead of image dimension? That will be a little more universal. So just square everything and multiply by 4.

dcrankshaw:
jenkins add to whitelist

Corey-Zumar:
@ryanhoque Awesome! Do we have memory allocation latencies as well?

Corey-Zumar (dismissed):
Just waiting on memory allocation latencies as well. These results are great so far!

Corey-Zumar:
@ryanhoque Awesome. This confirms that, relative to JSON parsing overhead, memory allocation latency is negligible. It also provides a valuable quantification of JSON parsing overhead associated with common workloads. Nice job! I'll reach out this week regarding next steps for developing an alternative, higher-performance RPC layer that does not depend on the JSON format.

@dcrankshaw I'm leaving this open in case you have additional comments and feedback based on these results. It can be closed if not.

dcrankshaw (changes requested):
@ryanhoque can you add change the table headers to include  the word "latency" in the relevant columns and add units? E.g. "mean latency (ms)", "standard deviation latency (ms)" etc.

dcrankshaw (approved):
lgtm

PR #420, opened 2018-02-25
[WIP] Query frontend fault tolerance

rbala19:
@dcrankshaw 

Here are the proposed changes for implementing query-frontend fault tolerance. These changes include polling for query-frontend downtime within kubernetes_container_manager.py and automatically redeploying any saved models to the newly deployed query-frontend.


Jeffwan:
Question: can you guarantee all query frontend instance have all deployed model instances? 
Once model is deployed, it is looking for CLIPPER_IP "query-frontend". I think models will connect to different QP pods so that every QP pod/instance has different model container associated? 

In that case, upper level prediction request might hit the QP which doesn't bind requested model. 

In your proposal, seems you assume all model containers will go to same QP instance behind QP service and this is the only one serving traffic. All other instances are just for fault tolerant? they are standby nodes and not serving traffic?

dcrankshaw:
@Jeffwan That's a really good point. This PR will actually bypass the Kubernetes service IP and provide the physical IP address of a specific query frontend replica to each container. Of course, in this case we need to manage failover ourselves in this case, because if a query frontend pod dies and is restarted by K8s, it may be assigned a new IP address. That's the reason this PR ended up being a little complicated.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -13,6 +13,8 @@
 import six
 
 from .container_manager import CONTAINERLESS_MODEL_IMAGE
+from .kubernetes.kubernetes_container_manager import KubernetesContainerManager
+from .docker.docker_container_manager import DockerContainerManager
```
dcrankshaw: Remove these imports

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1073,6 +1080,41 @@ def get_query_addr(self):
             raise UnconnectedException()
         return self.cm.get_query_addr()
 
+    @staticmethod
+    def upload_model_data(model_data_list):
```
dcrankshaw: Use this method instead of the version in the containermanagers, and it should not be static.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1073,6 +1080,41 @@ def get_query_addr(self):
...
+
+    def alert_management_upon_start(self):
+        url = "http://{}/begin_polling".format('127.0.0.1:5000')
+        # url = "http://{}/begin_polling".format(model_data['admin_addr'])
+
+        type = None
+        if isinstance(self.cm, KubernetesContainerManager):
+            type = "kubernetes"
+        elif isinstance(self.cm, DockerContainerManager):
+            type = "docker"
```
dcrankshaw: Instead of doing the comparison here, add a method to `ContainerManager` called something like `get_container_manager_type()` and have docker return "docker" and K8s return "kubernetes". This lets the implementations self-identify and makes Clipper admin easier to extend.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -1073,6 +1080,41 @@ def get_query_addr(self):
...
+            url = "http://{}/get_query_to_model_mapping".format('127.0.0.1:5000')
+            # url = "http://{}/get_query_to_model_mapping".format(model_data['admin_addr'])
+
+            r = requests.post(url, json=model_mapping)
+
+        logger.info("Posted successfully")
+
+    def alert_management_upon_start(self):
+        url = "http://{}/begin_polling".format('127.0.0.1:5000')
+        # url = "http://{}/begin_polling".format(model_data['admin_addr'])
```
dcrankshaw: Uncomment this

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -97,6 +99,10 @@ def stop_models(self, models):
         """
         return
 
+    @abc.abstractmethod
+    def upload_model_data(self, model_data):
```
dcrankshaw: Delete this

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -166,6 +167,22 @@ def connect(self):
         # No extra connection steps to take on connection
         return
 
+    def upload_model_data(self, model_data_list):
```
dcrankshaw: delete this

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -13,6 +13,12 @@
 import yaml
 import os
 import time
+from multiprocessing import Process, Manager, Value
+import requests
+import gevent
+import traceback
```
dcrankshaw: Delete these imports

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -70,9 +92,19 @@ def __init__(self,
...
+        # self.state_is_Unchanged = True
+        # manager = Manager()
+        # self.query_to_model = manager.dict()
+        self.balancer = RoundRobinBalancer(0)
+
+
+
     def start_clipper(self, query_frontend_image, mgmt_frontend_image,
-                      cache_size):
-        # If an existing Redis service isn't provided, start one
```
dcrankshaw: Don't delete code comments

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -119,6 +152,61 @@ def start_clipper(self, query_frontend_image, mgmt_frontend_image,
                     body=body, namespace='default')
         self.connect()
 
+        deployment_name = "query-frontend"
+
+        self._k8s_beta.patch_namespaced_deployment_scale(
```
dcrankshaw, line 141: Rather than patching after the deployment is started, just modify the initial deployment spec

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -165,55 +253,117 @@ def connect(self):
...
+                                           "name": model_data['name'],
+                                           "version": model_data['version'],
+                                           "input_type": model_data['input_type']
+                                           }
+            url = "http://{}/get_query_to_model_mapping".format('127.0.0.1:5000')
+            # url = "http://{}/get_query_to_model_mapping".format(self.get_admin_addr())
+
+            r = requests.post(url, json=model_mapping)
+
+        logger.info("Posted successfully")
```
dcrankshaw: Delete this

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -165,55 +253,117 @@ def connect(self):
...
+                        "name": deployment_name
+                    },
+                    'spec': {
+                        'replicas': 1,
+                        'template': {
+                            'metadata': {
+                                'labels': {
+                                    CLIPPER_MODEL_CONTAINER_LABEL:
+                                    create_model_container_label(name, version),
+                                    CLIPPER_DOCKER_LABEL:
```
dcrankshaw, line 252: Let's add another label here like:

```py
CLIPPER_MODEL_CONTAINER_QUERY_FRONTEND_IP = "ai.clipper..."
'labels': {..., CLIPPER_MODEL_CONTAINER_QUERY_FRONTEND_IP: clipper_ip, ...}
```

mgv2/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN apt-get install -y gcc
+COPY ./clipper_admin/ /clipper_admin
+RUN pip install /clipper_admin
+
+RUN pip install requests\
+&& pip install -r /mgv2/requirements.txt
+
+ENV HOME /home
+
+COPY ./mgv2/config /home/.kube/
```
dcrankshaw: Don't copy the config in. Check out [this issue](https://github.com/kubernetes-client/python/issues/254) for accessing K8s api from within a pod.

mgv2/mgv2.py
```diff
@@ -0,0 +1,307 @@
...
+import argparse
+import traceback
+
+
+logger = logging.getLogger(__name__)
+cur_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+
+class Mgv2():
```
dcrankshaw: If all the methods are static, you probably don't need a class at all

mgv2/mgv2.py
```diff
@@ -0,0 +1,307 @@
...
+    monkey.patch_all()
+    app = Flask(__name__)
+    pool = redis.ConnectionPool(host='127.0.0.1', port=6379)
+    r = redis.Redis(connection_pool = pool)
+    r.flushdb()
+    container_manager = None
+    config.load_kube_config()
```
dcrankshaw: Change this to load the in cluster config

mgv2/mgv2.py
```diff
@@ -0,0 +1,307 @@
...
+            logger.info("Starting polling")
+            polling = Process(target=Mgv2.poll_forever)
+            polling.start()
+            return 'Success'
+        except Exception as e:
+            traceback.print_exc()
+            return 'Failure'
+
+    @staticmethod
+    @app.route('/get_query_to_model_mapping', methods=['GET', 'POST'])
```
dcrankshaw: Delete the "GET" method

mgv2/mgv2.py
```diff
@@ -0,0 +1,307 @@
...
+        logger.info("Flask has received model info")
+
+        try:
+            query_to_model_mapping = request.get_json(force = True)
+
+
+            if not isinstance(query_to_model_mapping, dict):
+                query_to_model_mapping = json.loads(query_to_model_mapping)
+
+            logger.info("Query_to_model_mapping: {}".format(query_to_model_mapping))
```
dcrankshaw: Refactor the rest of this method into a separate method that takes the `query_to_model_mapping` dict. Then you can both call it here, and after you call deploy_model in `reassign...`

dcrankshaw (changes requested):
Added initial review

rbala19:
@dcrankshaw I addressed all the changes in the review. I performed all testing locally; I did not perform any tests where I deployed the mgv2 image to kubernetes. If you want to enable a local testing environment, uncomment the lines 1086 (clipper_admin.py),  1104 (clipper_admin.py), and 50 (mgv2.py) and comment the lines 1087 (clipper_admin.py), 1104 (clipper_admin.py), 49 (mgv2.py). These lines have to do with mgv2 URL creation and config loading. 

To test locally, you must also deploy a redis server using the command redis-server on terminal. To test locally, I start mgv2.py, then start clipper through the ipynb. Then I can add models through ipynb or take down frontends using kubectl delete 

One major omission is the feature that stops models on downed frontends. We discussed adding labels, for which I made the code changes, but there is an error with the syntax of the label selector in "stop_models_on_dead_ip" that I can't seem to figure out. I have caught the exception this call throws and it displays the stacktrace every time deploy_model is called within mgv2.py, in case that information will be helpful for debugging.

Please let me know if you have any further questions.

dcrankshaw:
@simon-mo figured out a better way to do this, and so this has been replaced by #462

PR #426, opened 2018-03-07
Fix installation order errors in deployer containers

Corey-Zumar:
Installing Cloudpickle after installing function-specific anaconda dependencies often yields an IOError. This seems to be fixed by ensuring that Cloudpickle is installed before all other anaconda dependencies.


dcrankshaw (approved):
lgtm

PR #411, opened 2018-02-16
[Metric Monitoring] Add Kubernetes Support

simon-mo:
This PR adds metrics monitoring to Kubernetes container manager. 

Todos:
- [x] Integration Test
- [x] Doc
- [x] ~~Modify the k8s example~~  Just add a readme

PR #427, opened 2018-03-08
[Metrics] Kubernetes Support & Maintenance 

simon-mo:
This PR replace #411 with cleaner commit history and different integration test.

The integration test issue come with #411 is hard to debug. It turns out to be:
- Inside each pod, the kubelet needs time to setup the network service within a pod. We just need to wait for the network service to be setup. It takes 30 seconds to 2 minutes. This is fixed by using retry and while loop inside integration test for kubernetes metrics. 

This PR also adds some maintenance:
- Modify the docker integration test as well to test for specific node health. It turns out the original test 1 calls API that returns the same information not matter the node is healthy or not. 
- ~~Let frontend-exporter to be more fault tolerant with some try-catch. Not a big fix so it just with this PR.~~ Independent PR in #429
- Update readme


simon-mo:
~~Blocked by #429~~

clipper_admin/clipper_admin/kubernetes/prom_deployment.yml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        ai.clipper.container.label: ""
+        ai.clipper.name: metrics
+    spec:
+      containers:
+      - image: prom/prometheus:latest
```
dcrankshaw: Let's use a versioned image rather than `latest`

clipper_admin/clipper_admin/kubernetes/kubernetes_metric_utils.py
```diff
@@ -0,0 +1,63 @@
...
+from contextlib import contextmanager
+from kubernetes.client.rest import ApiException
+import logging
+import json
+from ..version import __version__
+
+_cur_dir = os.path.dirname(os.path.abspath(__file__))
+prom_deployment_path = os.path.join(_cur_dir, 'prom_deployment.yml')
+prom_service_path = os.path.join(_cur_dir, 'prom_service.yml')
+prom_configmap_path = os.path.join(_cur_dir, 'prom_configmap.yml')
```
dcrankshaw: Change these file suffixes to `.yaml` to keep them consistent with the rest of the K8s yaml files

integration-tests/clipper_metric_kube.py
```diff
@@ -0,0 +1,232 @@
+# This integration test is taken straight fron `kubernetes_integration_test.py
+# With the setting 1 app, 1 model, 2 replica.
+# Adding container monitoring check
```
dcrankshaw, line 3: @simon-mo would it make sense to just add this test to the existing Kubernetes integration test, or do you think it's better to keep it separate? I'll defer to your judgment here.

monitoring/README.md
```diff
@@ -1,9 +1,22 @@
...
+We use prometheus as the metric tracking system. Once you spin up a clipper query frontend and a model containers. 
+If you are using `DockerContainerManager`, You can view prometheus UI at: [`http://localhost:9090`](http://localhost:9090). 
+If you are using `KubernetesContainerManager`, You can query the metric address by calling `get_metric_addr()`. 
 
-Please note that Prometheus UI is for debug purpose only. You can view certain metric and graph the timeseries. But for better visualization, we recommend [Grafana](https://grafana.com/). Grafana has default support for Prometheus Client. 
+Please note that Prometheus UI is for debug purpose only. You can view certain metric and graph the timeseries. But for better visualization, we recommend [Grafana](https://grafana.com/). Grafana has default support for Prometheus Client. Feel free to checkout `examples/monitoring` for example of displaying Clipper metrics in Grafana.
 
 ## Avaliable Metrics
-See `metrics_config.yaml` for the metrics detail and description
+See `metrics_config.yaml` for the metrics detail and description. This configuration file is used in the creation of `RPC.py` container
```
dcrankshaw: change `RPC.py` to `rpc.py`

monitoring/front_end_exporter.py
```diff
@@ -20,10 +20,16 @@
 
 ADDRESS = 'http://{}/metrics'.format(query_frontend_id)
 
+print("Scraping {}".format(ADDRESS))
```
dcrankshaw: Change print to a logging statement

monitoring/front_end_exporter.py
```diff
@@ -20,10 +20,16 @@
...
+
 
 def load_metric():
-    res = requests.get(ADDRESS)
-    return res.json()
+    try:
+        res = requests.get(ADDRESS)
+        return res.json()
+    except Exception as e:
+        print("Scrape Failed! Error: {}\n".format(e))
```
dcrankshaw: change print to a logging statement

monitoring/front_end_exporter.py
```diff
@@ -38,6 +44,10 @@ def multi_dict_unpacking(lst):
 
 
 def parse_metric(metrics):
+    if not len(metrics):
```
dcrankshaw: For the sake of being explicit, change this to `if len(metrics) == 0:`

dcrankshaw (changes requested):
Looks great. Just some minor formatting comments.

integration-tests/clipper_metric_kube.py
simon-mo: The metrics take about 3 minutes. I would keep them separate for now. But this can be a potential solution to speedup our integration test even though it makes the test less modular.

simon-mo:
Jenkins test this please

simon-mo:
jenkins test this please

integration-tests/clipper_metric_kube.py
```diff
@@ -31,12 +31,12 @@
 # TODO: Add kubernetes specific checks that use kubernetes API
```
dcrankshaw: Is this TODO still outstanding?

dcrankshaw (approved):
lgtm

PR #429, opened 2018-03-08
[Metrics] Frontend-exporter fault tolerance quick fix

simon-mo:
It turns out we do need a small PR to update frontend-exporter package. #427 won't pass the test with current build.

I prefer a PR instead of passing the frontend-exporter image as parameter inside the code because this complexity should be hidden to the user as best as we can.


monitoring/front_end_exporter.py
```diff
@@ -20,10 +20,16 @@
 
 ADDRESS = 'http://{}/metrics'.format(query_frontend_id)
 
+print("Scraping {}".format(ADDRESS))
```
dcrankshaw: Use Python logging instead of a print statement

monitoring/front_end_exporter.py
```diff
@@ -20,10 +20,16 @@
...
+
 
 def load_metric():
-    res = requests.get(ADDRESS)
-    return res.json()
+    try:
+        res = requests.get(ADDRESS)
+        return res.json()
+    except Exception as e:
+        print("Scrape Failed! Error: {}\n".format(e))
```
dcrankshaw: Use logging

monitoring/front_end_exporter.py
```diff
@@ -38,6 +44,10 @@ def multi_dict_unpacking(lst):
 
 
 def parse_metric(metrics):
+    if not len(metrics):
```
dcrankshaw: Just for the sake of being explicit, can you change this to `if len(metrics) == 0:`

dcrankshaw (changes requested):
This looks good. Just change the print statements to logging.

monitoring/front_end_exporter.py
```diff
@@ -16,14 +17,21 @@
     help='The name of docker container in clipper_network')
 args = parser.parse_args()
```
dcrankshaw, line 19: Add these lines right here:
```py
logging.basicConfig(
    format='%(asctime)s %(levelname)-8s [%(filename)s:%(lineno)d] %(message)s',
    datefmt='%y-%m-%d:%H:%M:%S',
    level=logging.INFO)
```

dcrankshaw (approved):
lgtm

PR #454, opened 2018-03-28
Added functionality to query previous version models

daminig:


dcrankshaw:
jenkins add to whitelist

integration-tests/clipper_admin_tests.py
```diff
@@ -391,7 +391,62 @@ def predict_func(xs):
         batch_pred_outputs = [batch['output'] for batch in batch_predictions]
         self.assertEqual(batch_pred_outputs,
                          test_batch_predict_result)  # tests batch input
+      
+      def test_version_query(self):
+        def predict_func1(xs):
+            return 1
+
+        def predict_func2(xs):
+            return 2
```
dcrankshaw: These predict functions won't work like this. Modify them to be
```py
def predict_func1(xs):
    return ["1" for _ in xs]

def predict_func2(xs):
    return ["2" for _ in xs]
```

integration-tests/clipper_admin_tests.py
```diff
@@ -391,7 +391,62 @@ def predict_func(xs):
...
+
+        headers = {"Content-type": "application/json"}
+        test_input = [1.0, 2.0, 3.0]
+        pred1 = requests.post(
+            url, headers=headers, data=json.dumps({
+                "input": test_input
+                "version": "1"
+            })).json()
+        
+        self.assertEqual([pred1['output']], 1)
```
dcrankshaw: Change this to be `self.assertEqual([pred1['output']], "1")`

integration-tests/clipper_admin_tests.py
```diff
@@ -391,7 +391,62 @@ def predict_func(xs):
...
+        
+        pred2 = requests.post(
+            url, headers=headers, data=json.dumps({
+                "input": test_input
+            })).json()
+        
+        self.assertEqual([pred2['output']], 2)
```
dcrankshaw: Change to `self.assertEqual([pred2['output']], "2")`

src/frontends/src/query_frontend.hpp
```diff
@@ -301,26 +315,52 @@ class RequestHandler {
...
-        std::vector<VersionedModelId> versioned_models;
-        {
-          std::unique_lock<std::mutex> l(current_model_versions_mutex_);
+        rapidjson::Document d;
+        clipper::json::parse_json(request->content.string(), d);
+        std::vector<VersionedModelId> versioned_models = {};
+        std::vector<std::string> models;
+        std::string version = clipper::json::get_string(d, "version");
+        if (!version.empty()) {
+          models = clipper::redis::get_model_versions(redis_connection_, name);
```
dcrankshaw: You need to acquire the mutex here:

```cpp
std::unique_lock<std::mutex> l(current_model_versions_mutex_);
```

src/frontends/src/query_frontend.hpp
```diff
@@ -301,26 +315,52 @@ class RequestHandler {
...
+            }
+            if (versioned_models.empty()) {
+              throw version_id_error("Requested model does not exist."); 
             }
           }
-        }
+        } else {
+          models = get_linked_models_for_app(name);
+          {
+            std::unique_lock<std::mutex> (current_model_versions_mutex_);
```
dcrankshaw: This syntax isn't quite right. It should be 
```cpp
std::unique_lock<std::mutex> l(current_model_versions_mutex_);
```

Note the **l** before `(current_model_versions_mutex)`

dcrankshaw (changes requested):
This looks good. Just some minor cleanup.

dcrankshaw (approved):
LGTM. Will merge once the tests pass.

dcrankshaw (changes requested):
Can we get this compiling and passing the format checker soon?

daminig:
jenkins test this please

dcrankshaw:
jenkins test this please

dcrankshaw (approved):
I fixed this PR up, and it should be good to go once the tests pass.

PR #457, opened 2018-03-30
Add functionality to delete applications

JayShenoy:


dcrankshaw:
@JayShenoy can you delete all the cmake files you committed? This PR has 13,000 files and over a million lines of code.

JayShenoy:
My mistake, I just removed that mess

JayShenoy:
@dcrankshaw The frontend and management tests now all pass, but the integration tests are failing because of issues with Docker that I believe are local to my machine.

JayShenoy:
Jenkins test this please

dcrankshaw:
jenkins add to whitelist

dcrankshaw (approved):
This looks great. Very nice work. I'll merge once the tests pass.

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -188,6 +188,25 @@ def register_application(self, name, input_type, default_output,
...
+        headers = {"Content-type": "application/json"}
+        r = requests.post(url, headers=headers, data=req_json)
+        logger.debug(r.text)
+        if r.status_code != requests.codes.ok:
+            msg = "Received error status code: {code} and message: {msg}".format(
+                code=r.status_code, msg=r.text)
+            logger.error(msg)
+            raise ClipperException(msg)
+        else:
+            logger.info("Application {app} was successfully deleted").format(
```
simon-mo: This line will error:
```
    logger.info("Application {app} was successfully deleted").format(
AttributeError: 'NoneType' object has no attribute 'format'
```

integration-tests/clipper_admin_tests.py
```diff
@@ -80,6 +80,18 @@ def test_link_not_registered_model_to_app_fails(self):
             self.clipper_conn.link_model_to_app(app_name, not_deployed_model)
         self.assertTrue("No model with name" in str(context.exception))
 
+    def test_delete_application_correct(self):
+        nput_type = "doubles"
+        default_output = "DEFAULT"
+        slo_micros = 30000
+        app_name = "testapp"
+        self.clipper_conn.register_application(app_name, input_type,
```
simon-mo: This line will error:
```
Traceback (most recent call last):
  File "clipper_admin_tests.py", line 88, in test_delete_application_correct
    self.clipper_conn.register_application(app_name, input_type,
NameError: global name 'input_type' is not defined
```

integration-tests/clipper_admin_tests.py
```diff
@@ -80,6 +80,18 @@ def test_link_not_registered_model_to_app_fails(self):
             self.clipper_conn.link_model_to_app(app_name, not_deployed_model)
         self.assertTrue("No model with name" in str(context.exception))
 
+    def test_delete_application_correct(self):
```
simon-mo, line 83: You also need to add this test function to SHORT_TEST_ORDERING list below:
https://github.com/ucbrise/clipper/blob/ff849d4ade612732a72427b156206ec9333073fe/integration-tests/clipper_admin_tests.py#L567

simon-mo (commented):
@JayShenoy Great work. A quick question: does this PR intend to just delete applications? Is it suppose to do any clean up between app_model links. 

For example, if I registered an application `simple-app`, and deploy model `sum-model` linking to `simple-app`, right now if I delete `simple-app`, the model container will not be deleted. This is intentional? 

Thanks for the great work!

dcrankshaw:
@simon-mo Yeah that behavior is correct. Deleting an application should not delete any models.

integration-tests/clipper_admin_tests.py
dcrankshaw: Oh good catch!

JayShenoy:
@simon-mo This PR is purely intended to delete applications. I believe someone else is working on deleting models.

@dcrankshaw I have made the necessary revisions.

JayShenoy:
Bless up 🙏

dcrankshaw:
Nice job!

PR #458, opened 2018-03-31
Fix QueryFrontend bug

### LINKED ISSUE #456: [Bug] Clipper returns an unintended result when prediction result is too long
I modified QuickStart codes like below,
```Python
import json
from datetime import datetime
import numpy as np
from clipper_admin import ClipperConnection, DockerContainerManager
from clipper_admin.deployers import python as python_deployer

clipper_conn = ClipperConnection(DockerContainerManager())
clipper_conn.start_clipper(cache_size=1)
clipper_conn.register_application(name="hello-world", input_type="doubles", default_output="-1.0", slo_micros=1000000)

def feature_sum(xs):
    def gen(i):
        output = {'stage': 'beta',
                  'input': str(i),
                  'created': datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
                  'result': list(np.repeat(1234567890123456789, 200))}  # too long!
        return json.dumps(output)
    return [gen(x) for x in xs]

python_deployer.deploy_python_closure(clipper_conn, name="sum-model", version=1, input_type="doubles", func=feature_sum)
clipper_conn.link_model_to_app(app_name="hello-world", model_name="sum-model")
```

And then, sent some following calls to the Clipper API,
```Python
import requests, json, numpy as np
headers = {"Content-type": "application/json"}
requests.post("http://localhost:1337/hello-world/predict", headers=headers, data=json.dumps({"input": list(np.random.random(1))})).json()
```
I received the following unintended result.
```
{u'batch_predictions': []}
```
QueryFrontend's error logs are here.
```
[15:29:03.606][info]  [QUERYPR...] Found 1 tasks
[15:29:03.606][info]  [TASKEXE...] Adding task to queue. QueryID: 0, model: sum-model:1
[15:29:03.606][info]  [TASKEXE...] Sending batch to model: sum-model:1 replica 0.Batch size: 1. Query IDs: 0
[15:29:03.617][info]         [RPC] Found message to receive
[15:29:03.617][error] [QUERYFR...] Returned response before all predictions in batch were processed
[15:29:05.007][info]  [QUERYFR...] METRICS
```
But if I reduce prediction result's size, everything is OK.
```python
def feature_sum(xs):
    def gen(i):
        output = {'stage': 'beta',
                  'input': str(i),
                  'created': datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
                  'result': list(np.repeat(1234567890123456789, 10))}  # 200 --> 10
        return json.dumps(output)
    return [gen(x) for x in xs]
```

What's wrong?

withsmilo:
* fixes #456


dcrankshaw:
jenkins ok to test

dcrankshaw:
Thanks for tracking this down

dcrankshaw (approved):
lgtm

Issue #475, opened 2018-04-20
Caffe2 Build Issue


simon-mo:
Caffe2OnnxContainer is having issue building from source. 

Some submodule checkout failed:
```
error: no such remote ref f3c627d517968c20e8269ead1d90cd3a6c199356
Fetched in submodule path 'third_party/aten', but it did not contain f3c627d517968c20e8269ead1d90cd3a6c199356. Direct fetching of that commit failed.
```

Jenkins Log: https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/1371/console

Related issue on PyTorch and Caffe2 issue page:
https://github.com/caffe2/caffe2/issues/2510
https://github.com/pytorch/pytorch/issues/6776

simon-mo:
I was able to reproduce it on my local machine:

```
git clone --recursive https://github.com/caffe2/caffe2


Cloning into 'caffe2'...
remote: Counting objects: 230439, done.
remote: Total 230439 (delta 0), reused 0 (delta 0), pack-reused 230439
Receiving objects: 100% (230439/230439), 393.42 MiB | 667.00 KiB/s, done.
Resolving deltas: 100% (211258/211258), done.
Submodule 'third_party/ComputeLibrary' (https://github.com/ARM-software/ComputeLibrary.git) registered for path 'third_party/ComputeLibrary'
Submodule 'third_party/NNPACK_deps/FP16' (https://github.com/Maratyszcza/FP16.git) registered for path 'third_party/FP16'
Submodule 'third_party/NNPACK_deps/FXdiv' (https://github.com/Maratyszcza/FXdiv.git) registered for path 'third_party/FXdiv'
Submodule 'third_party/NNPACK' (https://github.com/Maratyszcza/NNPACK.git) registered for path 'third_party/NNPACK'
Submodule 'third_party/aten' (https://github.com/zdevito/aten) registered for path 'third_party/aten'
Submodule 'third_party/benchmark' (https://github.com/google/benchmark.git) registered for path 'third_party/benchmark'
Submodule 'third-party/cpuinfo' (https://github.com/Maratyszcza/cpuinfo.git) registered for path 'third_party/cpuinfo'
Submodule 'third_party/cub' (https://github.com/NVlabs/cub.git) registered for path 'third_party/cub'
Submodule 'third_party/eigen' (https://github.com/RLovelett/eigen.git) registered for path 'third_party/eigen'
[...]

PR #476, opened 2018-04-20
Checkout Caffe2 source from PyTorch repo

simon-mo:
Hopefully this can address #475


dcrankshaw:
For now, rather than trying to wrestle with Caffe2's build issues, we are going to temporarily remove our support for Caffe2.

Issue #433, opened 2018-03-12
Allow models to be deleted


dcrankshaw:
As part of the effort to complete #342

rkooo567:
https://github.com/ucbrise/clipper/pull/654 Resolved

PR #478, opened 2018-04-20
Deletion of unlinked models

es1024:
This PR allows you to remove models that are not linked to applications (#433).

There is a race condition in the management frontend when checking if the model is linked, but fixing this requires using multi/exec, which requires a bit of restructuring of the redis code (the return value for all calls in a multi/exec block is a string).


dcrankshaw:
jenkins okay to test

dcrankshaw:
jenkins ok to test

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -634,6 +634,40 @@ def register_model(self,
                 "Successfully registered model {name}:{version}".format(
                     name=name, version=version))
 
+    def remove_model(self, name):
```
withsmilo: How about rename to 'delete_unlinked_model'?

src/management/src/management_frontend.hpp
```diff
@@ -53,6 +53,7 @@ const std::string ADD_APPLICATION = ADMIN_PATH + "/add_app$";
 const std::string DELETE_APPLICATION = ADMIN_PATH + "/delete_app$";
 const std::string ADD_MODEL_LINKS = ADMIN_PATH + "/add_model_links$";
 const std::string ADD_MODEL = ADMIN_PATH + "/add_model$";
+const std::string REMOVE_MODEL = ADMIN_PATH + "/remove_model$";
```
withsmilo: How about rename to 'delete_unlinked_model'?

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -300,6 +300,37 @@ class ModelQueueThreadPool : public ThreadPool {
...
+                          "Work queue does not exist for model {}, replica {}",
+                          vm.serialize(), std::to_string(replica_id));
+      return false;
+    } else {
+      queue->second.invalidate();
+      l.unlock();
+
+      auto thread = threads_.find(queue_id);
+      if (thread->second.joinable()) {
+        thread->second.join();
```
withsmilo, line 318: Did you have a success to join with the terminated thread in this code?
I think that it causes a failure because TaskExecutionThread is waiting for a signal of queue_not_empty_condition_ conditional variable in the TaskExecutor.

es1024: The relevant queue is first invalidated, which signals all waiters on the condition variable, so this shouldn't be an issue; the thread should finish any current running task and terminate.

withsmilo: It is right that the relevant queue is first [invalidated](https://github.com/es1024/clipper/blob/model-delete/src/libclipper/include/clipper/threadpool.hpp#L97-L108), but the 'invalidate' function signaled other [conditional variable](https://github.com/es1024/clipper/blob/model-delete/src/libclipper/include/clipper/threadpool.hpp#L122), not [queue_not_empty_condition_](https://github.com/es1024/clipper/blob/model-delete/src/libclipper/include/clipper/task_executor.hpp#L145) conditional variable for which TaskExecutorThread is waiting in the TaskExecutor. Please refer to the [these codes](https://github.com/es1024/clipper/blob/model-delete/src/libclipper/include/clipper/threadpool.hpp#L220-L223).

dcrankshaw:
@simon-mo Are these kubernetes failures related to the multi-tenancy PR debugging you've been doing?

https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/1611/console

simon-mo:
@dcrankshaw Nope. But this happened to my PR as well and it can’t be reproduced.

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
Closing because this is outdated. I'll try to submit a new PR re-implement this on a higher level management API.

PR #479, opened 2018-04-20
Temporarily Disable Caffe2 Build & Test

simon-mo:
Address #475

### DIFF
--- a/bin/build_docker_images.sh
+++ b/bin/build_docker_images.sh
@@ -259,7 +259,8 @@ build_images () {
     create_image pyspark-container PySparkContainerDockerfile $public
     create_image tf-container TensorFlowDockerfile $public
     create_image pytorch-container PyTorchContainerDockerfile $public
-    create_image caffe2-onnx-container Caffe2OnnxDockerfile $public
+    # See issue #475
+    # create_image caffe2-onnx-container Caffe2OnnxDockerfile $public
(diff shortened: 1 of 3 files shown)


goswamig:
Currently I cant even clone the caffe2 repo :-/

Is there any plan to fix it soon?

simon-mo:
Hi, @gautamkmr 

We chose not to investigate and find a fix for this issue temporarily. I believe you can still install caffe2 via `anaconda`. The pre-built binary should be stable.

dcrankshaw (approved):

dcrankshaw (changes requested):
Can you throw an exception in the pytorch onnx deployer as well with a link to the issue? That's where this docker image gets used.

simon-mo:
@dcrankshaw A warning may be more suitable? The old containers on dockerhub still works. We just need to re-tag the image for the release. What about a warning saying: the caffe 2 version is not up to date because {link to issue} however you can still use it (we will continue our effort to  maintain it once caffe2 build issue is fixed....)

dcrankshaw:
Yeah that sounds good.

simon-mo:
I just tried pulling the latest pytorch (which contains caffe now), it seems to work.

dcrankshaw (approved):
lgtm

Issue #468, opened 2018-04-13
QueryFrontend's memory usage increases continuously


withsmilo:
I have deployed so many different versions of the same model repeatedly using Clipper to our cluster for a week. As the result, QueryFrontend's memory usage increased continuously. (95MB -> 250MB) I think that ActiveContainers list of QueryFrontend is growing up quickly, but old `ModelContainer`s are not deleted from the ActiveContainers list.

Is this intentional?

dcrankshaw:
It's intended behavior but not desired behavior. #378 detects when old containers are no longer connected and GCs them from the active containers list, which should address this issue.

withsmilo:
@dcrankshaw : Thanks for your reply. For serving side of stability, I think that this issue is critical.
We will apply some workaround fixes before merging with #378.

withsmilo:
I would like to share some test results with you. My Clipper's QueryFrontend maintains latest ModelContainer only in the ActiveContainers list (containers_), but QueryFrontend's memory usage still increases continuously. 

Where do you think the cause is? ==> I think that RPCService's `connections_containers_map` is increasing continuously. Thank you!

dcrankshaw:
That's a good point. @amogkam is going to look into cleaning up the connections map now that #378 got merged.

amogkam:
@withsmilo #378 should actually remove inactive containers from both the `active_containers` list and `connections_containers_map`. 

If the problem is persistent even with this PR merged, then I can dig into it further, but for now it looks like everything should be good.

withsmilo:
@dcrankshaw @amogkam : 
Thank you very much for quick review and merging of #378 PR. But after merging #378 with my branch, I found the new following problems.

#### **1. QueryFrontend sometimes crashed and caused the critical segment fault.**
How about you? I cannot be certain of that this problem is for Clipper develop branch or not because my branch is a little different from it. Anyway, to solve this crash, I googled this [so post](https://stackoverflow.com/questions/38468844/erasing-elements-from-unordered-map-in-a-loop) and modified RPCService::check_container_activity() function like below and the crash was disappeared finally.
```c++
void RPCService::check_container_activity(
    std::unordered_map<std::vector<uint8_t>, ConnectedContainerInfo,
                       std::function<size_t(const std::vector<uint8_t> &vec)>>
        &connections_containers_map) {
  std::chrono::system_clock::time_point current_time =
      std::chrono::system_clock::now();

[...]

PR #481, opened 2018-04-23
Fix an infinite loop of GarbageCollectionThread

withsmilo:
* Changed list
  * TaskExecutionThread's worker calls try_pop() to spin.
  * GarbageCollectionThread's worker calls wait_pop() to block.
* related issue: #468

### DIFF
--- a/src/libclipper/include/clipper/threadpool.hpp
+++ b/src/libclipper/include/clipper/threadpool.hpp
@@ -289,10 +289,10 @@ class ModelQueueThreadPool : public ThreadPool {
     } else {
       queues_.emplace(std::piecewise_construct, std::forward_as_tuple(queue_id),
                       std::forward_as_tuple());
-      threads_.emplace(
-          std::piecewise_construct, std::forward_as_tuple(queue_id),
-          std::forward_as_tuple(&ModelQueueThreadPool::worker, this, queue_id,
-                                is_block_worker));
+      threads_.emplace(std::piecewise_construct,
+                       std::forward_as_tuple(queue_id),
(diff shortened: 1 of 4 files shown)


dcrankshaw:
jenkins ok to test

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -325,11 +335,18 @@ class FixedSizeThreadPool : public ThreadPool {
...
+    } catch (ThreadPoolError& e) {
+      log_error(LOGGING_TAG_THREADPOOL,
+                "Failed to submit task for FixedSizeThreadPool");
+      throw(e);
+    }
   }
 
  private:
   size_t queue_id_ = 1;
+  bool is_block_worker_ = true;
```
dcrankshaw: Rather than hardcoding this to one, can you make it a constructor argument, and then set it to `true` in `GarbageCollectionThreadpool::getThreadPool` where we call the `FixedSizeThreadpool` constructor? Whether the worker should block or not is a result of the semantics of the GarbageCollector, not a fundamental part of the FixedSizeThreadpool semantics.

dcrankshaw (changes requested):
Thanks for tracking this down! Just had a minor comment for improving the usability of `ThreadPool` in the future.

src/libclipper/include/clipper/threadpool.hpp
withsmilo: Good review. I will try to modify it.

dcrankshaw (approved):
lgtm

Issue #490, opened 2018-05-14
Kubernetes label selector should use = to match label to value, not :


maccam912:
https://github.com/ucbrise/clipper/blob/6388d9e8c07b780f9b0d74aaa05628f01f1f7bea/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py#L391

I haven't looked for where else this might be, but I was trying to stop all models with a certain name and noticed that this label selector needs an equal sign instead of a colon. The way it is  written now results in this error:

```
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Mon, 14 May 2018 18:18:52 GMT', 'Content-Length': '455'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"unable to parse requirement: invalid label key \"ai.clipper.model_container.label:addone-model_1\": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')","reason":"BadRequest","code":400}
```

maccam912:
I'll do a really quick PR for this.

maccam912:
This deletes the deployment just fine, but it doesn't delete replicaset or pods as far as I can tell. It looks like stop_all() does these manually.

maccam912:
I'll update my PR to include deleting the replica sets and pods too. And fix stop_all_model_containers too.

avinregmi:
Hey, I'm using your updated code with replacing (:) with (=). I'm still not able to delete deployment from google k8s. I've noticed, it turns into replicaset from deployment after calling stop_models() method.

withsmilo:
resolved by #729.

PR #492, opened 2018-05-14
Fix label selector when deleting model deployments, and delete replica sets and pods with deployment to follow stop_all() example of deletion

maccam912:
https://github.com/ucbrise/clipper/issues/490


simon-mo:
Hi @maccam912 thanks for contributing! I think deleting deployments will automatically deleting replica sets and pods since deployment manage all that:
https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
"""
While ReplicaSets can be used independently, today it’s mainly used by Deployments as a mechanism to orchestrate pod creation, deletion and updates. When you use Deployments you don’t have to worry about managing the ReplicaSets that they create. Deployments own and manage their ReplicaSets.
"""

dcrankshaw:
@maccam912 Were you observing replica sets and pods that were not deleted after deleting the deployment?

rkooo567:
@simon-mo I believe this PR can be closed?

PR #496, opened 2018-05-17
Issue with Number of Frontend Replicas

santi81:
The number of frontend replicas is not known if we call the KubernetesContainerManager connect method directly. 
The attribute self.num_frontend_replicas would not be defined . Hence a subsequent call to deploy would fail.

This needs to be fetched during connect and set. Since each front end creates an underlying deployment, the only way i could think of was to read and filter the deployments with the label ai.clipper.name=query-frontend.

### DIFF
--- a/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
+++ b/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
@@ -228,7 +228,6 @@ class KubernetesContainerManager(ContainerManager):
                 elif p.name == "7000":
                     self.clipper_rpc_port = p.node_port
 
-            # NOTE: status of the front end deployments not considered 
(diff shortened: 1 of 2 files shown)


simon-mo:
@santi81 Thanks for catching that!

simon-mo:
jenkins ok to test

simon-mo (approved):
lgtm

santi81:
@simon-mo @dcrankshaw the builds are failing with random errors every time.I checked and it has got nothing to do with this PR. I guess they are some build issues in general . Could you guys have a look

simon-mo:
jenkins test this please

santi81:
@simon-mo : can you pls merge this PR

simon-mo:
@santi81 Merged! Thanks for contributing!

PR #502, opened 2018-05-29
Fix the following build error of rpc_service.cpp

withsmilo:
```
/src/libclipper/src/rpc_service.cpp:389:16: error: no member named 'accumulate' in namespace 'std'
          std::accumulate(output_header, output_header + num_outputs, 0));
          ~~~~~^
1 error generated.
```


dcrankshaw:
jenkins ok to test

dcrankshaw (approved):
lgtm

Issue #460, opened 2018-04-04
Enable multiple Clipper clusters to be running concurrently in Docker & Kubernetes


dcrankshaw:
We should allow users to specify a cluster name for Clipper when starting an instance. Then a ClipperConnection object should be associated with that specific cluster name. This will enable multiple clusters to be running side by side on the same K8s cluster, which is useful for supporting multi-tenancy applications and will help with testing.

This can be achieved via a slight modification of the way we use Kubernetes labels to identify Clipper instances.

dcrankshaw:
Alternatively we can use Kubernetes namespaces to achieve this

simon-mo:
As a note: this will also fix #424 #448 ~#501~

simon-mo:
Implementation Note:

## Docker
Docker multi-tenancy will be implemented via adding labels to containers. In Docker, each `run` command will always create a new container. Each container will have a randomly generated name which is used as unique identifier.

## Kubernetes
We can't do the same in k8s because k8s use `name` as unique identifier. However, the `name` is specified by users. Therefore, we must:
- When creating new deployment/services, we generate name as `cluster_name`+`name`
- We also want to add labels for:
   - cluster
   - function {mgmt, query, etc..}

simon-mo:
Closed by #503

PR #500, opened 2018-05-25
Support k8s namespaces

santi81:
Run Clipper in its own user-specified namespace on Kubernetes #437

PR #503, opened 2018-05-29
Multi tenancy support

simon-mo:
This is a work in process, almost done.

Resolve issue #460


simon-mo:
jenkins test this please

bin/run_unittests.sh
```diff
@@ -144,6 +144,7 @@ function run_integration_tests {
   ../integration-tests/r_integration_test/rclipper_test.sh
   python ../integration-tests/clipper_metric_docker.py 
   python ../integration-tests/clipper_metric_kube.py 
+  python ../integration-tests/multi_tenency_test.py
```
dcrankshaw: Nit: tenancy is spelled with an "a"

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -30,7 +30,7 @@
 
 DEFAULT_LABEL = []
 DEFAULT_PREDICTION_CACHE_SIZE_BYTES = 33554432
-CLIPPER_TEMP_DIR = "/tmp/clipper"
+CLIPPER_TEMP_DIR = "/tmp/clipper"  # Used Internally for Test; Not Windows Compatible
```
dcrankshaw, line 33: Is this true that it's only used for testing?

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -1,19 +1,32 @@
...
 
 # Constants
 CLIPPER_INTERNAL_QUERY_PORT = 1337
 CLIPPER_INTERNAL_MANAGEMENT_PORT = 1338
 CLIPPER_INTERNAL_RPC_PORT = 7000
 CLIPPER_INTERNAL_METRIC_PORT = 1390
+CLIPPER_INTERNAL_REDIS_PORT = 6379
 
 CLIPPER_DOCKER_LABEL = "ai.clipper.container.label"
+CLIPPER_NAME_LABEL = "ai.clipper.name"
```
dcrankshaw, line 13: This refers to the cluster name? Can we change it to `ai.clipper.cluster.name`?

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -1,19 +1,32 @@
...
 CLIPPER_MODEL_CONTAINER_LABEL = "ai.clipper.model_container.label"
 CLIPPER_QUERY_FRONTEND_CONTAINER_LABEL = "ai.clipper.query_frontend.label"
 CLIPPER_MGMT_FRONTEND_CONTAINER_LABEL = "ai.clipper.management_frontend.label"
 CLIPPER_QUERY_FRONTEND_ID_LABEL = "ai.clipper.query_frontend.id"
 CONTAINERLESS_MODEL_IMAGE = "NO_CONTAINER"
 
+CLIPPER_DOCKER_PORT_LABELS = {
+    'redis': 'ai.clipper.redis.port',
+    'query_query': 'ai.clipper.query_frontend.query.port',
```
dcrankshaw: What is query_query? Can you use a more descriptive name here?

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -34,6 +47,50 @@ def parse_model_container_label(label):
     return splits
 
 
+def find_unbound_port(start=None,
+                      increment=True,
+                      port_range=(34256, 50000),
```
dcrankshaw: Why did you pick 34256 as the lower bound on the port range?

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -34,6 +47,50 @@ def parse_model_container_label(label):
     return splits
 
 
+def find_unbound_port(start=None,
```
dcrankshaw: Move this function into the docker container manager, as it only works for finding locally unbound ports. The K8s container shouldn't use it (or need to).

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -115,22 +125,43 @@ def start_clipper(self,
             msg = "Unable to Connect to Docker. Please Check if Docker is running."
             raise ClipperException(msg)
 
+        containers_in_cluster = self.docker_client.containers.list(
+            filters={
+                'label':
+                ['ai.clipper.container.label={}'.format(self.cluster_name)]
```
dcrankshaw: Don't use the literal string here for the label key, use a variable instead

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -115,22 +125,43 @@ def start_clipper(self,
...
+            })
+        if len(containers_in_cluster) > 0:
+            raise ClipperException(
+                "Cluster {} cannot be started because it already exists. "
+                "Please use clipper_conn.connect() to connect to it.".format(
```
dcrankshaw: s/clipper_conn.connect()/ClipperConnection.connect()

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -115,22 +125,43 @@ def start_clipper(self,
...
+                    self.cluster_name))
+
         if not self.external_redis:
             logger.info("Starting managed Redis instance in Docker")
+            self.redis_port = find_unbound_port(self.redis_port)
+            redis_labels = self.common_labels.copy()
+            redis_labels[CLIPPER_DOCKER_PORT_LABELS['redis']] = str(
+                self.redis_port)
```
dcrankshaw: What is this port label used for?

clipper_admin/clipper_admin/container_manager.py
```diff
@@ -34,6 +47,50 @@ def parse_model_container_label(label):
...
+                      increment=True,
+                      port_range=(34256, 50000),
+                      verbose=False,
+                      logger=None):
+    """
+    Fina a unbound port.
```
dcrankshaw: Nit: s/Fina/Find

simon-mo:
jenkins test this please

bin/run_unittests.sh
```diff
@@ -91,7 +91,7 @@ function run_jvm_container_tests {
   echo "Running JVM container tests..."
   cd $DIR
   cd ../containers/jvm
-  mvn test
+  mvn test -q
```
dcrankshaw, line 94: 👍

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -447,17 +453,18 @@ def build_model(self,
...
             docker_client = docker.from_env()
-            logger.info(
+            self.logger.info(
                 "Building model Docker image with model data from {}".format(
                     model_data_path))
             image_result, build_logs = docker_client.images.build(
                 fileobj=context_file, custom_context=True, tag=image)
             for b in build_logs:
-                logger.info(b)
+                if 'stream' in b and b['stream'] != '\n':  #log build steps only
```
dcrankshaw, line 471: 👍

clipper_admin/clipper_admin/container_manager.py
dcrankshaw: It looks like this is for the client-facing service. Can you call it `query_rest`?

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -146,12 +181,19 @@ def start_clipper(self,
...
                          cache_size=cache_size)
-        query_labels = self.common_labels.copy()
-        query_labels[CLIPPER_QUERY_FRONTEND_CONTAINER_LABEL] = ""
+
         query_container_id = random.randint(0, 100000)
         query_name = "query_frontend-{}".format(query_container_id)
+        self.clipper_query_port = find_unbound_port(self.clipper_query_port)
+        self.clipper_rpc_port = find_unbound_port(self.clipper_rpc_port)
+        query_labels = self.common_labels.copy()
+        query_labels[CLIPPER_QUERY_FRONTEND_CONTAINER_LABEL] = ""
```
dcrankshaw, line 199: Why is this label value empty?

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -170,16 +212,49 @@ def start_clipper(self,
...
-        return
+        """
+        Use the cluster name to update ports. Because they might not match as in
+        start_clipper the ports might be changed.
+        :return: None
+        """
+        containers = self.docker_client.containers.list(
+            filters={
+                'label':
+                ['ai.clipper.container.label={}'.format(self.cluster_name)]
```
dcrankshaw: Don't use the literal value of the label, use a variable to reference it instead

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -62,6 +91,9 @@ def __init__(self,
             throw an exception if none of the nodes have ExternalDNS.
             If ``useInternalIP`` is set to true, Clipper will use the Internal IP of the K8S node
             if no ExternalDNS exists for any of the nodes.
+        kubernetes_namespace : str, optional
+            Which Kubernetes namespace to start the Clipper cluster in. The namespace must already exist.
+            If no namespace is provided, Clipper will be started in the default namespace.
```
dcrankshaw: Note: Make sure to rebase this on #500

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -85,103 +119,133 @@ def __init__(self,
...
+            time.sleep(sleep_time)
 
-        for name, img in zip(['mgmt-frontend', 'query-frontend'],
-                             [mgmt_frontend_image, query_frontend_image]):
+            self.redis_ip = 'redis-at-{cluster_name}'.format(
+                cluster_name=self.cluster_name)
+
+    def _start_mgmt(self, mgmt_image):
+        with _pass_conflicts():
+            mgmt_depolyment_data = self._generate_config(
```
dcrankshaw: nit: mispelled deployment here

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -85,103 +119,133 @@ def __init__(self,
...
+                CONFIG_FILES['management']['deployment'],
+                image=mgmt_image,
+                redis_service_host=self.redis_ip,
+                redis_service_port=self.redis_port,
+                cluster_name=self.cluster_name)
+            self._k8s_beta.create_namespaced_deployment(
+                body=mgmt_depolyment_data, namespace='default')
```
dcrankshaw: Missed a 'default' namespace reference. It's definitely worth grepping for "default" in this file to double check you didn't miss any (although I know #500 takes care of this for us).

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -85,103 +119,133 @@ def __init__(self,
...
+
+        with _pass_conflicts():
+            service_data = self._generate_config(
+                CONFIG_FILES['metric']['service'],
+                cluster_name=self.cluster_name,
+            )
+            self._k8s_v1.create_namespaced_service(
+                body=service_data, namespace=self.namespace)
+
+    def _generate_config(self, file_path, **kwargs):
```
dcrankshaw: Nice idea on using a template engine. This is much cleaner.

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -250,85 +326,30 @@ def connect(self):
...
+                version=version,
+                query_frontend_id=query_frontend_id,
+                input_type=input_type,
+                image=image,
+                cluster_name=self.cluster_name)
+
+            with _pass_conflicts():
                 self._k8s_beta.create_namespaced_deployment(
-                    body=body, namespace='default')
+                    body=generated_body, namespace='default')
```
dcrankshaw: Missed a reference to default namespace

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -250,85 +326,30 @@ def connect(self):
...
 
-                while self._k8s_beta.read_namespaced_deployment_status(
-                    name=deployment_name, namespace='default').status.available_replicas \
-                        != num_replicas:
-                    time.sleep(3)
+            while self._k8s_beta.read_namespaced_deployment_status(
+                name=deployment_name, namespace='default').status.available_replicas \
```
dcrankshaw: Missed a reference to default namespace

integration-tests/multi_tenency_test.py
```diff
@@ -0,0 +1,78 @@
+from clipper_admin import ClipperConnection, DockerContainerManager, KubernetesContainerManager
```
dcrankshaw: Nit: mispelled multi-tenancy (s/tenency/tenancy)

dcrankshaw (changes requested):
This looks great. Just had some minor comments.

dcrankshaw:
@simon-mo can you add a test for running Clipper in a user-specified namespace as well?

dcrankshaw:
Another place where we will run into contention issues is the remote docker registry for the kubernetes tests. We should tag the images we push with the commit hash, rather than using `develop`.

clipper_admin/clipper_admin/clipper_admin.py
simon-mo: There's one more occurrence of /tmp/clipper in deployer_utils (just fixed it). It is used to save python function so it's on our critical path. The python code is now fully windows compatible; however the R code still ned to fix. 

```
./containers/R/rclipper_user/R/build.R:  base_model_path = "/tmp/clipper/"
```

clipper_admin/clipper_admin/container_manager.py
simon-mo: No this refers to `ai.clipper.name` label in all our kubernetes yaml file. Cluster name set using `CLIPPER_DOCKER_LABEL` as key. (I guess we should change it in the next round of refactoring?)

clipper_admin/clipper_admin/container_manager.py
simon-mo: `query_rest` is a great name!

clipper_admin/clipper_admin/container_manager.py
simon-mo: idk no special reason it comes from previous code; i'll do 10000

clipper_admin/clipper_admin/docker/docker_container_manager.py
simon-mo: TL;DR: useful for ClipperConnection.connect()

Because the port is now floating, we no longer have knowledge of port mapping. So when user calls connect, we will know which port correspond to which container.

There will an alternative by just using docker python client but requires a "three way join" among docker containers, docker network, and some docker client magics; current approach is more straightforward to implement and understand.

clipper_admin/clipper_admin/docker/docker_container_manager.py
simon-mo: It's used to pin down specific container:

https://github.com/ucbrise/clipper/blob/1394de0867ac53efa5ed601bd91806d93d2f3ef4/clipper_admin/clipper_admin/docker/docker_container_manager.py#L211

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
simon-mo: This is an intermediate step. In the future we should adopt helm chart.

simon-mo:
@dcrankshaw 

- test added
- Contention issue goes much deeper with multi-tenancy & namespace. When user deploy models with the same name & version tag under different cluster name, we are tagging them the same way. However, I don't think this is something to address in this PR. We can say multi-tenancy support is experimental for now and fix this type of issue in management re-write.

dcrankshaw:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
There are few action items to move this PR forward:
- [x]  Change version tag in the unittest container (in `build_docker.sh`) from `develop` to git commit hash
- [x] In model deployment, image building and tagging should include the cluster-name to prevent name collision within the same registry
- [x] Change k8s docker registry from AWS to dockerhub using clipper's test account

simon-mo:
@dcrankshaw This PR is ready for another round of review now.

bin/README.md
```diff
@@ -0,0 +1,21 @@
+# Description of CI process
+
+The wonderful AMPlab Jenkins is responsible for running our integration test.
+
+## How does the CI process work
+
+
+0. Jenkins pull the PR and sandbox it.
```
dcrankshaw: Jenkins actually doesn't do any sandboxing.

bin/README.md
```diff
@@ -0,0 +1,21 @@
...
+1. Jenkins inject environment variables configured in admin page
```
dcrankshaw: Let's list what environment variables are set

bin/build_docker_images.sh
```diff
@@ -211,9 +211,14 @@ create_image () {
 
                      
     echo "Building $namespace/$image:$sha_tag from file $dockerfile"
-    time docker build --build-arg CODE_VERSION=$sha_tag $rpc_version -t $namespace/$image:$sha_tag \
+    time docker build --build-arg CODE_VERSION=$sha_tag --build-arg REGISTRY=$namespace $rpc_version -t $namespace/$image:$sha_tag \
         -f dockerfiles/$dockerfile $CLIPPER_ROOT
-    docker tag $namespace/$image:$sha_tag $namespace/$image:$version_tag
+
+    echo "Publishing $namespace/$image:$sha_tag from file $dockerfile"
+    docker push $namespace/$image:$sha_tag
```
dcrankshaw: There shouldn't be a docker push here. We only want to push images if `publish` is True.

bin/build_docker_images.sh
```diff
@@ -211,9 +211,14 @@ create_image () {
...
+
+    # We will NOT tag the image to version tag to prevent collision
+    # docker tag $namespace/$image:$sha_tag $namespace/$image:$version_tag
```
dcrankshaw: You still need this tag call to publish, just move it inside the `if publish` block.

bin/run_ci.sh
```diff
@@ -21,12 +21,7 @@ tag=$(<VERSION.txt)
 # Build docker images
 ./bin/build_docker_images.sh
 
-# Tag and push the latest version of the Clipper Docker images to the container registry
-# for the Kubernetes testing cluster
-docker tag clipper/query_frontend:$tag 568959175238.dkr.ecr.us-west-1.amazonaws.com/clipper/query_frontend:$tag
-docker push 568959175238.dkr.ecr.us-west-1.amazonaws.com/clipper/query_frontend:$tag
-docker tag clipper/management_frontend:$tag 568959175238.dkr.ecr.us-west-1.amazonaws.com/clipper/management_frontend:$tag
-docker push 568959175238.dkr.ecr.us-west-1.amazonaws.com/clipper/management_frontend:$tag
```
dcrankshaw, line 29: We don't want the docker build images script to automatically push the images. Remember, that script is very useful during development for rebuilding images locally. Instead, let's change `build_docker_images.sh` to write the names of all the containers it builds to a file (something like `clipper_docker_images.txt`) and then you can push all those images with the following one-liner (from https://stackoverflow.com/a/13939059/814642):

```
while read in; do docker push "$in"; done < clipper_docker_images.txt
```

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Hmm I'm still not sure I'm following. How exactly is this label used?

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -146,20 +156,29 @@ def __init__(self,
...
             self.k8s_namespace = namespace
         else:
-            msg = "Error connecting to Kubernetes cluster. Namespace does not exist"
+            msg = "Error connecting to Kubernetes cluster. Namespace does not exist. You can pass in KubernetesContainerManager(create_namespace_if_not_exists=True) to crate this namespcae"
             logger.error(msg)
             raise ClipperException(msg)
 
+        self.cluster_identifier = "{ns}{cluster}".format(
+            ns=self.k8s_namespace + '-'
+            if self.k8s_namespace != "default" else "",
```
dcrankshaw: This syntax is pretty convoluted. Can you write this as just a normal if-statement outside the format argument (for maintainability reasons)?

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
dcrankshaw: Helm is definitely worth looking into. Our Kubernetes cluster is a bit more dynamic than most (as we dynamically create and delete deployments and services). The right way to do it is probably to use a Helm chart to start the initial cluster, then not use Helm for creating model deployments. This is probably worth looking into when we move to having the active API server managing containers for us.

The other thing to keep in mind is that we want to be compatible with different container orchestration systems, so we never want to only support Kubernetes.

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -277,46 +297,46 @@ def connect(self):
...
-            logger.warn(msg)
+            self.logger.warn(msg)
             for addr in node.status.addresses:
                 if addr.type == "InternalIP":
                     external_node_hosts.append(addr.address)
 
         if len(external_node_hosts) == 0:
-            msg = "Error connecting to Kubernetes cluster. No external node addresses found"
-            logger.error(msg)
+            msg = "Error connecting to Kubernetes cluster. No external node addresses found. If you are running Kubernetes locally, you can pass in KubernetesContainerManager(useInternalIP=True) to connect to local Kubernetes cluster"
```
dcrankshaw: You don't need to be running Kubernetes locally to use internal ip addresses. Remove that part of the error message.

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -414,8 +435,8 @@ def get_logs(self, logging_dir):
                 log_file_name = "{pod}_{num}.log".format(
                     pod=pod.metadata.name, num=str(i))
                 log_file_alt = "{cname}.log".format(cname=c.name)
-                logger.info("log file name: {}".format(log_file_name))
-                logger.info("log alt file name: {}".format(log_file_alt))
+                self.logger.info("log file name: {}".format(log_file_name))
+                self.logger.info("log alt file name: {}".format(log_file_alt))
```
dcrankshaw: What is the purpose of `log_file_alt`? It doesn't look like it's being used at all.

dockerfiles/ClipperLibBaseDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move this below the comment, as the `REGISTRY` arg isn't being used either

dockerfiles/FrontendExporterDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw, line 2: Move below unused arg comment

dockerfiles/Py2RPCDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move below unused arg comment

dockerfiles/Py35RPCDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move below unused arg comment

dockerfiles/Py36RPCDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move below unused arg comment

dockerfiles/RContainerDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move below unused arg comment

dockerfiles/SparkScalaContainerDockerfile
```diff
@@ -1,3 +1,4 @@
+ARG REGISTRY
```
dcrankshaw: Move below unused arg comment

dcrankshaw (changes requested):
Nice job. Almost there.

clipper_admin/clipper_admin/docker/docker_container_manager.py
simon-mo: It was used here: https://github.com/ucbrise/clipper/blob/062b968f96c3821a354374f0b90d9c3776618419/clipper_admin/clipper_admin/docker/docker_container_manager.py#L207-L218

to search for live query frontend containers. The value is empty simply because we don't need to add anything. "ai.clipper.query_frontend.label" is enough.

This was introduced from last summer's big refactoring: https://github.com/ucbrise/clipper/pull/251

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
simon-mo: frankly i have no idea. I just deleted this line

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Reminder to address this comment

clipper_admin/clipper_admin/docker/docker_container_manager.py
dcrankshaw: Reminder

clipper_admin/clipper_admin/docker/docker_container_manager.py
```diff
@@ -169,17 +218,55 @@ def start_clipper(self,
...
-        return
+        """
+        Use the cluster name to update ports. Because they might not match as in
+        start_clipper the ports might be changed.
+        :return: None
+        """
+        containers = self.docker_client.containers.list(
+            filters={
+                'label': [
+                    '{key}={val}'.format(
```
simon-mo, line 253: @dcrankshaw the literal label is fixed here (same above). I don't know why github is still showing the old diff

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

simon-mo:
jenkins test this please

dcrankshaw (approved):
LGTM

Issue #518, opened 2018-06-15
[Bug] QueryFrontend's CPU usage is very high


withsmilo:
I found that QueryFrontend's CPU usage is very high on testing latest Clipper. I suspect that EstimatorFittingThread added by Adaptive Batching v2(@dcrankshaw) is running infinitely. 
To reproduce, try below steps.

### Step1 : Start hello-world sample
```python
from clipper_admin import ClipperConnection, DockerContainerManager
from clipper_admin.deployers import python as python_deployer

def feature_sum(xs):
    return [str(sum(x)) for x in xs]

clipper_conn = ClipperConnection(DockerContainerManager())
clipper_conn.start_clipper()
clipper_conn.register_application(name="hello-world", input_type="doubles", default_output="-1.0", slo_micros=100000)
python_deployer.deploy_python_closure(clipper_conn, name="sum-model", version=1, input_type="doubles", func=feature_sum)
clipper_conn.link_model_to_app(app_name="hello-world", model_name="sum-model")
```

### Step2 : Login to QueryFrontend
```
$ docker exec -it <QueryFrontend_Container_ID> /bin/bash
```

### Step3 : Check that query_frontend's CPU usage is 104.7%
```
root@837434e92e00:/# top

top - 00:28:32 up  4:23,  0 users,  load average: 1.62, 1.94, 2.00
Tasks:   3 total,   1 running,   2 sleeping,   0 stopped,   0 zombie
%Cpu(s): 25.7 us,  0.3 sy,  0.0 ni, 74.1 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem :  2046652 total,   950480 free,   284348 used,   811824 buff/cache
KiB Swap:  2097148 total,  1928304 free,   168844 used.  1583912 avail Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
    1 root      20   0  975452  30040  11224 S 104.7  1.5   0:24.44 query_frontend
   20 root      20   0   18184   3156   2712 S   0.0  0.2   0:00.02 bash
   26 root      20   0   43144   3232   2752 R   0.0  0.2   0:00.00 top
```

### Step4 : Deploy a new versioned sum-model
```python
python_deployer.deploy_python_closure(clipper_conn, name="sum-model", version=2, input_type="doubles", func=feature_sum)
```

### Step5 : Check that query_frontend's CPU usage is 197.7%
```
[...]

PR #519, opened 2018-06-16
Fix a bug that EstimatorFittingThread is running infinitely

withsmilo:
* related issue: [#538](https://github.com/ucbrise/clipper/issues/518)

### DIFF
--- a/src/libclipper/include/clipper/threadpool.hpp
+++ b/src/libclipper/include/clipper/threadpool.hpp
@@ -212,20 +212,18 @@ class ThreadPool {
   void worker(size_t worker_id, bool is_block_worker) {
     while (!done_) {
       std::unique_ptr<IThreadTask> pTask{nullptr};
-      ThreadSafeQueue<std::unique_ptr<IThreadTask>> *queue;
+      ThreadSafeQueue<std::unique_ptr<IThreadTask>>* queue;
       {
         boost::shared_lock<boost::shared_mutex> l(queues_mutex_);
         queue = &(queues_[worker_id]);
       }
       if (is_block_worker) {
-        if(queue->wait_pop(pTask))
-           pTask->execute();
+        if (queue->wait_pop(pTask)) pTask->execute();
       } else {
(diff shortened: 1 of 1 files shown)


dcrankshaw:
jenkins ok to test

src/libclipper/include/clipper/threadpool.hpp
```diff
@@ -194,6 +194,7 @@ class ThreadPool {
     boost::shared_lock<boost::shared_mutex> l(queues_mutex_);
     auto queue = queues_.find(queue_id);
     if (queue != queues_.end()) {
+      l.unlock();
```
dcrankshaw, line 197: Why do you unlock here? It looks like this would be unsafe,  and the lock will be released very soon anyway (as soon as the function completes and `l` goes out of scope).

dcrankshaw (changes requested):
Thanks for doing this. I just had a question about your unlock call.

src/libclipper/include/clipper/threadpool.hpp
withsmilo: @dcrankshaw : 
I think that `l(queues_mutex_)` is a lock to resolve mutual exclusion condition for `queues_` only, not for `ThreadSafeQueue` (the same with queue->second).

dcrankshaw: Agreed. But `queue` is an iterator that points to an entry in the `queues_` object. If `queues_` is modified, the iterator can become invalid and the `queue->second.push(...)` call could result in a segfault. Also, note that the lock `l` is a shared lock, so in the common case when you have multiple threads adding work to queues, those threads can all acquire the lock and won't block each other.

withsmilo: I understand. In line with your comment, I think that [worker codes](https://github.com/ucbrise/clipper/pull/519/files#diff-0ab69d323c13a62e04fbe37aa268462bR212) must be reverted also. And then how we can resolve this problem?

PR #526, opened 2018-07-01
Revert namespace in build_docker_images.sh

withsmilo:
`docker info` no longer shows username information for the DockerHub Registry. So 'namespace' must be reverted to build Clipper images.

### DIFF
--- a/bin/build_docker_images.sh
+++ b/bin/build_docker_images.sh
@@ -189,7 +189,12 @@ set_version_tag () {
 
 set_version_tag
 
-namespace="clipper"
+if [[ $OSTYPE == darwin* ]]
+then
+    namespace="clipper"
+else
+    namespace=$(docker info | grep Username | awk '{ print $2 }')
+fi;
 
 # Clear clipper_docker_images.txt for future write
 rm -f $CLIPPER_ROOT/bin/clipper_docker_images.txt
--- a/bin/build_docker_images.sh
+++ b/bin/build_docker_images.sh
@@ -189,7 +189,7 @@ set_version_tag () {
 
 set_version_tag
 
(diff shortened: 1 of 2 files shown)


withsmilo:
@simon-mo 
My Docker's version is 18.03.1-ce-mac65 (24312). What is the version of your Docker?

simon-mo:
@withsmilo Thanks for the fix.

It turns out docker won't printout Username if the platform is Mac but it will printout username if the platform is linux. 

We still need this line for integration tests on jenkins though. I left a comment for workaround.

bin/build_docker_images.sh
```diff
@@ -189,7 +189,7 @@ set_version_tag () {
 
 set_version_tag
 
-namespace=$(docker info | grep Username | awk '{ print $2 }')
+namespace="clipper"
```
simon-mo: Can you do:
```
# See PR #562
if [[ $OSTYPE == darwin* ]]
then 
    namespace="clipper"
else
     namespace=$(docker info | grep Username | awk '{ print $2 }')
fi;
```

simon-mo (changes requested):

bin/build_docker_images.sh
withsmilo: @simon-mo : I pushed your codes to this PR. Thank you!

simon-mo:
jenkins ok to test

withsmilo:
The error message by Jenkins is here,
```
18-07-02:02:34:43 ERROR    [clipper_metric_docker.py:126] Wrong number of targets
```
I think that my PR didn't affect `clipper_metric_docker` of the integration test.

withsmilo:
dropped because this PR is too old.

PR #527, opened 2018-07-01
[WIP] Make unittests run in parallel

simon-mo:
Use python subprocess to make py2 and py3 running in parallel.


simon-mo:
jenkins test this please

PR #532, opened 2018-07-10
Replace 'default' of 'use_default' as a kind of the prediction response key

withsmilo:
So many languages use 'default' as a keyword. 
How about replace 'default' of 'use_default' as a kind of the prediction response key?

### DIFF
--- a/containers/R/rclipper_user/vignettes/Rclipper.Rmd
+++ b/containers/R/rclipper_user/vignettes/Rclipper.Rmd
@@ -145,6 +145,6 @@ $ curl -X POST --header "Content-Type:application/json" -d '{"input": [1,2,3,4]}
 ```
 
 ```
-$ {"query_id":2,"output":4,"default":false}
+$ {"query_id":2,"output":4,"use_default":false}
 ```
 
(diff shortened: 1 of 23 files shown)


RehanSD:
Hi @withsmilo I was just curious as to why you increased the sleep time in the tests?

withsmilo:
@RehanSD : Deploying `v2` model are always failed if I didn't increase the sleep time in my Macbook. I think that we need more elegant checking method instead of sleep timer.

withsmilo:
Jenkins test this please

withsmilo:
Jenkins test this please

withsmilo:
Jenkins test this please

withsmilo:
Jenkins ok to test

simon-mo:
A quick discussion here, other please feel free to chime in. 

- This is a breaking change. Can we enable it with a flag?
- Alternarively, since we are going to expose http configuration to clipper admin, maybe we can just set the default keyword configurable?
- instead of `use_default`, maybe we can say `used_default` in past tense.

withsmilo:
@simon-mo : I am really sorry that I did not share enough information before creating this PR. In our team, there is an API server implemented by java in front of the Clipper cluster. Due to the 'default' keyword of java, the API server could not properly parse the result that Clipper returns. So our team is using it by changing it to 'use_default'.

However, through googling today, I found an interesting article([Gson Java reserved keyword - Stack Overflow](https://stackoverflow.com/questions/6258796/gson-java-reserved-keyword)), and I will discuss the possibility of applying it with my team tomorrow.

withsmilo:
@simon-mo ,
our team decided to apply the patch to our ML system. So this PR will be dropped. Thank you for your effort.

PR #559, opened 2018-08-24
Fixing serialization problem with PyTorch 0.4.1

janetsungczi:
I updated pytorch deployer and container to use pytorch built-in save and load for serialization. New image is needed. An example can be found here: clipper/examples/pytorch_updated_container/ . First run build_new_image.sh to build the new image with tag "custom-pytorch36-container:1". A better way will be update the image in dockerhub and change the image part (line 218) in pytorch deployer back to "clipper/pytorch36-container:{}".

I did not test for all python and clipper version. But regardless of python version (3.5 and 3.6), when using PyTorch 0.4.1, this method solves the bug caused by the original serialization method.


simon-mo:
jenkins test this lease

simon-mo:
jenkins ok to test

simon-mo:
This is our infrastructure tooling bug. I’ll try to fix this within a month (sorry there’s a paper deadline)

RehanSD:
@simon-mo Shouldn't this be fixed by just rolling out a new version of the pytorch deploying container with 0.4.1 installed? PyTorch is backwards compatible and the only issue we have is that the new models are serialized differently, but installing the latest version fixes that problem right?

rkooo567:
@RehanSD Is this issue fixed by your PR? Can I close?

simon-mo:
The version is updated with our newest build. A new release should resolve this.

PR #581, opened 2018-10-05
Fix 100% CPU utilization issue

dcrankshaw:
This should fix the issue. I changed the batch estimator threadpool to block if there are no tasks.

### DIFF
--- a/src/libclipper/include/clipper/threadpool.hpp
+++ b/src/libclipper/include/clipper/threadpool.hpp
@@ -392,7 +392,7 @@ inline auto submit_job(Func&& func, Args&&... args) {
 }
 
 inline void create_queue(VersionedModelId vm, int replica_id) {
-  get_thread_pool().create_queue(vm, replica_id, false);
+  get_thread_pool().create_queue(vm, replica_id, true);
 }
 
 }  // namespace EstimatorFittingThreadPool


dcrankshaw:
Jenkins test this please

withsmilo:
This patch will not fix the issue. I tested it on local with this patch.

withsmilo:
@dcrankshaw @simon-mo : 
I resolved this issue. We will have to apply the following patch.
```c++
diff --git a/src/libclipper/include/clipper/threadpool.hpp b/src/libclipper/include/clipper/threadpool.hpp
index 8da0867..4b4e282 100644
--- a/src/libclipper/include/clipper/threadpool.hpp
+++ b/src/libclipper/include/clipper/threadpool.hpp
@@ -278,7 +278,7 @@ class ModelQueueThreadPool : public ThreadPool {
   }

   bool create_queue(VersionedModelId vm, int replica_id, bool is_block_worker) {
-    boost::unique_lock<boost::shared_mutex> l(queues_mutex_);
+    boost::upgrade_lock<boost::shared_mutex> l(queues_mutex_);
     size_t queue_id = get_queue_id(vm, replica_id);
     auto queue = queues_.find(queue_id);
     if (queue != queues_.end()) {
@@ -287,6 +287,7 @@ class ModelQueueThreadPool : public ThreadPool {
                           vm.serialize(), std::to_string(replica_id));
       return false;
     } else {
+      boost::upgrade_to_unique_lock<boost::shared_mutex> upgraded_l(l);
       queues_.emplace(std::piecewise_construct, std::forward_as_tuple(queue_id),
                       std::forward_as_tuple());
       threads_.emplace(std::piecewise_construct,
@@ -392,7 +393,7 @@ inline auto submit_job(Func&& func, Args&&... args) {
 }

 inline void create_queue(VersionedModelId vm, int replica_id) {
-  get_thread_pool().create_queue(vm, replica_id, false);
+  get_thread_pool().create_queue(vm, replica_id, true);
 }

 }  // namespace EstimatorFittingThreadPool
```

simon-mo:
Thanks for the patch! I still can't reproduce it. The work queue is created only twice:
- one for TaskExecutor
- one for EstimatorFitting
(verified by logs), however the cpu remains high.

withsmilo:
@simon-mo 
I'm sorry it's my mistake. I am digging more now.

simon-mo:
Closing in favor of #608

Issue #656, opened 2019-03-25
[CI] Bug report


withsmilo:
@simon-mo , I listed up the CI bugs here.

1. ~https://amplab.cs.berkeley.edu/jenkins//job/Clipper-PRB/1818/ -> PR  https://github.com/ucbrise/clipper/pull/657 -> merged~
```
[integration_py2_admin_unit_test] 19-03-25:03:04:31 INFO     [clipper_admin.py:697] [admin-test-cluster-297] Successfully registered model jpj:iv 
 [integration_py2_admin_unit_test] 19-03-25:03:04:31 INFO     [clipper_admin.py:615] [admin-test-cluster-297] Done deploying model jpj:iv. 
 [integration_py2_admin_unit_test] 19-03-25:03:04:31 INFO     [docker_container_manager.py:353] [admin-test-cluster-297] Found 0 replicas for johnbohnam:i. Adding 1 
 [integration_py2_admin_unit_test] 19-03-25:03:04:49 INFO     [clipper_admin.py:697] [admin-test-cluster-297] Successfully registered model johnbohnam:i 
 [integration_py2_admin_unit_test] 19-03-25:03:04:49 INFO     [clipper_admin.py:615] [admin-test-cluster-297] Done deploying model johnbohnam:i. 
 [integration_py2_admin_unit_test] 19-03-25:03:04:49 INFO     [docker_container_manager.py:353] [admin-test-cluster-297] Found 0 replicas for johnbohnam:ii. Adding 1
[integration_py2_admin_unit_test] ERROR 
 [integration_py2_admin_unit_test] 19-03-25:03:05:50 INFO     [test_utils.py:75] Creating DockerContainerManager 
 [integration_py2_admin_unit_test] 19-03-25:03:05:50 INFO     [test_utils.py:80] Cleaning up Docker cluster admin-test-cluster-297 
```

2. ~https://amplab.cs.berkeley.edu/jenkins//job/Clipper-PRB/1824/ -> PR https://github.com/ucbrise/clipper/pull/655 -> merged~
This error occurred because `docker push` failed.
```
 [publish_tf36-container] Head https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/dd/dde37b0b47901bfc6be0c2857a4e9a757175ad82246ed224bd405a02f7f0e020/data?verify=1553522760-379DR9XVMPr%2FkUMzi9C2M5iBO6I%3D: net/http: TLS handshake timeout 
[...]

PR #661, opened 2019-03-28
[CI] Increase sleep time from 500us to 1000us in timers_test to avoid unintended errors

withsmilo:
* related issue : no.3 in https://github.com/ucbrise/clipper/issues/656


simon-mo (commented):
Can we use `1ms` instead of `1000us`?

withsmilo:
@simon-mo , sure, but I prefer `1000us` for code readability because the all the codes in timers_test.cpp use **microsecond** as the base unit. Do you want to change it from `1000us` to `1ms`?

simon-mo:
Good point. Let’s keep it `us` because we specify latency slo in us

simon-mo (approved):

PR #662, opened 2019-03-28
Retry when docker push fails

withsmilo:
* related issue : no.7 in https://github.com/ucbrise/clipper/issues/656
* I imported a Retry decorator from https://www.saltycrane.com/blog/2009/11/trying-out-retry-decorator-python.

### DIFF
--- a/clipper_admin/clipper_admin/clipper_admin.py
+++ b/clipper_admin/clipper_admin/clipper_admin.py
@@ -126,27 +126,30 @@ class ClipperConnection(object):
             self.cm.start_clipper(query_frontend_image, mgmt_frontend_image,
                                   frontend_exporter_image, cache_size,
                                   num_frontend_replicas)
-            while True:
-                try:
-                    query_frontend_url = "http://{host}/metrics".format(
-                        host=self.cm.get_query_addr())
-                    mgmt_frontend_url = "http://{host}/admin/ping".format(
-                        host=self.cm.get_admin_addr())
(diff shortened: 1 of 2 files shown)


simon-mo:
This looks great. We have several cases of sleep and retry in clipper_admin, do you want to refactor them as well in this PR?

withsmilo:
@simon-mo 
Yep, I will try to refactor [these codes](https://github.com/ucbrise/clipper/blob/develop/clipper_admin/clipper_admin/clipper_admin.py#L127-L142).

clipper_admin/clipper_admin/clipper_admin.py
```diff
@@ -124,28 +126,31 @@ def start_clipper(self,
...
-                    self.logger.info("Clipper still initializing: \n {}".format(e))
-                    time.sleep(1)
-            self.logger.info("Clipper is running")
-            self.connected = True
         except ClipperException as e:
             self.logger.warning("Error starting Clipper: {}".format(e.msg))
             raise e
 
+        # Wait for maximum 5 min.
+        @retry(RequestException, tries=300, delay=1, backoff=1, logger=self.logger)
```
withsmilo, line 134: @simon-mo 
Is this value suitable?

simon-mo: Yes. 5min is reasonable. We might want to increase the interval later if it prints too much intonations to the console.  However, current interval looks good enough.

simon-mo (approved):

Issue #597, opened 2018-11-10
Kubernetes model deployment permissions issue


dipayan90:
I see this as a common scenario where in organizations that use a k8s cluster with multi tenancy. Usually every team within the Org is given `admin` `ClusterRole` permissions. Now using the `admin` role, one has permissions to `get`, `delete`, `update`, `list` deployments but not the `status` sub-resource. The admin role permissions do not let you write deployment status. They let you create and delete the deployment objects, controlling the "spec" portion of the object. Status modification permissions are granted to the deployment controller.

When deploying a model `clipper` relies on getting the `available replicas` here: https://github.com/ucbrise/clipper/blob/3c5a1cc6ce59e0ccd778f526a50808d0e7b2576f/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py#L392

Now a `namespace` admin will not have access to `status` sub resource. However the information about number of available replicas can be obtained by just calling the https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/ExtensionsV1beta1Api.md#read_namespaced_deployment as well . the status subresource doesn't give you any more information than simply fetching the deployment

It will be great if the switch is made to using the `read_namespaced_deployment` api endpoint instead of the `read_namespaced_deployment_status` endpoint.

withsmilo:
Interesting point! 
@simon-mo : What do you think of this issue?

dipayan90:
@withsmilo @simon-mo Also another question is there a reason why `NodePort` was used as the K8s `ServiceType` instead of `LoadBalancer` ?

simon-mo:
@dipayan90 
For using `read_namespaced_deployment` api: 
- Yes this will be great! 
- We don't have much experience implement Clipper with permission control. We will probably start with exploring RBAC for prometheus. 

For `NodePort` instead of `LoadBalancer`:
- When we started implementing it, we found `NotePort` is sufficient. But we should probably switch to `LoadBalancer` for production usage. 
- As I undestand, `Loadbalancer` requires external IP address when it's deployed in a real K8s cluster. This requires a bit more special handling on Clipper side as well.

withsmilo:
@dipayan90 Fixed. Thanks! These commits will be contained within next release.

PR #667, opened 2019-04-04
Change k8s ServiceTypes to be changeable

withsmilo:
When creating a KubernetesManager object, you could define own ServiceTypes. For example,
```python
cm = KubernetesContainerManager(
         ....,
         service_types={
             'redis': 'NodePort',
             'management': 'LoadBalancer',
             'query': 'LoadBalancer',
             'query-rpc': 'ClusterIP',
             'metric': 'LoadBalancer'})
```
* related issue : https://github.com/ucbrise/clipper/issues/597

### DIFF
--- a/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
+++ b/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
@@ -153,14 +153,6 @@ class KubernetesContainerManager(ContainerManager):
         else:
             self.use_k8s_proxy = False
 
-        self.redis_ip = redis_ip
-        self.redis_port = redis_port
-        self.useInternalIP = useInternalIP
-        config.load_kube_config()
-        configuration.assert_hostname = False
-        self._k8s_v1 = client.CoreV1Api()
-        self._k8s_beta = client.ExtensionsV1beta1Api()
-
         # Create the template engine
         # Config: Any variable missing -> Error
         self.template_engine = jinja2.Environment(
@@ -168,6 +160,14 @@ class KubernetesContainerManager(ContainerManager):
             undefined=jinja2.StrictUndefined)
 
         self.service_types = self._determine_service_types(service_types)
+
+        self.redis_ip = redis_ip
+        self.redis_port = redis_port
+        self.useInternalIP = useInternalIP
+        config.load_kube_config()
+        configuration.assert_hostname = False
+        self._k8s_v1 = client.CoreV1Api()
+        self._k8s_beta = client.ExtensionsV1beta1Api()
 
         # Check if namespace exists and if create flag set ...create the namespace or throw error
         namespaces = []
--- a/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
+++ b/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 from ..container_manager import (
     create_model_container_label, ContainerManager, CLIPPER_DOCKER_LABEL,
-    CLIPPER_MODEL_CONTAINER_LABEL, CLIPPER_QUERY_FRONTEND_ID_LABEL,
+    CLIPPER_MODEL_CONTAINER_LABEL, CLIPPER_INTERNAL_RPC_PORT,
     CLIPPER_INTERNAL_MANAGEMENT_PORT, CLIPPER_INTERNAL_QUERY_PORT,
     CLIPPER_INTERNAL_METRIC_PORT, CLIPPER_NAME_LABEL, ClusterAdapter)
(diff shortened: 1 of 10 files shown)


simon-mo:
@withsmilo  Thanks for this patch. However, I think we also need to patch how `connect()` works because it currently assumes `NodePort` like line:
https://github.com/ucbrise/clipper/blob/227c8e46e5a801e9f94e7e42a5bf0b693a5f9919/clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py#L350

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -168,6 +192,18 @@ def __init__(self,
             self.cluster_identifier = "{ns}-{cluster}".format(
                 ns=self.k8s_namespace, cluster=self.cluster_name)
 
+        self.service_types = DEFAULT_CLIPPER_SERVICE_TYPES
```
rkooo567: Why don't we make this part as a function and add some tests?

withsmilo: @rkooo567 , OK. Let me refactor this part.

withsmilo:
@simon-mo , I will try to add something to need.

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
withsmilo: @rkooo567 , What tests are we needed here in additional?

rkooo567: I thought just a simple unit test? Like if this function raises Exception correctly when wrong input is given and etc. I could find that the current CI fails on line 200 (when you raise `ClipperException`), so it can be good if we can write some unit tests inside one of integration test files.

wcwang07: @withsmilo @rkooo567 can we add

def set_clipper_svc_types(self, svctype, svcname):
        for key in self.OFFICIAL_K8S_SERVICE_TYPE:
            if(svctype.lower()==key.lower()):
                self.DEFAULT_CLIPPER_SERVICE_TYPES[svcname]=key

withsmilo: @rkooo567 @wcwang07 
Sorry about late reply. I refactored my PR entirely.

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -315,54 +372,79 @@ def _generate_config(self, file_path, **kwargs):
...
+            mgmt_frontend_ports = v1service.spec.ports
             for p in mgmt_frontend_ports:
-                if p.name == "1338":
+                if int(p.name) == CLIPPER_INTERNAL_MANAGEMENT_PORT:
                     self.clipper_management_port = p.node_port
-                    self.logger.info("Setting Clipper mgmt port to {}".format(
-                        self.clipper_management_port))
 
-            query_frontend_ports = self._k8s_v1.read_namespaced_service(
+            if self.service_types['management'] in [CLUSTER_IP, NODE_PORT]:
```
rkooo567, line 429: Why don't we read service_types from our Kubernetes definitions (yaml file) in the `connect` function? Otherwise we should always specify the correct `self.service_types` whenever we initiate `KubernetesContainerManager`. 

It shouldn't matter when we initiate `KubernetesContainerManager` for the first time for `start_clipper`. But, let's say we already started clipper and just wanted to connect to the cluster using `ClipperConnection(KubernetesContainerManager()).connect()`. If you forget to write `service_types`, it will be set to be None and this part might be broken.

withsmilo: @rkooo567 You're right. Thank you for your kind advice. I will fix it as soon as possible.

withsmilo: @rkooo567 : According to your advice, I added a new mechanism to read/write `service_types` from/to a yaml file. You can connect to the executing Clipper cluster easily with `service_types`(None), which will be loaded automatically from a yaml file saved when we started it. If you call 'stop_all()', the yaml file will be removed.

rkooo567: Thank you! Everything looks good to me :)

withsmilo:
@simon-mo : Please review this PR. Can we use 'LoadBalancer' service type with Travis CI?

rkooo567:
This PR looks good to me. But since it is not a simple PR, I will wait for @simon-mo's approval.

simon-mo (approved):
lgtm

Issue #678, opened 2019-04-29
Current CI Broken at docker_metric.


rkooo567:
<img width="848" alt="Screen Shot 2019-04-28 at 10 41 52 PM" src="https://user-images.githubusercontent.com/18510752/56915921-8d3b5980-6a6c-11e9-88cb-0144f043d170.png">

Current CI has been broken at `docker_metric` integration test due to the error,
```
[integration_py3_docker_metric] 19-04-29:10:25:29 ERROR [clipper_metric_docker.py:127] Failed to parse: http://localhost:23298/api/v1/series?match[]=clipper_mc_recv_time_ms_sum
``` 
I found out `urllib3` recently released their new version 1.25 on April 22nd (which is the day our CI starts to be broken). 

Our requests library uses `urllib3` as its dependency. I strongly suspect the version of urllib3 is the cause of our broken CI. I also found this issue from `requests` library. It says there's an issue with `urllib3 1.25`. (It is not directly related to our issue btw). 
https://github.com/kennethreitz/requests/issues/5067

Here is also the logs that I found from our Jenkins
```
[integration_py3_docker_metric] /usr/local/lib/python3.5/dist-packages/requests/__init__.py:91: RequestsDependencyWarning: urllib3 (1.25.1) or chardet (3.0.4) doesn't match a supported version!

[integration_py3_docker_metric] RequestsDependencyWarning)
```

As you can see, we are currently using `urllib3 1.25.1`. I will create a PR to restrict the urllib3's version to be less than 1.25 and see if it resolves our problem. 

-> I verified the problem is because urllib3 version is set to be 1.25.1. In local vm environment, it was broken if I use 1.25.1. We should use either 1.25.2 or <1.25

rkooo567:
It is merged and resolved. I will keep opening the issue so that we can remove urllib3 version constraint later once it becomes stabilized

rkooo567:
https://github.com/kennethreitz/requests/pull/5092 says request library is now compatible to urllib 1.25.2. But it is still not compatible to 1.25.1 or 1.25. I will just keep them as is because using 1.25.2 wouldn't make any difference.

PR #679, opened 2019-04-29
add urllib3 version restriction to requirements.txt

rkooo567:
Details are in https://github.com/ucbrise/clipper/issues/678. Once `requests` library starts to be compatible with `urllib3 >= 1.25`, we should remove this line.

### DIFF
--- a/clipper_admin/requirements.txt
+++ b/clipper_admin/requirements.txt
@@ -6,3 +6,4 @@ kubernetes==6.0.0
 mock
 prometheus_client
 redis
+urllib3<1.25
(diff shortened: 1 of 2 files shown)


rkooo567:
I added urllib3 constraint to Dev35Dockerfile as well. And the previous CI shows that `urllib3==1.25.2` does not generate `failed to parse` error. I think the problem is within `urllib3==1.25.1`.

simon-mo:
I'm re-running travis. Once it pass we should be fine.

simon-mo (approved):

PR #685, opened 2019-05-05
[CI] Try to stabilize Travis job

withsmilo:
There are some cases where Travis rarely fails due to an unknown error.
I think we need a little bit of waiting time to finish one test, delete the clipper cluster,
and then recreate the clipper cluster for a new test. Let's try one patch to solve this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ sudo: required
 
 language: python
 
+python: "2.7"
+
 env:
   matrix:
   # This will make sure .minikube config resides in current user folder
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,6 +10,8 @@ sudo: required
 
 language: python
 
+python: "2.7"
```
rkooo567, line 13: Why don't we specify all the versions we support? I believe we support 2.7, 3.5, and 3.6.

simon-mo: This will actually create many builds. Just 2.7 is fine for now.

simon-mo (approved):

.travis.yml
withsmilo: Simon is right. Currently the Travis is used only for the Kubernetes integration tests.

Issue #683, opened 2019-05-03
What you need to Modify the current develop version to get ensemble results


coldtomatojuice:
Currently Clipper can only serve result from a single model that is linked to an application where a REST end point is set. however, most functions needed for serving results from multiple models are implemented in the code and I guess it is for the future revision that might enable clipper to select optimized models among all linked models according to a policy chosen. 

As per the issue #579 and others, there are some requirements to use multiple models linked to a single application, that is, an ensemble method. However, since it is only possible to get a single model result, at one REST request, to implement ensemble model, users must program their own method, and this was annoying.

Below is slight modifications of the code to let clipper to serve multiple model results at one REST request. To do that, 

it changes 
 - code that disables to run functions if there exists more thatn a model linked to an application
 - output return types of data
 - process multiple output result iteratively
 - post-process for multi model outputs

it uses the default policy, such that a single model result can still be served

it currently merges data from models and returns, but if you want ensembles of results,
you can implement your own ensemble methods in 'query_frontend.cpp'

below are what you need to fix in the Clipper
you can also check it from my fork : https://github.com/coldtomatojuice/clipper (Branch: ensemble)

[clipper_admin.py]
in the function 'def link_model_to_app(self, app_name, model_name)'
get argument 'model_name' as a string list (not a string)
so that, it can pass REST API a model name list

line:285
```
req_json = json.dumps({
            "app_name": app_name,
            "model_names": model_name
        })
```

[datatypes.cpp]
Change the output return type
Output output -> std::vector<Output> output

line:238
```
Response::Response(Query query, QueryId query_id, const long duration_micros,
                   std::vector<Output> output, const bool output_is_default,
                   const boost::optional<std::string> default_explanation)

```
[datatypes.hpp]
Change the output return type
Output output -> std::vector<Output> output

line:284
```
Response(Query query, QueryId query_id, const long duration_micros,
           std::vector<Output> output, const bool is_default,
           const boost::optional<std::string> default_explanation);

```
Change the output return type

line:300
`std::vector<Output> output_;
`
[query_processor.cpp]
Get a vector<Output> return from combine_predictions
Output -> vector<Output>

line:138
```
std::pair<vector<Output>, bool> final_output = current_policy->combine_predictions(
        selection_state, query, *outputs_ptr);

```
[selection_policies.hpp]
Modify a parameter for the 'combine_predictions'

line:67
```
virtual const std::pair<std::vector<Output>, bool> combine_predictions(
[...]

PR #686, opened 2019-05-07
enable clipper to receive results from multiple linked models

coldtomatojuice:
details are described in the issue #683 

There is no model selection policy, 
it only enables users to link multiple models to a single application, 
and receive all the results from models linked.

return values are in an array form. 

ensemble method can be implemented in 'query_frontend.hpp' if one wants to merge results before hand it over to a REST API response.

the single model result is the same as before, except that it is covered with an additional list bracket


withsmilo:
@coldtomatojuice : Thank you for you contribution! :) We will talk about this on https://github.com/ucbrise/clipper/issues/683.

simon-mo:
close for now. @rkooo567 will add this to contrib doc

Issue #551, opened 2018-08-15
[Bug] SIGSEGV occurred in the SimpleWebServer very rarely


withsmilo:
The stacktrace is here.
```
Error: signal 11:
/clipper/release/src/frontends/query_frontend(_Z7handleri+0x18)[0x562370b80278]
/lib/x86_64-linux-gnu/libc.so.6(+0x33060)[0x7fcbd0df5060]
/clipper/release/src/frontends/query_frontend(_ZN5boost4asio6detail28reactive_socket_service_base8start_opERNS2_24base_implementation_typeEiPNS1_10reactor_opEbbb+0x69)[0x562370b8a509]
/clipper/release/src/frontends/query_frontend(_ZN5boost4asio6detail26read_until_delim_string_opINS0_19basic_stream_socketINS0_2ip3tcpENS0_21stream_socket_serviceIS5_EEEESaIcEZN9SimpleWeb10ServerBaseIS8_E24read_request_and_contentERKSt10shared_ptrIS8_EEUlRKNS_6system10error_codeEmE_EclESK_mi+0x3cc)[0x562370bb86fc]
/clipper/release/src/frontends/query_frontend(_ZN5boost4asio6detail23reactive_socket_recv_opINS0_17mutable_buffers_1ENS1_26read_until_delim_string_opINS0_19basic_stream_socketINS0_2ip3tcpENS0_21stream_socket_serviceIS7_EEEESaIcEZN9SimpleWeb10ServerBaseISA_E24read_request_and_contentERKSt10shared_ptrISA_EEUlRKNS_6system10error_codeEmE_EEE11do_completeEPNS1_15task_io_serviceEPNS1_25task_io_service_operationESM_m+0x207)[0x562370bb97c7]
/clipper/release/src/frontends/query_frontend(_ZN5boost4asio6detail13epoll_reactor16descriptor_state11do_completeEPNS1_15task_io_serviceEPNS1_25task_io_service_operationERKNS_6system10error_codeEm+0x13e)[0x562370b8973e]
/clipper/release/src/frontends/query_frontend(_ZNSt6thread11_State_implISt12_Bind_simpleIFZN9SimpleWeb10ServerBaseIN5boost4asio19basic_stream_socketINS5_2ip3tcpENS5_21stream_socket_serviceIS8_EEEEE5startEvEUlvE_vEEE6_M_runEv+0x428)[0x562370b8d638]
[...]

PR #691, opened 2019-05-12
Update SimpleWebServer to the latest version(29 Apr, 2019 - d6985ed6)

withsmilo:
* related issue : https://github.com/ucbrise/clipper/issues/551
* Diff between original SimpleWebServer(left) and new SimpleWebServer modified for Clipper(right) is [here](https://www.diffchecker.com/JTVrqOWs).

### DIFF
--- a/src/libs/httpserver/server_http.hpp
+++ b/src/libs/httpserver/server_http.hpp
@@ -12,6 +12,9 @@
 #include <thread>
 #include <unordered_set>
 
+// [CLIPPER]
+#include <boost/thread.hpp>
+
 #ifdef USE_STANDALONE_ASIO
 #include <asio.hpp>
 #include <asio/steady_timer.hpp>
--- a/src/libs/httpserver/server_http.hpp
+++ b/src/libs/httpserver/server_http.hpp
@@ -1,339 +1,523 @@
-/*
-   The MIT License (MIT)
-
-   Copyright (c) 2014-2016 Ole Christian Eidheim
-
-   Permission is hereby granted, free of charge, to any person obtaining a copy
-   of this software and associated documentation files (the "Software"), to deal
-   in the Software without restriction, including without limitation the rights
-   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-   copies of the Software, and to permit persons to whom the Software is
-   furnished to do so, subject to the following conditions:
-
-   The above copyright notice and this permission notice shall be included in
-   all
-   copies or substantial portions of the Software.
-
-   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


src/libs/httpserver/server_http.hpp
```diff
@@ -1,339 +1,526 @@
...
+
+      regex_orderable(const char *regex_cstr) : regex::regex(regex_cstr), str(regex_cstr) {}
+      regex_orderable(std::string regex_str_) : regex::regex(regex_str_), str(std::move(regex_str_)) {}
+      bool operator<(const regex_orderable &rhs) const noexcept {
+        return str < rhs.str;
+      }
+    };
+
+  public:
+    // [CLIPPER]
```
rkooo567, line 381: I found in the previous code, these were formatted (I believe we ran the formatter before?). We don't use the C++ formatter anymore? @simon-mo @withsmilo

rkooo567 (commented):
Thank you for the PR! I appreciate you added [CLIPPER] comment line so that we can easily update the part whenever we update this dependency. 

Everything looks good to me, but I have a few questions.

1. Aren't we using C++ formatter anymore? I found some parts of code were formatted in the previous SimpleWebServer V2.1.1. 
2. This version (d6985) is literally the newest commit of SimpleWebServer. Do you think it is stable enough? I am afraid we have to keep updating the version because the newest master is unstable. Is it the version your team is currently using?

withsmilo:
@rkooo567 
1. AFAIK, the C++ code formatter is turned off now.
2. We cannot use SimpleWebServer V2.1.1 in the production environment because it has https://github.com/ucbrise/clipper/issues/551 issue. Our team is using ver3(`21 Sep, 2018 - cfe98038`), and will test / update to the latest ver(`29 Apr, 2019 - d6985ed6`) this week. I think that there will not be a big problem because the changes applied between them are very minor. Please wait for the test result to run for a week.

src/libs/httpserver/server_http.hpp
```diff
@@ -1,339 +1,526 @@
...
-   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-   SOFTWARE.
-*/
-
 #ifndef SERVER_HTTP_HPP
```
rkooo567, line 1: Why don't we add a README.md to record the version/revision hash information? Seems like some C++ dependencies keep track of its version in its directory name, but httpserver does not. I believe it is because we have some additional custom code inside. We can also probably mention whenever we update the version of `httpserver`, we should remain the parts with `// [CLIPPER]`.

simon-mo (approved):

simon-mo:
Thanks! I'll directly merge this one since test passed already.

PR #700, opened 2019-05-17
Print necessary stackstace information when crashed

withsmilo:


withsmilo:
https://amplab.cs.berkeley.edu/jenkins//job/Clipper-PRB/1975/ shows this error,
```
 [pytorch-container] The command '/bin/sh -c pip install -q torch==1.0.* torchvision==0.2.*' returned a non-zero code: 2 
CI_build.Makefile:472: recipe for target 'pytorch-container' failed
make: *** [pytorch-container] Error 2
```

withsmilo:
Jenkins test this please

rkooo567 (commented):
LGTM in general! I still have a couple questions. 

1. How is it different from just using `docker logs` on a broken container? Is it for our CI environment? 2. I found this notes from `backtrace_symbols_fd`

backtrace() and backtrace_symbols_fd() don't call malloc()
          explicitly, but they are part of libgcc, which gets loaded
          dynamically when first used.  Dynamic loading usually triggers a
          call to malloc(3).  If you need certain calls to these two
          functions to not allocate memory (in signal handlers, for
          example), you need to make sure libgcc is loaded beforehand.

Are we loading `libgcc` beforehand? Is it already included in our compilation process?
EDIT -> Actually 2 might not matter because we print logs when containers are broken? Is docker container freeing all of its memory when it is broken?

withsmilo:
@rkooo567 
This PR will capture SIGSEGV(segmentation fault) only.
1. After applying this, we could view stacktrace logs by using `docker logs` on a broken container.
2. You're right. The resources of the broken container will be freed automatically.

withsmilo:
Jenkins test this please

rkooo567 (approved):
Looks nice!

Issue #703, opened 2019-05-21
Looks like our main build still fails.


rkooo567:
```
 [integration_py2_mxnet] Sleep 122 secs before starting a test 
 [integration_py2_mxnet] Starting Trial 0 with timeout 2400.0 seconds 
 [integration_py2_mxnet] Sleep 265 
 [integration_py2_mxnet] Starting Trial 1 with timeout 2400.0 seconds 
 [integration_py2_mxnet] Sleep 390 
 [integration_py2_mxnet] All retry failed. 
CI_test.Makefile:152: recipe for target 'integration_py2_mxnet' failed
make: *** [integration_py2_mxnet] Error 1
make: *** Waiting for unfinished jobs....
```

Sometimes, our mxnet doesn't even start and fail. I think it is one of reasons why our main build is failing. Any idea about why?

rkooo567:
Also, does anyone know why we have tag integration_test_py2 when we use python3?
`python3 ./bin/colorize_output.py --tag integration_py2_mxnet`
https://github.com/ucbrise/clipper/blob/9e41fb24cb2fd2983e10d1aa71e0751e72b7be5c/bin/shipyard/shipyard.py#L121

withsmilo:
### 1. integration_py2_mxnet
The 'integration_py2_mxnet' test failed to start `docker run --rm --network=host -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp         -e CLIPPER_REGISTRY=clipper         -e CL    IPPER_TESTING_DOCKERHUB_PASSWORD=$CLIPPER_TESTING_DOCKERHUB_PASSWORD         clipper/unittests:develop         "python /clipper/integration-tests/deploy_mxnet_models.py"`, but didn't print any error logs. Let's print them.
```
bin/retry_with_timeout.py
@@ -49,7 +49,7 @@ except AssertionError:

 def run_once_with_timeout(command_to_run, timeout):
     proc = subprocess.Popen(
-        command_to_run, stdout=sys.stdout, stderr=subprocess.STDOUT
+        command_to_run, stdout=sys.stdout, stderr=sys.stdout
     )
     start = time.time()
     while True:
```

### 2. python3 ./bin/colorize_output.py --tag integration_py2_mxnet
This is just for printing logs. I think it's not the cause.

rkooo567:
I created a PR #707. Please don't merge the PR.

rkooo567:
[...]

PR #675, opened 2019-04-23
Updagrade the version of mxnet to 1.4.* for mxnetcontainerdockerfile.

rkooo567:
This PR handles an issue https://github.com/ucbrise/clipper/issues/674. Because of internal changes of `mxnet`, `cloudpickle` is not able to deserialize a function with `mxnet`. I resolved this issue by updating the mxnet version to 1.4.*. 

# Test summary
https://github.com/rkooo567/clipper/tree/mxnet-version-upgrade
1. Created a custom dockerfile with new mxnet version
2. When deploying using `mxmodel_deployer`, I pulled a custom dockerfile from my dockerhub.
3. Run `deploy_mxnet_models.py` from `integration_test`

I verified it worked for `mxnet==1.4.0` and `mxnet==1.1.*`

PR #707, opened 2019-05-24
[CI] Fix maintenance scripts

rkooo567:
This will print why mxnet fails even before it starts. We should probably not merge it. Let's keep running it until `mxnet` fails without running. 

Reference: https://github.com/ucbrise/clipper/issues/703

### DIFF
--- a/bin/cleanup_jenkins.sh
+++ b/bin/cleanup_jenkins.sh
@@ -7,7 +7,7 @@ CLIPPER_REGISTRY=clippertesting
 try_cleanup() {
     (docker ps --all | grep "$@" | awk '{ print $1 }' | xargs docker kill) || true
     (docker container ls --all | grep "$@" | awk '{ print $3 }' | xargs docker container rm) || true
-    (docker image ls --filter "label=maintainer=Dan Crankshaw <dscrankshaw@gmail.com>" | awk '{ print $3 }' | xargs docker image rm -f) || true
+    (docker image ls --all | grep "$@" | awk '{ print $3 }' | xargs docker image rm -f) || true
     docker image prune -f
 }
 
--- a/bin/cleanup_jenkins.sh
+++ b/bin/cleanup_jenkins.sh
@@ -10,5 +10,9 @@ try_cleanup() {
     (docker image ls --all | grep "$@" | awk '{ print $3 }' | xargs docker image rm -f) || true
 }
 
+try_cleanup_docker_volume() {
+    (docker volume ls -f dangling=true | awk '{ print $2 }' | xargs docker volume rm -f) || true
+}
 try_cleanup CLIPPER_REGISTRY > /dev/null
 try_cleanup shipyard > /dev/null
+try_cleanup_docker_volume > /dev/null
--- a/bin/cleanup_jenkins.sh
+++ b/bin/cleanup_jenkins.sh
@@ -5,9 +5,9 @@ CLIPPER_REGISTRY=clippertesting
 
 # best effort, ignore all errors
 try_cleanup() {
-    (docker ps | grep "$@" | awk '{ print $1 }' | xargs docker kill) || true
-    (docker container ls | grep "$@" | awk '{ print $3 }' | xargs docker container rm) || true
-    (docker image ls | grep "$@" | awk '{ print $3 }' | xargs docker image rm -f) || true
+    (docker ps --all | grep "$@" | awk '{ print $1 }' | xargs docker kill) || true
+    (docker container ls --all | grep "$@" | awk '{ print $3 }' | xargs docker container rm) || true
+    (docker image ls --all | grep "$@" | awk '{ print $3 }' | xargs docker image rm -f) || true
 }
 
 try_cleanup CLIPPER_REGISTRY > /dev/null
--- a/bin/cleanup_jenkins.sh
+++ b/bin/cleanup_jenkins.sh
@@ -7,12 +7,13 @@ CLIPPER_REGISTRY=clippertesting
 try_cleanup() {
     (docker ps --all | grep "$@" | awk '{ print $1 }' | xargs docker kill) || true
     (docker container ls --all | grep "$@" | awk '{ print $3 }' | xargs docker container rm) || true
-    (docker image ls --all | grep "$@" | awk '{ print $3 }' | xargs docker image rm -f) || true
+    (docker image ls --filter "label=maintainer=Dan Crankshaw <dscrankshaw@gmail.com>" | awk '{ print $3 }' | xargs docker image rm -f) || true
+    docker image prune -f
 }
 
 try_cleanup_docker_volume() {
     (docker volume ls -f dangling=true | awk '{ print $2 }' | xargs docker volume rm -f) || true
 }
-try_cleanup CLIPPER_REGISTRY > /dev/null
+try_cleanup ${CLIPPER_REGISTRY} > /dev/null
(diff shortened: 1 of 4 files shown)


rkooo567:
@withsmilo Looks like even with this change, the error log is not printed. Let me think of other way to do this.

rkooo567:
jenkins test this please

withsmilo:
@rkooo567 
```
 [integration_py2_mxnet] Sleep 482 secs before starting a test 
 [integration_py2_mxnet] Starting Trial 0 with timeout 2400.0 seconds 
 [integration_py2_mxnet] output: None 
 [integration_py2_mxnet] err: None 
 [integration_py2_mxnet] Sleep 230 
 [integration_py2_mxnet] Starting Trial 1 with timeout 2400.0 seconds 
 [integration_py2_mxnet] output: None 
 [integration_py2_mxnet] err: None 
 [integration_py2_mxnet] Sleep 295 
 [integration_py2_mxnet] All retry failed. 
CI_test.Makefile:152: recipe for target 'integration_py2_mxnet' failed
make: *** [integration_py2_mxnet] Error 1
make: *** Waiting for unfinished jobs....
```

withsmilo:
@rkooo567 
I found an interesting thing. If Jenkins executes a job on `amp-jenkins-staging-worker-02`, it will always have success about  'integration_py2_mxnet', but if on `amp-jenkins-staging-worker-07` or `amp-jenkins-staging-worker-08`, it will always fail.

rkooo567:
@withsmilo That's interesting... @simon-mo Do you have any input about this? 

Also, it looks like the error is still not printed. I will push a new commit to resolve this.

rkooo567:
@withsmilo Seems like if the error is due to Docker command failure, it still should print the error logs. I suppose the docker daemon is not even responding to the command. I added a ping to check if docker daemon responds correctly.

rkooo567:
@withsmilo If it is a docker command failure, it should still print out the logs. I guess it is that the request to docker daemon is hanging and timeout (It is still weird because then other tests should fail as well, but mxnet is the only test it fails. I will also modify the order of tests and see if mxnet is still broken). I will try to call some docker commands to check connectivity.

rkooo567:
@simon-mo @withsmilo One possibility might be that the docker volume is not cleaned properly at worker 7 & 8, and because of that the daemon is hanging. I printed volumes in the new commit here, and seems like there are lots of volumes that are not deleted (I am not sure if it is not deleted, or there are just lots of volumes because we are running lots of tests. I need some verification.)

withsmilo:
Good point. According to the https://docs.docker.com/storage/volumes/#remove-volumes, the anonymous volume will be deleted automatically with —rm option. However, in case of the named volume, we have to execute `docker volume prune` to remove it. Therefore we have to add a cleanup command to ‘cleanup_jenkins.sh’ for docker volume. I will do it tonight.

withsmilo:
@rkooo567 
There were too many debugging logs in there. So I changed to check Docker's status only if a timeout has occurred.

withsmilo:
jenkins test this please

rkooo567:
@withsmilo After changing the order of tests, the problem still occurs at mxnet test. I suppose specifically mxnet has a problem. One possibility is this PR. https://github.com/ucbrise/clipper/pull/675
Also, I will print debugging logs only for mxnet test

withsmilo:
@rkooo567 
It is useless to change the order of tests because it is defined as the target and dependencies relationship in the CI_test.Makefile. And your codes(https://github.com/ucbrise/clipper/pull/707/commits/7c3412947df453ac3c0140d24f8dcc65cc51ae23) were not executed at all in https://amplab.cs.berkeley.edu/jenkins//job/Clipper-PRB/2029/.
```
integration_py2: integration_py2_admin_unit_test ... integration_py2_mxnet ... integration_py2_fluentd
```
 Let's test integration_py*_mxnet only. New Jenkins job have to be executed at work 7 or 8 for test.

simon-mo:
Hi all,

I limited the pull request builder to only use node7 and 8.

withsmilo:
@simon-mo 
Thanks. https://amplab.cs.berkeley.edu/jenkins//job/Clipper-PRB/2030/ was executed at worker 2. Let's run it again.

withsmilo:
@rkooo567 
https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/2031/ shows that many old Docker images were not deleted.

withsmilo:
I added a command `docker image ls --filter "label=maintainer=Dan Crankshaw <dscrankshaw@gmail.com>" | awk '{ print $3 }' | xargs docker image rm -f` to cleanup old Docker images. **This is a temporary patch for maintenance.** After cleaning, I will add a new label to all the Clipper images, and use it to cleanup.

rkooo567:
@withsmilo Sounds good.

Also, why do you think it only occurs at mxnet test? If it is because of out of disk (which we are thinking), isn't it supposed to happen in other tests as well? I guess there might be some other causes, but cannot think of any. 

@simon-mo Let us know if you have any clue.

withsmilo:
@rkooo567 I don't know why. 
@simon-mo Can you connect to the worker 7 or 8 through terminal? If you can, it might be a best solution to execute a mxnet test command directly at worker 7 or 8.

withsmilo:
@rkooo567 @simon-mo 
https://amplab.cs.berkeley.edu/jenkins/job/Clipper-PRB/2033/ shows,
```
 [integration_py3_mxnet] /usr/bin/python: line 2:     6 Illegal instruction     (core dumped) python3 "$@" 
```
I guess that CPU in worker 7 & 8 is incompatible with the latest mxnet. https://github.com/apache/incubator-mxnet/issues?utf8=%E2%9C%93&q=is%3Aissue+Illegal+instruction
Can we exclude worker 7 & 8 from Clipper workers?

withsmilo:
Some of the modifications included in this PR are worth merge with `develop` branch. After resolving `integration_py3_mxnet` issue, let's create a new PR or modify this one.

rkooo567:
@withsmilo Agree. Let's just remove debug logs and merge this PR. Once it is merged, I will create an issue to handle how to delete dangling images.

simon-mo:
@withsmilo, @rkooo567 and I found out that only worker 1 and 2 have avx2 and mxnet requires avx2. I constrained the test to run on worker 2 for now. (worker 1 is down).

This is a temporary fix. I will start investigating alternative ways to do our CIs.

simon-mo:
Also, the volumes and images are cleaned each night via a cronjob. 

```
1) docker-gc (https://github.com/spotify/docker-gc)
2) stackoverflow knowledge:


$ cat cleanup-docker.sh
#!/bin/bash

# remove exited containers:
docker ps --filter status=dead --filter status=exited -aq | xargs -r docker rm -v

# remove unused images:
docker images --no-trunc | grep '<none>' | awk '{ print $3 }' | xargs -r docker rmi

# remove unused volumes:
find '/var/lib/docker/volumes/' -mindepth 1 -maxdepth 1 -type d | grep -vFf <(
  docker ps -aq | xargs docker inspect | jq -r '.[] | .Mounts | .[] | .Name | select(.)'
) | xargs -r rm -fr

docker volume ls -qf dangling=true | xargs -r docker volume rm```

withsmilo:
@simon-mo Great! I understood that we have to run just one Jenkins job only before revolving CI worker's problem.

withsmilo (approved):
LGTM

simon-mo (approved):

Issue #566, opened 2018-09-08
Why not use torch.save(model, PATH) to serialize the entire model?


wnagchenghku:
Hi, in https://github.com/ucbrise/clipper/blob/develop/clipper_admin/clipper_admin/deployers/pytorch.py#L201, I found clipper serializes the model and weight separately using `cloudpickle` and `torch.save(the_model.state_dict(), PATH)`. I'm curious why does clipper do this serialization separately? From https://pytorch.org/docs/stable/notes/serialization.html, pytorch can actually serialize the entire model in one step. I tested that loading the entire model using torch.load is much much much faster than cloudpickle.

rkooo567:
@simon-mo @withsmilo Any thought?

simon-mo:
We should. Can you submit a PR?

rkooo567:
@simon-mo Yep sounds good. I will wait for the author for a couple days and start if he doesn't respond.

@wnagchenghku Would you want to create a PR for this?

PR #713, opened 2019-05-31
Used pytorch's standalone serialization function. instead of unnecess…

rkooo567:
# Related
https://github.com/ucbrise/clipper/issues/566

# Summary
Currently pytorch deployer serializes its state and model separately. But, as you can see from this link, https://pytorch.org/docs/stable/notes/serialization.html, pytorch supports standalone model serialization. It means that cloudpickle is unnecessary. The PR removes cloudpickle model serialization from pytorch `container` and `deployer`.

# Note
Dont't review yet. Once test passes, I will clean up code. I will let you guys know after that.


rkooo567:
@simon-mo @withsmilo Will close this PR because it can break the backward compatibility. (The previously pickled models cannot be retrieved from new pytorch containers).

withsmilo:
@rkooo567 @simon-mo hmm.. Wouldn't it be okay to do that because it's an internal change?

rkooo567:
@withsmilo Let's say old version is 0.4 and the new version is the one that applies this change. What I concerned was that old version's pickled mode cannot be run by this new pytorch container because it doesn't know how to deserialize old version's models. 

But actually, old version's models are already deployed in the cluster. So the models that will use this new version's pytorch container will be ones that will be deployed by the new deployers. So, I think you are right. (It doens't matter). I will reopen the PR.

Issue #576, opened 2018-09-27
Deal with trailing slash on predict url


hsubbaraj:
Deal with trailing slash on predict url

querying 

`localhost:1337/pong/predict/`

should behave the same as

`localhost:1337/pong/predict`

rkooo567:
@simon-mo Is it possible in the current version?

rkooo567:
Resolved with #721

PR #721, opened 2019-06-04
Handle trailing slash on predict url

withsmilo:
* related issue : https://github.com/ucbrise/clipper/issues/576

### DIFF
--- a/integration-tests/clipper_admin_tests.py
+++ b/integration-tests/clipper_admin_tests.py
@@ -437,7 +437,9 @@ class ClipperManagerTestCaseShort(unittest.TestCase):
         time.sleep(60)
 
         addr = self.clipper_conn.get_query_addr()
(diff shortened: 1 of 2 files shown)


integration-tests/clipper_admin_tests.py
```diff
@@ -674,6 +674,28 @@ def test_deployed_model_queried_successfully(self):
         self.assertNotEqual(parsed_response["output"], self.default_output)
         self.assertFalse(parsed_response["default"])
 
+    def test_deployed_model_queried_with_training_slash_successfully(self):
```
rkooo567: Why don't we add a check in `test_test_predict_function` instead of creating a new test?

withsmilo: Good. I was thinking a little bit about whether to add a new test.

rkooo567 (approved):
Looks nice!

Issue #580, opened 2018-10-01
Python version error ( add support for 3.7)


gauthamzz:
` ClipperException: Python closure deployer only supports Python 2.7, 3.5, and 3.6. Detected 3.7 `

when excecuting the basic example 
```
 python_deployer.deploy_python_closure(clipper_conn, name="sum-model", version=1, input_type="doubles", func=feature_sum) 
   ...:                                                                                                                                                
18-10-01:18:04:52 INFO     [deployer_utils.py:44] Saving function to /tmp/clipper/tmp0rq77bb7
18-10-01:18:04:52 INFO     [deployer_utils.py:54] Serialized and supplied predict function
18-10-01:18:04:52 INFO     [python.py:192] Python closure saved
18-10-01:18:04:52 ERROR    [python.py:214] Python closure deployer only supports Python 2.7, 3.5, and 3.6. Detected 3.7
---------------------------------------------------------------------------
ClipperException                          Traceback (most recent call last)
<ipython-input-8-f507994c395f> in <module>
----> 1 python_deployer.deploy_python_closure(clipper_conn, name="sum-model", version=1, input_type="doubles", func=feature_sum)

~/hacks/thesis/clipper/venv/lib/python3.7/site-packages/clipper_admin/deployers/python.py in deploy_python_closure(clipper_conn, name, version, input_type, func, base_image, labels, registry, num_replicas, batch_size, pkgs_to_install)
    215             # Remove temp files
    216             shutil.rmtree(serialization_dir)
--> 217             raise ClipperException(msg)
    218 
    219     # Deploy function

ClipperException: Python closure deployer only supports Python 2.7, 3.5, and 3.6. Detected 3.7
```

Python version 3.7.0
OS Manjaro Linux

simon-mo:
@gauthamzz Thanks for the report! We are aware of it. Python3.7 will be added very soon

gauthamzz:
Is there a workaround. Other than changing Python version for whole system?
Anything with virtualenv or something ?

simon-mo:
If you run your code under virtualenv it should be choosing the corresponding version.

rkooo567:
@gauthamzz I believe @withsmilo will handle this. I will create an issue for this.

withsmilo:
resolved by #723

PR #723, opened 2019-06-04
Support Python 3.7

withsmilo:
* related issue : https://github.com/ucbrise/clipper/issues/580
* This PR will create new Docker images.
  * py37-rpc
  * mxnet37-container
  * pytorch37-container
  * tf37-container
  * pyspark37-container
  * python37-closure-container
  * keras37-container


clipper_admin/clipper_admin/deployers/keras.py
```diff
@@ -221,9 +221,13 @@ def predict(model, inputs):
             logger.info("Using Python 3.6 base image")
             base_image = "{}/keras36-container:{}".format(
                 __registry__, __version__)
+        elif py_minor_version == (3, 7):
+            logger.info("Using Python 3.7 base image")
+            base_image = "{}/keras36-container:{}".format(
```
jacekwachowiak: I think this should be `keras37`

withsmilo: Oh! my mistake! I will fix it tonight.

rkooo567:
CI jobs are pended a lot. The job for the last commit will be running from now on.

clipper_admin/clipper_admin/deployers/keras.py
withsmilo: fixed!

rkooo567:
jenkins test this please.

rkooo567 (dismissed):
Looks good to me! I will merge if the last CI is passed.

rkooo567:
@simon-mo I will wait for you to merge.

simon-mo:
@withsmilo Thank you for the great work!

Can we:
- [ ] Build a python3.7 test container and test deploy_python_closure?
- [ ] In the doc, say that 3.7 support is still experimental (because we haven't tested the compatibility with other frameworks)

withsmilo:
@simon-mo 
Current `py35tests` Docker image built by `ClipperPy35TestsDockerfile`, uses Python 3.6. We have two choices.

1. How about try to upgrade the Python version from 3.6 to 3.7 in `py35tests`? If we do this, we will use python 3.7 as the base version for unit testing. If we don't use a library of dependencies on Python 3.6, this approach will be no problem.

2. Let's conduct a temporary unit test with python 3.7.

\+ we need to rename `ClipperPy35TestsDockerfile` to `ClipperPy3TestsDockerfile`.

withsmilo:
@simon-mo I will test to use python 3.7 as the base version for unit testing.

withsmilo:
Jenkins test this please

rkooo567:
@withsmilo Current CI is very unstable with this log. 

```
ReadTimeout: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60) 
```

It can be one of our dependency issues like `urllib3` last time. I will run the `develop` on my vm and see if it is reproduced.

rkooo567:
2 things.

1. Based on logs of job 2074, it seems like new python dev container is failed to be built. (So, I think it is independent to our main build failure)
2. I could success to pass test on my local environment for current `develop`, so the error might be Jenkins issue. We'll see if build keeps failing tomorrow.

withsmilo:
@rkooo567 Wow, so unstable. :disappointed_relieved: Our Jenkins worker is working too hard now. (https://amplab.cs.berkeley.edu/jenkins/computer/amp-jenkins-staging-worker-02/builds)
I will run this PR this night again.

withsmilo:
Jenkins test this please

withsmilo:
Docker daemon might be killed on Jenkins worker. I’ll waiting for recovery. 

>`
failed to get digest sha256:fe3ef29c73f3ebaffa0dead8391b75be18894d771a841c28ca1140fec358c5e2: open /var/lib/docker/image/aufs/imagedb/content/sha256/fe3ef29c73f3ebaffa0dead8391b75be18894d771a841c28ca1140fec358c5e2: no such file or directory`

rkooo567:
Jenkins test this please

rkooo567:
@withsmilo @simon-mo The temporary test passes. I think we should decide what version of python we will test. What I am worried is that we will break python 3.6 or 3.5 because of features that are only supported by 3.7, but we don't know that happens because we only test with python 3.7. What do you guys think is the best idea for testing?

1. Use python 3.7 since most of new users will be use python 3.7.
2. Use python 3.6. Since we proved python 3.7 passes unit tests, it should be fine to keep developing without testing python 3.7 because it is backward compatible.
3. Test both. (I think it is not a good idea given our resources).

withsmilo:
@rkooo567 @simon-mo I agree we can't keep all the Python versions for unit test. How about this?
* current unit test targets
  * python 2.7
  * python 3.6
* changed unit test targets on Jan. 1, 2020. (python 2.7 will be retired. ([link](https://pythonclock.org/)))
  * python 3.6
  * python 3.7

withsmilo:
hmm.. I will add some retry routines to handle Docker's timeout errors.

rkooo567:
@withsmilo I agree that we can stop supporting python 2.7 around the time python 2.7 retires. 👍 

Also, about CI, if retry doesn't work, we should probably investigate the problem like the last time (mxnet). If that happens, I will create a new issues!

withsmilo:
Please do not merge yet. I will test to change unit test targets to python 3.6 and python 3.7.

withsmilo:
@rkooo567 @simon-mo 
After Python 2.7 retired, we have to activate these lines to use Python 3.6 and 3.7 for unit test.
https://github.com/ucbrise/clipper/pull/723/files#diff-72ca7a5f8b86c887fd5695574f26fce0R65

simon-mo:
@withsmilo After 2020, I'm fine with even just testing with 3.6. and only run a small subset of the tests in other python version if only necessary. bascially:

- python3.6: full set
- python3.5, python 3.7: only deploy python closure.

simon-mo (approved):

withsmilo:
@simon-mo Good idea!

Issue #729, opened 2019-06-10
Method `stop_models` gives unexpected behaviour on Docker and Kubernetes


jacekwachowiak:
I am running the current develop branch version of Clipper.
I have encountered strange outputs while using the `clipper_conn.stop_models`.
1. On Docker the method crashes with a lengthy output but does remove the running container (the output is pasted at the bottom).

2. On Kubernetes the method returns:
```
clipper_conn.stop_models('stop-test')
19-06-10:15:43:40 INFO     [clipper_admin.py:1285] Model stop-test:1 was successfully deleted
19-06-10:15:43:40 INFO     [clipper_admin.py:1319] [default-default-cluster] Stopped all containers for these models and versions:
{'stop-test': ['1']}

clipper_conn.stop_versioned_models({"stop-test": "1"})
19-06-10:15:43:51 ERROR    [clipper_admin.py:1280] Received error status code: 400 and message: model with name 'stop-test' and version '1' does not exist

clipper_conn.get_all_models()
[]

[cloud-user@k1]$ date
Mon Jun 10 15:46:09 JST 2019
[cloud-user@k1]$ kubectl get po -o wide | grep stop
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7-7bpcb   1/1     Running   0          3m29s   10.233.96.7    node2   <none>           <none>
```
so if I understand correctly Clipper says that the model was removed correctly, but then as you can see the model pod is still running, it is orphaned and undetected by Clipper but still managed by Kubernetes.  Using `stop_all_model_containers` does nothing. Only running `stop_all` cleans the pods.

Before I could delete the Docker models without the error so I guess it must be some very recent change. I am not sure if the pod removal for Kubernetes worked though. This is the Docker error output:
```
---------------------------------------------------------------------------
RemoteDisconnected                        Traceback (most recent call last)
~/.pyenv/versions/3.6.8/lib/python3.6/site-packages/urllib3/connectionpool.py in urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, **response_kw)
    600                                                   body=body, headers=headers,
--> 601                                                   chunked=chunked)
    602 

~/.pyenv/versions/3.6.8/lib/python3.6/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    386                     # otherwise it looks like a programming error was the cause.
[...]

PR #732, opened 2019-06-13
Fixing the `stop_models` functions by calling explicitly the removal of the replica sets and pods

jacekwachowiak:
I have noticed that by calling the `stop_models` or `stop_all_model_containers` only a Kubernetes deployment is being removed, while the replica sets and pods remain. Even worse is that the pods are orphaned and cannot be managed or detected by Clipper. 
I added 2 calls to remove the corresponding replica sets and pods. From what I understand, using the Python API stops the Kubernetes calls from cascading, which means that the deployment may not remove the replica set and in effect the pods. I checked that even with a replica set deletion the pods are not gone, so to clean completely both remaining resources are removed with separate calls.
Issue #729

Some console output to confirm that it works
```
clipper_conn.get_all_models()

['stop-test:1', 'sum-model:1']

[cloud-user@node1 ~]$ kubectl get rs | grep deployment
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7   1         1         1       78s
sum-model-1-deployment-at-0-at-default-cluster-6c4f5bbdc7   1         1         1       92s

[cloud-user@node1 ~]$ kubectl get po | grep deployment
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7-tnvcp   1/1     Running   0          93s
sum-model-1-deployment-at-0-at-default-cluster-6c4f5bbdc7-fqtpx   1/1     Running   0          107s

clipper_conn.stop_models('sum-model')
19-06-13:11:54:01 INFO     [clipper_admin.py:1285] Model sum-model:1 was successfully deleted
19-06-13:11:54:01 INFO     [clipper_admin.py:1319] [default-default-cluster] Stopped all containers for these models and versions:
{'sum-model': ['1']}

[cloud-user@node1 ~]$ kubectl get rs | grep deployment
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7   1         1         1       2m8s
[cloud-user@node1 ~]$ kubectl get po | grep deployment
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7-tnvcp   1/1     Running       0          2m9s
sum-model-1-deployment-at-0-at-default-cluster-6c4f5bbdc7-fqtpx   1/1     Terminating   0          2m23s

clipper_conn.stop_all_model_containers()
19-06-13:11:54:54 INFO     [clipper_admin.py:1285] Model stop-test:1 was successfully deleted
19-06-13:11:54:54 INFO     [clipper_admin.py:1410] [default-default-cluster] Stopped all Clipper model containers:
{'stop-test': ['1']}

[cloud-user@node1 ~]$ kubectl get rs | grep deployment
[cloud-user@node1 ~]$ kubectl get po | grep deployment
stop-test-1-deployment-at-0-at-default-cluster-7b586b67f7-tnvcp   1/1     Terminating   0          2m58s
[cloud-user@node1 ~]$ kubectl get po | grep deployment
```


simon-mo:
jenkins ok to test

simon-mo (approved):
@jacekwachowiak It's strange to see this kind of behavior in k8s api. but we have to live with it. Thanks for catching it

simon-mo:
I'll wait for @withsmilo to review this before merging https://github.com/ucbrise/clipper/issues/729#issuecomment-501526343

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -632,6 +632,25 @@ def stop_models(self, models):
                             val=create_model_container_label(m, v),
                             cluster_label=CLIPPER_DOCKER_LABEL,
                             cluster_name=self.cluster_name))
+
+                    self._k8s_beta.delete_collection_namespaced_replica_set(
+                        namespace=self.k8s_namespace,
+                        label_selector=
```
withsmilo: It seems there are redundant codes. Would like to define `label_selector` as a variable?

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
```diff
@@ -646,6 +665,23 @@ def stop_all_model_containers(self):
                     label=CLIPPER_MODEL_CONTAINER_LABEL,
                     cluster_label=CLIPPER_DOCKER_LABEL,
                     cluster_name=self.cluster_name))
+
+            self._k8s_beta.delete_collection_namespaced_replica_set(
+                namespace=self.k8s_namespace,
+                label_selector="{label}, {cluster_label}={cluster_name}".
```
withsmilo: It seems there are redundant codes. Would like to define `label_selector` as a variable?

jacekwachowiak:
@simon-mo As I discussed with @withsmilo in the issue #729, there is a cleaner way to do  this with `propagation_policy='Foreground'` optional argument using `delete_namespaced_deployment()` instead of `delete_collection_namespaced_deployment()`, but this applies only to a single deployment and I cannot find a way to get all the models while calling the `stop_all_model_containers`. And I don't think it's a good idea to have two different implementations for the single model stopping and for all of them, so unless there is a way to get the list of all models inside the `stop_all_model_containers`, I see no other way

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
jacekwachowiak: Sure, I copy-pasted the parts so it's not pretty, I will fix it now :)

clipper_admin/clipper_admin/kubernetes/kubernetes_container_manager.py
jacekwachowiak: Done

withsmilo (approved):
LGTM!

rkooo567:
LGTM too! Let's merge it when test is passed.

rkooo567 (approved):

PR #738, opened 2019-07-01
Fix a bug around internal node search

rkooo567:
When searching internal ip, KubernetesContainerManager does not properly search nodes. It should iterate all the nodes options like external IP search. Otherwise, it will always use the last `node` of previous node.items iteration.


rkooo567:
Jenkins test this please

withsmilo (approved):
Good catch! (i'm sorry about late reply. I'm so busy.)

PR #757, opened 2019-11-25
add instructions to install fmt

coldtomatojuice:
recently I tried to bulid Clipper using 0.4 release clone, and I got the error message as below

`CMake Error at CMake/folly-deps.cmake:214 (find_package):
  Could not find a package configuration file provided by "fmt" with any of
  the following names:

    fmtConfig.cmake
    fmt-config.cmake

  Add the installation prefix of "fmt" to CMAKE_PREFIX_PATH or set "fmt_DIR"
  to a directory containing one of the above files.  If "fmt" provides a
  separate development package or SDK, be sure it has been installed.
Call Stack (most recent call first):
  CMakeLists.txt:113 (include)
`

It seems there have been some minor changes in folly
therefore it needs to install fmt from source before install folly

check below issue for the details 
https://github.com/facebook/folly/pull/1263

thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+This is a forked repository from the original clipper library. ucbrise/clipper <br>
+It enables clipper to output results from multiple models <br>
+You can check the details of this modification from below <br>
+https://github.com/ucbrise/clipper/issues/683#issue-439847004
+
 # Clipper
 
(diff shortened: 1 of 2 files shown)


robberlang:
This helped me, thanks. The `git clone` line (line 20) needs a backslash at the end of it though.

